PR #53, opened 2020-09-04
validate config

prasunanand:
Fixes https://github.com/Quansight/qhub/issues/7

Should we create a github-action for validate ?

I am unable to comment on PR with the errors found.


prasunanand:
Better solution: https://github.com/Quansight/qhub/pull/119

PR #69, opened 2020-09-24
adding GPU profile based on qhub-ops-config

prasunanand:


qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/gcp.tf
```diff
@@ -33,6 +33,14 @@ module "kubernetes" {
       instance_type = "{{ nodegroup_config.instance }}"
       min_size      = {{ nodegroup_config.min_nodes }}
       max_size      = {{ nodegroup_config.max_nodes }}
+    {% if nodegroup == "user_gpu" %}
```
costrouc, line 36:
```
{% if "accelerators" in nodegroup_config %}
guest_accelerators = [
{% for accelerator in nodegroup_config.guest_accelerators %}
        {
          type  = "{{ accelerator.name }}"
          count = {{ accelerator.count }}
        }
{% endfor %}
      ]
{% endif %}
```

tests/assets/config_gcp_gpu.yaml
```diff
@@ -0,0 +1,140 @@
...
+      max_nodes: 1
+    user:
+      instance: n1-standard-2
+      min_nodes: 0
+      max_nodes: 2
+    user_gpu:
+        instance: n1-standard-2
+        min_nodes: 0
+        max_nodes: 2
+        gpu: nvidia-tesla-p4
```
costrouc, line 50:
```yaml
       accelerators:
          - name: nvidia-tesla-p4
            count: 1
```

costrouc (changes requested):
Suggesting these changes so that we can add gpus to any node_group and allow for multiple accelerators if needed

prasunanand:
@costrouc Your comments have been addressed. Please review!

qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/gcp.tf
```diff
@@ -33,6 +33,18 @@ module "kubernetes" {
       instance_type = "{{ nodegroup_config.instance }}"
       min_size      = {{ nodegroup_config.min_nodes }}
       max_size      = {{ nodegroup_config.max_nodes }}
+    {% if cookiecutter.gpu_enabled is defined %}
+    {% if ((cookiecutter.gpu_enabled) * (nodegroup == "user")) == 1 %}
```
costrouc, line 37: Why is this defined this way? We want the user to be able to create arbitrary node groups with gpus in them. Not necessarily just "user". 

```
{% if cookiecutter.gpu_enabled is defined %}
{% if ((cookiecutter.gpu_enabled) * (nodegroup == "user")) == 1 %}
```

Can just be `{% if nodegroup_config.guest_accelerators is defined %}` and remove `gpu_enabled = True`

prasunanand: yeah. You are right

costrouc: Motivation here is the user may possibly want a node group with small gpus and another with large expensive ones that require a larger cpu/ram.

PR #74, opened 2020-09-24
Qhub init

prasunanand:
Fixes https://github.com/Quansight/qhub/issues/73


costrouc:
Having the config files in templates will cause them to be rendered into the cookiecutter template. Could you move the config files outside of the templates folder. I think you will have to change MANIFEST.in to include those yaml files.

costrouc:
Also please add documentation on the new command into the docs

prasunanand:
@costrouc  Only this directory is rendered https://github.com/Quansight/qhub/tree/master/qhub/template/%7B%7B%20cookiecutter.repo_directory%20%7D%7D

prasunanand:
```
prasun@prasun-xps:~/dev/new$ qhub render -c qhub-ops-config.yaml -o ./ --force
prasun@prasun-xps:~/dev/new$ tree | grep qhub-ops-config
├── qhub-ops-config.yaml
```

prasunanand:
I put these files in `templates` directory because it sounds like we are following the `template` pattern.

costrouc:
Thanks for the docs and awesome glad we can keep it in the templates directory. Could you also add a test to render do, aws, and gcp?

costrouc (approved):

PR #76, opened 2020-09-24
Adding patch ability to qhub generated repository allowing for arbitray customization to configuration

costrouc:
This is a solution to address https://github.com/Quansight/qhub/issues/58. This works by generating the repository like normal from `qhub-ops-config.yaml` and then applying an arbitrary number of `git apply <patch-filename>` to repository. This will allow any customization to the repositiory. Also since patches can apply to multiple files in a single patch we can have well named patches e.g. "add-prefect-shorterm-fix.patch" etc. Also another benifit of patches is that they will complain if the code it is patching changes too much making. For an example see `tests/assets/config_do_patch.yaml`. There is a pytest test to ensure this works.

# How to create a git patch

https://nithinbekal.com/posts/git-patch/ start with nothing staged in the directory

```
# add the files that you changed
git add <files>
# create a patch from these changes `patches/<patch-name>.patch` is a good directory/filename
git diff --cached > <patch-filename>
# add the patch filename to `qhub-ops-config.yaml` under the list of top level key `patches`
```

The benefit of this approach is that it completely covers the complex use case, is reasonably hard to do creating an intentional barrier to entry for this feature.

### DIFF
--- a/qhub/render/__init__.py
+++ b/qhub/render/__init__.py
@@ -101,6 +101,7 @@ def render_template(
         with filename.open() as f:
             config = yaml.safe_load(f)
             config["repo_directory"] = repo_directory
+            config["run_directory"] = pathlib.Path.cwd()
             patch_dask_gateway_extra_config(config)
 
         with (input_directory / "cookiecutter.json").open() as f:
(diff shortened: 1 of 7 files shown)


prasunanand:
```
git add <myfiles>
git diff -cached > mypatch.patch
```

costrouc:
Closing to prefer terraform cdk https://github.com/Quansight/qhub-cloud/issues/327

PR #83, opened 2020-09-25
Enable DigitalOcean to use private container registry

costrouc:
Due recent support added in qhub-terraform-modules https://github.com/Quansight/qhub-terraform-modules/pull/22

This is still a WIP need to find out how to automate https://www.digitalocean.com/docs/images/container-registry/quickstart/#use-images-in-your-registry-with-kubernetes


costrouc:
Note that this does not fully solve it. Need to set Kubernetes permissions -- https://www.digitalocean.com/docs/images/container-registry/quickstart/#use-images-in-your-registry-with-kubernetes

Issue #107, opened 2020-10-07
Check formatting of generated templates for qhub rendered templates


costrouc:

costrouc:
Closing with PR

PR #108, opened 2020-10-07
Fix terraform formatting issue with aws kubernetes.tf generation

costrouc:
Causes downstream errors with generated terraform files. Need to add to qhub github-actions to do a formatting check #107 which will be in new PR


tylerpotts (approved):
LGTM

PR #109, opened 2020-10-07
bump release

tylerpotts:


costrouc:
For future reference this PR did not increment the qhub/VERSION file which caused the release to fail.

Issue #99, opened 2020-09-29
PV backups / snapshots


prasunanand:
Add a method to backup Persistent Volume claims

tylerpotts:
- Investigate and document how to backup persistent volumes

prasunanand:
https://kubernetes.io/blog/2019/12/09/kubernetes-1-17-feature-cis-volume-snapshot-beta/

tylerpotts:
Based on a conversation with @costrouc , I'm going to investigate creating a kubernetes job to tarball the directories and upload to s3

rsignell-usgs:
@tylerpotts and @costrouc , I believe we heard on the call last week that this capability now exists? 
cc @jkellndorfer

tylerpotts:
@rsignell-usgs This was my work-in-progress branch to solve this problem:

https://github.com/Quansight/qhub/tree/backup_jobs/backup_files

This was my approach:

- Created a docker image with (fsspec)[https://filesystem-spec.readthedocs.io/en/latest/] installed
- Wrote a script that tarballs the user directory and uploads to S3/Google Storage
- Create a kubernetes yaml that mounts the NFS volume and runs the script in the docker container I created

I didn't get around to writing the restoration yaml config, but it's similar to the above, in reverse (download tarball, untar in same directory as the original tar). 

This is how you would use my current work:
- Create a cloud storage bucket and add the name of the bucket in (this line)[https://github.com/Quansight/qhub/blob/backup_jobs/backup_files/backup_spec.yaml#L20]
- Make sure your environment variables are set for your cloud credentials
- Run `kubectl apply -f backup_spec.yaml`

I plan on coming up with a backup solution that is more elegant than this in the future. If either of you come up with a better solution I'd love to see it in a PR :)

dharhas:
https://cloud.google.com/blog/products/storage-data-transfer/introducing-filestore-backups @tylerpotts @costrouc

laisbsc:
As for a suggestion of an architectural diagram for the backup strategy check [Tyler's comment](https://github.com/Quansight/qhub/issues/227#issuecomment-755384525).

tylerpotts:
- [ ] Figure out cronjob
- [ ] Put secret in qhub-config.yaml

Longer term
- [ ] Put secrets in CICD
Secrets get rendered, don't get stored in repo

tylerpotts:
[...]

PR #133, opened 2020-10-17
[WIP] Add shared volume backup functionality

tylerpotts:
Currently, this code only works on GCP

This yaml file will kick off a job that will tarball the shared nfs volume and upload it to Google Cloud Storage.

A similar yaml could be used to restore the volume. I've verified that using this tar command will keep all the groups and permissions of the files: `tar -xvf shared_backup.tar.gz --same-owner `

### DIFF
--- a/backup_files/Dockerfile
+++ b/backup_files/Dockerfile
@@ -1,0 +1,5 @@
+FROM python:3
+
+RUN pip install --upgrade pip && pip install --no-cache-dir fsspec gcsfs s3fs
+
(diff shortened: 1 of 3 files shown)


tylerpotts:
This is an out-dated PR. This method of backing up is below. The latest issue tracking backups is #99 

https://github.com/Quansight/qhub/blob/main/docs/source/04_how_to_guides/5_system_backup.md

danlester:
I have closed this PR.

PR #136, opened 2020-10-20
Fixes typo in step-by-step

arokem:
The existing text didn't work for us, but fixing thusly made it go


tylerpotts:
@arokem Right you are!

PR #137, opened 2020-10-21
Jkellndorfer patch step by step guide 1

jkellndorfer:
Here are some proposed changes. Foremost I thought the "scripts" reference was not necessary since config and deployment is handled nicely with "qhub render" and "qhub deploy". Tried to also make it more generic for other DNS services and be a little more prescriptive of how to use AWS vs DO/GCP. And since it's October, and the PR is accepted, I guess there is the Hacktoberfest merit possible...


costrouc (approved):

tylerpotts:
@jkellndorfer I think this label should get you kudos with hacktoberfest. Let me know if there's any trouble with it

PR #139, opened 2020-10-21
Jkellndorfer patch step by step 3

jkellndorfer:
Even more subtle changes fixing numbering for 1) to 1. and setting the environment variable in section 4. To be checked by the team if that makes sense  ... Ignore previous request.


tylerpotts (approved):
@jkellndorfer Thanks for the contribution! There was a merge conflict that I fixed. LGTM!

jkellndorfer:
@tylerpotts thanks, Tyler. I also just noticed that when I build qhub from the github repo, it needs python 3.8. That should probably be noted in the installation instructions in the guidle. I let you handle where it would go best.

PR #181, opened 2020-11-14
Fix bug in `initialize.py`

filippo82:
The GitHub client secret wasn't saved because of a bug: probably a copy and paste error.


tylerpotts (approved):
LGTM

PR #184, opened 2020-11-17
Define terraform version in qhub config

tylerpotts:


qhub/schema.py
```diff
@@ -186,6 +188,15 @@ class Main(Base):
     ci_cd: CiEnum
     security: Security
     profiles: Profiles = []
+    terraform_version: str
+
+    @pydantic.validator("terraform_version")
+    def verify_terraform_version(cls, v):
```
aktech, line 194: I think we should have terraform version as optional, otherwise this would not be backwards compatible.

qhub/schema.py
```diff
@@ -186,6 +188,15 @@ class Main(Base):
...
+        minor_release = re.search(r"(\d+)\.(\d+)", v).group(0)
+        if minor_release not in SUPPORTED_TERRAFORM_MINOR_RELEASES:
+            raise ValueError(
+                f"terraform_version must be defined in qhub-config.yaml and a release of one of the following minor versions: {SUPPORTED_TERRAFORM_MINOR_RELEASES}"
```
aktech, line 198: The above [comment](https://github.com/Quansight/qhub/pull/184/files#r525339925), is for this line.

costrouc (approved):

qhub/schema.py
tylerpotts: I think we are ok with it not being backward compatible. If the version is not defined then users will be informed that it needs to be added to the error log. Especially since we're not at a 1.0 release yet, I don't think we have set an expectation for backward compatibility.

aktech: Alright, cool!

PR #187, opened 2020-11-17
Update to include installation with conda

rsignell-usgs:
Add installation with Conda to the step-by-step now that we have https://anaconda.org/conda-forge/qhub


tylerpotts (approved):
Thanks for getting this all setup! LGTM

PR #191, opened 2020-11-24
Add logging around output parsing failure

aktech:
Sometimes parsing the terraform output fails for no reason, this basically adds some logging around it and also spits the command output for debugging.


tylerpotts (approved):
LGTM

PR #208, opened 2020-12-15
Local deployment

brl0:
Adds option to deploy locally.  Need to change reference to branch in `qhub-terraform-modules` before merging.

### DIFF
--- a/qhub/cli/initialize.py
+++ b/qhub/cli/initialize.py
@@ -9,7 +9,7 @@ def create_init_subcommand(subparser):
         "platform",
         help="Cloud to deploy qhub on",
         type=str,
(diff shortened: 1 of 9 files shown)


costrouc:
Closing for bigger PR #287 which has a fully tested local deployment (almost complete)

brl0:
@costrouc, This PR uses `kind` (Kubernetes in Docker), which can be provisioned directly with Terraform, which is distinct from deploying to an existing local Kubernetes cluster.  Just a thought, but perhaps this feature could still be integrated as a separate `kind` provider?  I can make the updates and changes necessary if you are open to the idea.

PR #226, opened 2021-01-05
Fix json parsing issue by using regex instead

aktech:
Currently we have a problem with parsing json on Github Actions, this replaces json parsing and uses regex matching for solving this problem.


tylerpotts:
@aktech Made a minor update to the comment numbering. LGTM. Merging

tylerpotts (approved):

PR #245, opened 2021-01-15
Adding password based auth for QHub

### LINKED ISSUE #207: Make auth more flexible and not require oauth
Make the default authorization for QHub Bcrypt. This will remove the requirement for DNS and make testing easier (could go to IP of cluster and login instead of waiting for DNS registration) as well get us one step closer to on-prem cluster deployment.

* Make auth more flexible and not require oauth. 
    * OAuth requires https which requires DNS. 
        * So if we use oauth -> we have to have https. 
        * This prevents an on prem deployment along with making testing harder and slower.
    * Make auth more flexible:
        * username/password based auth

Suggested modification to `qhub-config.yaml`

```
security:
  authentication:
    type: password
  users:
    "chris.ostrouchov@gmail.com":
      uid: 1000
      primary_group: users
      password: "salted password with bcrypt base64?" "$5:thesalt:encryptedpassword"
    "dharhas@quansight.com":
      uid: 1001
      primary_group: admin
      password: "salted password with bcrypt base64?"
  groups:
    users:
      gid: 1000
    admin:
      gid: 1001
````

### PR DESCRIPTION
**Pull request by costrouc**
Closes #207

### CODE REVIEW

File: qhub/initialize.py [hunk 1]
```diff
@@ -207,8 +211,8 @@ def render_config(
     config["domain"] = qhub_domain
     oauth_callback_url = f"https://jupyter.{qhub_domain}/hub/oauth_callback"
 
-    if oauth_provider == "github":
-        config["security"]["authentication"] = OAUTH_GITHUB
+    if auth_provider == "github":
```
aktech, line 214: How about something like: `AUTH_OAUTH_GITHUB.get('type').lower`?

aktech (approved):
This is excellent work, I haven't tried creating an instance with password authentication myself, but otherwise it looks good to me.

costrouc:
Closing for bigger PR #287 which has a fully tested local deployment (almost complete)

PR #287, opened 2021-01-29
Enable QHub local deployment

costrouc:


costrouc:
This PR is close to being ready for review. The last missing piece is getting the minikube test working and deploying with each PR. This will add significantly more testing to each deployment

costrouc:
Ready for review and merging

aktech (commented):
Is there a specific value, you would like us to add here for the start and end IPs?:
```
minikube addons configure metallb
-- Enter Load Balancer Start IP: 172.17.10.0
```

I tried deploying it and I get the following error:


```

Warning: Interpolation-only expressions are deprecated

  on .terraform/modules/qhub/modules/kubernetes/services/meta/qhub/main.tf line 16, in module "kubernetes-jupyterhub":
  16:           "${var.general-node-group.key}" = var.general-node-group.value

Terraform 0.11 and earlier required all non-constant expressions to be
provided via interpolation syntax, but this pattern is now deprecated. To
silence this warning, remove the "${ sequence from the start and the }"
sequence from the end of this expression, leaving just the inner expression.

Template interpolation syntax is still used to construct strings from
expressions when the template includes multiple interpolation sequences or a
mixture of literal strings and interpolations. This deprecation applies only
to templates that consist entirely of a single interpolation sequence.

(and 3 more similar warnings elsewhere)


Error: context deadline exceeded



Error: Waiting for rollout to finish: 1 replicas wanted; 0 replicas Ready



Error: context deadline exceeded



Error: Waiting for rollout to finish: 1 replicas wanted; 0 replicas Ready


INFO:qhub.deploy:deploying QHub took 661.937 [s]
```
Any idea what could I have missed?

docs/docs/development.md
```diff
@@ -0,0 +1,152 @@
...
+
+Configure `metallb`
+
+```shell
+minikube addons configure metallb
+```
+
+If you don't want to configure metallb interactively here is a short
+bash command to run. This is used in the github actions since the
+minikube command does [not provide a no interactive simple way to
```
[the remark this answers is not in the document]
costrouc: changes made.

docs/docs/development.md
```diff
@@ -0,0 +1,152 @@
...
+### Initialize kubernetes cluster
+
+Testing done with minikube
+
+```shell
+minikube start --cpus 2 --memory 4096 --driver=docker
+```
+
+The jupyterlab instances require mounting nfs pvcs. This required
+nfs-common drivers be installed.
```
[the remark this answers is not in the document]
costrouc: changes made.

docs/docs/development.md
```diff
@@ -0,0 +1,152 @@
+# Testing
+
+## Local Testing
+
+Local testing is a great way to test the components of QHub. While it
+does test most of everything is does not test cloud provisioned
```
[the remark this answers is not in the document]
costrouc: changes made.

costrouc:
Yes the deployment will not complete without a proper ip range. I added more detailed notes about what the specific ip range needs to be. https://github.com/Quansight/qhub-cloud/blob/db1b3a4b1c9aac53221a099afcacfe6f643c7658/docs/docs/development.md

> Configure `metallb` to have a start ip of `172.17.10.100` and end ip
> of `172.17.10.100`. These ips were not randomly chosen. You must make
> sure that the ip range is within the docker interface subnet. You can
> see your docker subnet via

```shell
$ ip route
default via 192.168.1.1 dev wlp4s0 proto dhcp metric 600 
172.17.0.0/16 dev docker0 proto kernel scope link src 172.17.0.1 
```

> This means that you just have to ensure that the ip range is within
> the `172.17.0.0/16` subnet. Your docker subnet may be different. You
> can run `metallb` manually as shown bellow or use the python command
> shown bellow. We suggest using the values we suggested since there is
> a dns name that already points to the address.

costrouc:
Had to rebase to resolve conflicts. Only reason it looks like there are many new commits.

docs/docs/development.md
```diff
@@ -1,6 +1,24 @@
...
+This guide assumes that you have the QHub Cloud repository downloaded
+and you are at the root of the repository.
+
+### Dependencies
+
+In order to develop with QHub you will need to download the following
+dependencies and have them available in your path.
+
+ - [terraform](https://www.terraform.io/downloads.html)
+ - [minikube](https://v1-18.docs.kubernetes.io/docs/tasks/tools/install-minikube/)
```
laisbsc, line 19: As a newbie, I need a bit more detail to understand this part.
It seems like to have MiniKube I will need to have Docker installed. Do I also need Docker Engine?

What do you recommend for the VMs? KVM? VirtualBox? Are VMs even necessary at all?

docs/docs/development.md
```diff
@@ -63,7 +63,7 @@ mkdir -p data
 Initialize the `qhub-config.yaml`
 
 ```shell
-python -m qhub init local --project=thisisatest  --domain qhub.test --auth-provider=password --terraform-state=local
+python -m qhub init local --project=thisisatest  --domain github-actions.qhub.dev --auth-provider=password --terraform-state=local
 ```
 
 Give a password to the default user. For this the example password is
```
laisbsc, line 102: Where do I import this? Is using `bcrypt` mandatory or can I choose a personal password?

docs/docs/development.md
```diff
@@ -35,8 +35,8 @@ filename = os.path.expanduser('~/.minikube/profiles/minikube/config.json')
 with open(filename) as f:
```
laisbsc, line 68: How do I run this? Simple copy and paste on the terminal?

docs/docs/development.md
```diff
@@ -91,23 +91,29 @@ python -m qhub render --config qhub-config.yaml -f
...
+`172.17.10.100` from the previous commands. This can be done in many
 ways possibly the easiest is modifying `/etc/hosts` and adding the
 following line. This will override any dns server.
 
 ```ini
-172.17.1.100 jupyter.qhub.test
+172.17.10.100 jupyter.github-actions.qhub.dev
 ```
 
 Finally if everything is set properly you should be able to curl the
```
laisbsc, line 141:
```suggestion
Finally, if everything is set properly you should be able to `curl` the JupyterHub Server.
```

laisbsc (commented):
A few tiny things that were not 100% clear for me.

docs/docs/development.md
aktech: You can install `bcrypt` with pip (https://pypi.org/project/bcrypt/). I think you should be able to choose a personal password.

docs/docs/development.md
aktech: Yes, simply paste in terminal

docs/docs/development.md
laisbsc: Also, the starting point was not 100% clear for me.

Do I start on a clean virtual env in an empty folder, install minikube and then run minikube commands, use `python -m qhub init` to generate the initial config file, and go from there?

aktech: > It seems like to have MiniKube I will need to have Docker installed. Do I also need Docker Engine?

In general not necessarily, you can use hyperkit as well, but the instructions for that might change a lot so I would recommend you have docker installed.

> What do you recommend for the VMs? KVM? VirtualBox? Are VMs even necessary at all?

Docker will take care of that, when you run the minikube start command as mentioned below

laisbsc: Alright. So Docker is a necessary requirement so VMs are not. Understood! 🎖️ 
For now, I will stay away from HyperKit but will have a look at it. Thanks :)

docs/docs/development.md
laisbsc: Alright, I got that bcrypt is just a python package that will help composing the password.
My question is **where to import it?**. The user is asked to import the package but not told where to do it. It is not clear for me if I should use `bcrypt` on my own and copy and paste the generated password o n the CLI command or if I have to import the package somewhere in QHub to have that working.

Sorry the highlighting is not helping more.

laisbsc:
One more thing about the [docs](https://github.com/Quansight/qhub-cloud/pull/287/files/c69379c13a23431ffabe30331329960aee8b8739..0080279c63bb218b618de5f71c5d97e5d96d0a70#diff-4fe08b88edc98df611722319b6584c9c99ba3429bfdd87f8cf7320f1f2801d78).

The starting point was not 100% clear for me. Just trying to figure out the best workflow.

Do I start on a clean virtual env in an empty folder, install minikube and then run minikube commands, use `python -m qhub init` to generate the initial config file, and go from there?

tylerpotts:
commited some minor edits to the docs

docs/docs/development.md
```diff
@@ -0,0 +1,169 @@
...
+The jupyterlab instances require mounting nfs pvcs. This requires
+nfs-common drivers be installed on the nodes themselves.
+
+```shell
+minikube ssh "sudo apt update; sudo apt install nfs-common -y"
+```
+
+Configure the `metallb` load balancer to have a start ip of
+`172.17.10.100` and an end ip of `172.17.10.200`. These ips were not
+randomly chosen. You must make sure that the ip range is within the
```
aktech, line 39: > These ips were not randomly chosen.

I think it would be good to document, the motivation behind choosing this IP Range, since the doc says it was not randomly chosen, but doesn't mentions the reason behind it.

tylerpotts:
Per @aktech :

Docker on mac doesn't support routing traffic to docker containers. This PR will only enable local testing for Linux. 

https://docs.docker.com/docker-for-mac/networking/#known-limitations-use-cases-and-workarounds

Amit will spend a small amount of time debugging, but otherwise we'll only support local testing with Linux.

tylerpotts:
To develop on a remote system with Linux and access the jupyterhub login, utilize this ssh command:

`ssh <user>@<ip> -N -f -L localhost:8000:<docker_ip>:8000`

aktech (approved):
I did a quick test with hyperkit driver for minikube on Mac, but I couldn't got it working in the boxed time. I think since this works well on Linux and we can develop on Linux, let's merge it now, we can anyways add instructions for Mac later. I have added a note for Mac.

I'll merge this in next few hours, unless someone has any objections.

laisbsc:
> I did a quick test with hyperkit driver for minikube on Mac, but I couldn't got it working in the boxed time. I think since this works well on Linux and we can develop on Linux, let's merge it now, we can anyways add instructions for Mac later. I have added a note for Mac.

Sure. I will add the comments about Linux compatibility only to the new docs. Thanks for testing it out.

> I'll merge this in the next few hours, unless someone has any objections.

Please DO merge ASAP. Too many open PRS, it's driving me nuts. 😅

PR #288, opened 2021-02-01
Fix prefect cookiecutter check

### LINKED ISSUE #286: Terraform throws an error if prefect_token variable not set
## Describe the bug
Terraform throws an error if prefect_token variable not set

aktech:
Fixes #286 

The jinja boolean check syntax was wrong. This fixes it.


tylerpotts (approved):
LGTM

PR #302, opened 2021-02-04
Add contribution guidelines

trallard:
This PR adds the Contribution guidelines for qhub

Note this is WIP and will wait also until we have a release strategy to ensure the workflow is correct


danlester (approved):
Thank you for this - I know you said it's not necessarily finished, but definitely better than nothing and can be improved over time!

PR #318, opened 2021-02-15
Update to jupyterhub 0.11.1

brl0:
I have tested with GitHub but not with Auth0.


brl0:
This is related to this PR: https://github.com/Quansight/qhub-terraform-modules/pull/38

costrouc:
@brl0 going to close this PR since we have moved toward building our own jupyterhub image and would require another approach. The motivation was to have tigher control over the packages available in the docker image and keep the size small. That said we are using the latest version of jupyterhub so the goal of this PR should be already done.

PR #319, opened 2021-02-17
Update docker images

brl0:
- Update docker images to ubuntu:20.04
- Use miniforge rather than miniconda
- Push docker images to both github sha tag and latest tag
- Add common tools to jupyterlab image
- Update to jupyterlab 3

### DIFF
--- a/qhub/template/{{ cookiecutter.repo_directory }}/image/jupyterlab/environment.yaml
+++ b/qhub/template/{{ cookiecutter.repo_directory }}/image/jupyterlab/environment.yaml
@@ -2,39 +2,58 @@ name: base
 channels:
   - conda-forge
 dependencies:
- # general
- - pip
+  # general
+  - pip
+  - setuptools
 
- # jupyterhub/jupyterlab
- - nb_conda_kernels
- - ipython > 7
- - jupyter-server-proxy
- - jupyterlab-git
- - ipywidgets
- - jupyterlab >= 2
- - jupyter_client
- - jupyter_console
- - jupyterhub
- - nodejs >= 10
- - nbconvert
- - nbval
- - dask_labextension
+  # jupyterhub/jupyterlab
+  - nb_conda_kernels
+  - ipython > 7
+  - jupyter-server-proxy
+  - jupyterlab-git
+  - ipywidgets
+  - jupyterlab >= 3
+  - jupyter_client
+  - jupyter_console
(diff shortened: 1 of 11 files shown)


qhub/template/{{ cookiecutter.repo_directory }}/image/jupyterlab/environment.yaml
```diff
@@ -2,39 +2,58 @@ name: base
...
- # testing, docs, linting
- - pytest
- - hypothesis
- - flake8
- - sphinx
- - pytest-cov
- - black
- - isort
- - importnb
+  # viz tools
```
trallard, line 30: Genuine question: why are we not pinning viz dependencies but Jlab/Dask ones are?

brl0: Good question.  I did not really put much thought into what was pinned, I pretty much just updated versions for things that were previously pinned.  
I added dask as an explicit dependency to the jupyterlab environment because it was being installed indirectly anyway, and it needed to be pinned for compatibility with the version of dask-gateway qhub is currently using.  The dask pins will need to be updated or preferrably removed once dask-gateway is updated.

trallard: Gotcha - I totally see the importance of pinning deps for compatibility. 

Though it would be good to also pin the viz and update periodically to breaking changes to creep in silently for our users

dharhas: dask and distributed are things we really need to pin since mismatched in those between jlab node and dask workers has been the cause of many subtle and not so subtle bugs in using dask. We really want to handle pinning through a qhub metapackage though https://github.com/Quansight/qhub-cloud/issues/86

tylerpotts:
@brl0 does this need to stay open, or does the merged PR resolve this?

brl0:
@tylerpotts, I think that most of the changes I mentioned in my comment above would still be beneficial, although there is a bit of work needed to update from the main branch since many of these files have been changed in the meantime.

I've been meaning to revisit this sometime soon.  I'll leave it up to you whether it should be closed or not.

brl0:
Closing since this is so outdated.

Issue #349, opened 2021-03-01
Remove `costrouc` as the default user in qhub


costrouc:
I think a sane default would be the one we use in onprem. This is `example-user` as the username.

This will involve modifying:

1. https://github.com/Quansight/qhub-cloud/blob/main/qhub/initialize.py#L22-L26

We need to change the docs to use to create a password `example-user` also you will need to modify the docs anywhere they mention the `costrouc` user. Which may be in a few locations.

https://github.com/Quansight/qhub-cloud/blob/dev/docs/source/06_developers_contrib_guide/04_local_development.md#deploy-qhub

costrouc:
Closing as resolved by @viniciusdc's PR

PR #352, opened 2021-03-01
Remove `costrouc` as the default user in qhub

viniciusdc:
To follow the standard default username we currently use in onprem, the original `costrouc` was modified to `example-user` as well as it's correlated mentions in the docs. The default password fingerprint in the [docs](https://github.com/Quansight/qhub-cloud/blob/dev/docs/source/06_developers_contrib_guide/04_local_development.md#initialize-kubernetes-cluster) was also modified to match the default username.

#349


costrouc (approved):
Looks great to me!

PR #383, opened 2021-03-05
Remove default `jupyter.<domain>` from url and use domain

costrouc:


.github/workflows/test.yaml
```diff
@@ -145,7 +145,8 @@ jobs:
         run: |
           mkdir -p local-deployment
           cd local-deployment
-          qhub init local --project=thisisatest  --domain github-actions.qhub.dev --auth-provider=password --terraform-state=local
+          qhub init local --project=thisisatest  --domain jupyter.github-actions.qhub.dev --auth-provider=password --terraform-state=local
```
aktech, line 148: Do we need `jupyter.` in here?

costrouc: No we don't I only did this for consistency. Should we just use `github-actions.qub.dev` so as to set the example that `jupyter` is no longer required?

aktech: I think that will be great.

aktech:
@costrouc Thanks a lot for this!

PR #399, opened 2021-03-10
fix docs

Adam-D-Lewis:
Fixes example docs to be consistent with recent parameter name changes in dev


tylerpotts (approved):
LGTM

laisbsc (approved):
Wonderful!
Now, I think this should also be changed on the [usage docs](https://github.com/Quansight/qhub-cloud/blob/dev/docs/source/02_get_started/03_usage.md#cloud-deployment).

Let me know if you can fix it otherwise, I can help :)

Thanks a million for the PR, btw.

PR #408, opened 2021-03-15
Qhub linter

viniciusdc:
This  approach aims to execute a specific file for linting called `linter.py`, the idea behind it leave us with for freedom for parsing the bot messages and communication with the end user, besides can me further enhanced to other use cases, as directly contact within the engineer group of qhub-cloud, when some big problem occurs by setting something as `qhub-linter call admin` or something similar.... 

By now it only utilizes `qhub validate`  but we are able to put even more linting abilities to it in the future, without changing all the process.


viniciusdc:
@costrouc I will add a deploy test here with some response examples :smile:

viniciusdc:
even though, this still needs a test with a cloud deployment. I don't have any account to test with. @aktech @costrouc

qhub/provider/cicd/linter.py
```diff
@@ -0,0 +1,101 @@
...
+import subprocess
+import requests
+
+
+def qhub_validate():
+    # Gather the output of `qhub validate`.
+    print("Validate: info: validating QHub configuration in qhub-config.yaml")
+
+    command = "qhub validate qhub-config.yaml"
+    validate_output = subprocess.run([command], shell=True, capture_output=True)
```
costrouc, line 14: I'd prefer to not use subprocess. See this for catching errors https://github.com/Quansight/conda-store/blob/5bf691068561986756740a0baab1ee035ffbeffb/conda-store-server/conda_store_server/build.py#L216-L220.

PR #442, opened 2021-03-28
Set group permission to shared subdirs

### LINKED ISSUE #268: permissions for sub-directories created in shared directory
Shared directory should be RW for all, which is how the permissions are set.
However, directories created under shared default to RO.  This can cause difficulty when trying to collaborate with others.

viniciusdc:
fix #268 

Whith the SetGID bit enabled, every  file created within the `shared` directory will  be automatically modified to be in the group of the directory (permission level).

### DIFF
--- a/qhub/template/{{ cookiecutter.repo_directory }}/templates/jupyterhub_config.py.j2
+++ b/qhub/template/{{ cookiecutter.repo_directory }}/templates/jupyterhub_config.py.j2
@@ -62,6 +62,7 @@ def qhub_configure_profile(username, safe_username, profile):
        'LD_PRELOAD': 'libnss_wrapper.so',
        'NSS_WRAPPER_PASSWD': '/tmp/passwd',
        'NSS_WRAPPER_GROUP': '/tmp/group',
+       'NB_UMASK': '0002',
        'HOME': '/home/jovyan',
     }
 
--- a/qhub/template/{{ cookiecutter.repo_directory }}/templates/jupyterhub_config.py.j2
+++ b/qhub/template/{{ cookiecutter.repo_directory }}/templates/jupyterhub_config.py.j2
@@ -88,7 +88,7 @@ def qhub_configure_profile(username, safe_username, profile):
                   "chown {uid}:{primary_gid} /mnt/home/{username}",
                   "mkdir -p /mnt/home/shared",
                   "chmod 777 /mnt/home/shared"
-             ] + ["mkdir -p /mnt/home/shared/{groupname} && chmod 770 /mnt/home/shared/{groupname} && chown 0:{gid} /mnt/home/shared/{groupname}".format(groupname=groupname, gid=config['gid']) for groupname, config in QHUB_GROUP_MAPPING.items()]).format(username=safe_username, uid=uid, primary_gid=primary_gid)],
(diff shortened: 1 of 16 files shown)


viniciusdc:
This should solve the problem with the groups permission on `shared`. It need a test though, I was not able to test it with the current status of Qhub #443

costrouc:
Looks to be almost everything we need. I tested this locally and we also need to set https://stackoverflow.com/questions/55060753/how-do-i-set-a-default-umask-in-jupyterhub. I'm going to add to this PR tonight to add umasks.

viniciusdc:
> Looks to be almost everything we need. I tested this locally and we also need to set https://stackoverflow.com/questions/55060753/how-do-i-set-a-default-umask-in-jupyterhub. I'm going to add to this PR tonight to add umasks.

Really, I tested it here and the permissions were there, have I missed something during that (thanks for the reply and test too)

qhub/template/{{ cookiecutter.repo_directory }}/templates/jupyterhub_config.py.j2
```diff
@@ -88,7 +88,7 @@ def qhub_configure_profile(username, safe_username, profile):
                   "chown {uid}:{primary_gid} /mnt/home/{username}",
                   "mkdir -p /mnt/home/shared",
                   "chmod 777 /mnt/home/shared"
```
viniciusdc, line 90:
```suggestion
                  "chmod 2777 /mnt/home/shared"
```
Actually I think that I should have changed this line instead...

costrouc: Nope you got the previous one completely right! I've been working on testing this PR.

viniciusdc: yeah, indeed changing that line will not grant the correct groups permissions for each folder (instead I think root will receive it right?). Nvm, thanks for the reply

costrouc:
So I had to do some extra work to get umasks to work with jupyterlab and update qhub in the process to get this working. See commits:
  - https://github.com/Quansight/qhub-cloud/commit/2ad4310d220de2664525cac9b1d77934b14ca8e9
  - https://github.com/Quansight/qhub-cloud/commit/69a3cd545f75e4bbaadae8442c2ef0c900ed4898

I tested on a digital ocean fresh install and got the following:

image

Going to close this PR. @viniciusdc your PR got this working and thanks for the discussion earlier about umask and setgid.

viniciusdc:
> So I had to do some extra work to get umasks to work with jupyterlab and update qhub in the process to get this working. See commits:
> 
> * [2ad4310](https://github.com/Quansight/qhub-cloud/commit/2ad4310d220de2664525cac9b1d77934b14ca8e9)
> * [69a3cd5](https://github.com/Quansight/qhub-cloud/commit/69a3cd545f75e4bbaadae8442c2ef0c900ed4898)
> 
> I tested on a digital ocean fresh install and got the following:
> 
> image
> 
> Going to close this PR. @viniciusdc your PR got this working and thanks for the discussion earlier about umask and setgid.

Thanks for the feedback @costrouc !

PR #462, opened 2021-04-02
Support for flexible forms of authentication we need support for generic oauth

costrouc:
Closes https://github.com/Quansight/qhub-cloud/issues/450


costrouc:
Will also add support for generic authenticator support

PR #470, opened 2021-04-06
Cypress local

danlester:
Cypress tests in tests.yaml workflow for Minikube. Tests login only.


danlester:
Closing because the git history is messy... Will open another one off the latest main.

PR #472, opened 2021-04-07
cds_hide_user_named_servers available in qhub-config.yaml and true on qhub init

### LINKED ISSUE #461: The user can access a list of other servers [bug]
## Describe the bug

@costrouc mentioned that there doesn't seem to be a situation where we might wish the user to have access to a list of their servers or even give them the option of launching more than one jupyterlab instance. Removing this from the UI would remove all such temptation!

danlester:
Default and recommended settings are now:

```
cdsdashboards:
  enabled: true
  cds_hide_user_dashboard_servers: false
  cds_hide_user_named_servers: true
```

This means that the user doesn't get a 'Named Server' section on their home page, so can only start their single JupyterLab 'My Server'.

Dashboard servers will be displayed.

Please note it is still possible to start a named server through the API, or it is conceivable that one appears if there is a faulty dashboard start (perhaps due to misconfigured QHub), in which case the user will never be able to see that named server. It would only be listed to the admin.

Upgrades to existing QHubs should include the above settings in the qhub-config.yaml.

Closes #461

### DIFF
--- a/qhub/cli/deploy.py
+++ b/qhub/cli/deploy.py
@@ -25,9 +25,7 @@ def create_deploy_subcommand(subparser):
         help="Attempt to automatically provision DNS. For Auth0 is requires environment variables AUTH0_DOMAIN, AUTH0_CLIENTID, AUTH0_CLIENT_SECRET",
     )
     subparser.add_argument(
-        "--disable-prompt",
-        action="store_true",
-        help="Disable human intervention",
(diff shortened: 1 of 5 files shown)


costrouc:
Since the options are already namespaced could we make it shorter by removing the `cds_`?

costrouc (approved):

Issue #454, opened 2021-03-30
Deploy Dashboards for multiple user types including non-jupyterhub users


kcpevey:
## Description
I have a qhub-cloud instance. Separately, I also have Panel app that I'm serving at a specific URL. I'd like to start using CDS dashboards integration to serve my panel app through my existing qhub instance, but I'm not sure there is enough control over user access. 

I have two user types:
User A - gets full access to JupyterHub and the Dask workers and any served apps
User B - get zero access to JuypterHub or Dask workers, and ONLY is allowed to access a served app

We are currently using Auth0 for authentication. 

Can the Panel app be served at any URL?

Is this possible?

dharhas:
@danlester @costrouc would like your thoughts on how we accomplish this.

dharhas:
User C: Access to served app & Dask workers via the app?

kcpevey:
> User C

Yes, we should consider this as well. Though I imagine the next ask will be to add monitoring and control over user's worker usage. In other words, User C is dangerous.

costrouc:
User B and User C will be able to be supported. This issue is more of a documetation issue once the cdsdashboard PR is merged. Jupyterlab profiles can be limited to sets of users/groups https://cloud.qhub.dev/en/latest/source/02_get_started/04_configuration.html#profiles. 

Will mark this issue as documentation required.

danlester:
Actually, I hadn't really noticed that you said the Panel app was "served separately". It would be good to understand the significance of that, and to what extent it relies on the QHub instance. If only for authentication, maybe Auth0 should handle the different users.

danlester:
@kcpevey Is there any way within your Auth0 setup to determine who should be in user A and B?

Otherwise, all new users in the QHub would need to be 'B' to start with (can't start their own JupyterLabs) until explicitly added to a JupyterHub group by an admin using the user interface in JupyterHub.

So it would be ideal if there is a flag of some sort in Auth0 that we can inspect to see if they are already in the correct department or similar so should automatically become 'A' users or not.

kcpevey:
@danlester Thanks. I'll have to look into Auth0 and see what is available.

danlester:
There may well be a groups feature in Auth0, but the question would be whether this can be synced naturally from whatever system they are really using. Otherwise, they might as well manage the groups in JupyterHub, if they would just be going into Auth0 set up only to manage these groups to affect their JupyterHub.

[...]

PR #473, opened 2021-04-07
[WIP] Restrict spawn to users in a group 

danlester:
Would solve part of #454.
So far, can add:

```
cdsdashboards:
  enabled: true
  spawn_allow_group: spawners-group
```

This restricts JupyterLab spawn to only members of spawners-group (a JH group). We need to:

1. Ensure members can be easily added to this group via Auth0 if possible
2. Restrict access to Dask Gateway too
3.


viniciusdc:
Hi @danlester, any thoughts on this?

tonyfast:
we've marked this as Stale. if y'all want to continue working on this respond before 11/22.

Issue #478, opened 2021-04-12
Jupyterlab env solve takes too long [bug]


leej3:
## Describe the bug
The jupyterlab solve is currently causing the CI to time out because the environment solve takes approximately 1 hour.

## How can we help?

I have discussed this briefly with @viniciusdc as this is making #406 harder to work on. @viniciusdc please post any progress on this.

Some ideas we had on this:
+ Some hackiness could potentially allow us to use docker layer caching so that the solve only happens when environment.yaml changes. I'm not sure whether that is desirable because the image that is built is deployed. Perhaps layer caching could be disabled for deployment though? We could deploy only upon releases and the tests would be substantially faster.
+ There are some suggestions [here](https://www.anaconda.com/blog/understanding-and-improving-condas-performance) for speeding up the solve... possible candidates are `conda config --set channel_priority strict` and more exact version pinning. i don't think there's anything there to get this into the range of a reasonable solve time.
+ Another solution might be to use a conda lock file. This could be updated as required from a local solve using the environment.yaml so CI doesn't have to suffer its way through a solve each time it is triggered.
+ Use mamba... I don't know whether a mamba solve is deterministic... it would have to be to be useful for CI testing. My understanding is that it is not an optimal solve though.... which may be something we want... certainly when deploying.

leej3:
I chatted with @costrouc about this. There is currently a PR to reduce the frequency of this build but agreed that the conda lock file would be a nice way to drop the time of the build. So someone would rebuild from environement.yaml, and update the lock file which would result in a much faster solve on CI.

viniciusdc:
> I chatted with @costrouc about this. There is currently a PR to reduce the frequency of this build but agreed that the conda lock file would be a nice way to drop the time of the build. So someone would rebuild from environement.yaml, and update the lock file which would result in a much faster solve on CI.

I agree with that idea too, so are we going to create the necessary env with `conda-lock` and use that directly (for all the necessary dependencies) or are we removing some packages from the env list to maintain their upgradability?

viniciusdc:
@costrouc if it's just to create a conda-lock file and test it with the Jupyterhub docker image, I might be able to test it this afternoon.

leej3:
I'll sort this out... once the environment actually solves on my laptop :)  the lock file will fully specify the environment. Updates will be to the environment.yaml... which will then be used to create a new environment and lock file that can be used by CI. It means there is no solve on the CI so things should run more smoothly.

Issue #56, opened 2020-09-18
Rewrite Guided install script in Python


aktech:
Terraform is installed and it still says terraform is not available.

```bash
$ terraform --version
Terraform v0.12.24

Your version of Terraform is out of date! The latest version
is 0.13.3. You can update by downloading from https://www.terraform.io/downloads.html

$ ./scripts/00-guided-install.sh
Error: Please install Terraform v0.12.24
```

aktech:
Currently the bash script is not working well due to the difference sed/grep on mac/linux and etc.

PR #479, opened 2021-04-12
some updates to docker image build, optionally using conda-lock, consolidate shell setup

leej3:
Fixes https://github.com/Quansight/qhub-cloud/issues/268

Or at least it will once I'm done...


leej3:
Will also fix #478 and #56. Perhaps a refactor of the commits is warranted. I just scrambled to get something working. Overall the problems I was attempting to solve were:

+ Improve dockerfile caching behavior (cache invalidations were slowing me down repeatedly so I improved the logic). It's a more verbose dockerfile because files are copied in one by one but a single change in the scripts directory now will not automatically invalidate all layer caches.
+ Conda-lock is now used if docker build is run with `--build-arg SKIP_CONDA_SOLVE=yes`. The lock file installs the environment with exact pins and skips the hour long solve. 
+ Default shell behavior is fixed. This is described in #56 ... in the cluster /home/jovyan is mounted over with the users own data on the NFS share. This means that configuration for the user (and even some software) that is installed into the home directory in the Dockerfile is not available to users during deployment.

I haven't yet fixed all the various files that are inappropriately written to /home/jovyan

@costrouc, /etc/profile (which sources /etc/profile.d/*sh) is fragile because the shell will not necessarily be a login shell. I have additionally tweaked /etc/bash.bashrc and they now reciprocally source each other (with the infinite loop broken appropriately). I'm hoping this results in a robust default shell setup. Thoughts are welcom for something more robust/tidier though.

leej3:
Some minor edits to fix-permissions script and then i think this is roughly done.

And passing the tests would be good.

leej3:
@costrouc I'm happy with this now.

leej3:
And don't tell anyone I changed the user.... there may be some docs that need to be changed...

PR #490, opened 2021-04-14
Cypress full tests from main

danlester:
Cypress tests in tests.yaml workflow for Minikube. Tests login only.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -7,7 +7,6 @@ on:
   push:
     branches:
       - main
-      - cypress-main
 
 jobs:
   test-general:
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yaml
```diff
@@ -154,7 +153,34 @@ jobs:
           nslookup github-actions.qhub.dev
       - name: Curl jupyterhub login page
         run: |
-          curl -k https://github-actions.qhub.dev/hub/home -i
+          curl -k --retry 5 --retry-connrefused https://github-actions.qhub.dev/hub/login
```
costrouc, line 156: `curl -k https://github-actions.qhub.dev/hub/home -i` should now work. Retry is no longer needed

.github/workflows/test.yaml
```diff
@@ -7,6 +7,7 @@ on:
   push:
     branches:
       - main
+      - cypress-main
```
costrouc, line 10: Assume this was only for testing? And can be removed?

costrouc (commented):
Just those two changes and after that looks good to me. Looking forward to having the web UI testing.

danlester:
Great, thanks for your comments - have accepted all and I will just merge.

PR #501, opened 2021-04-19
Force 'git init' branch to be main

danlester:
Small change to "git checkout -b main" after "git init". This is to overcome some problems seen in auto-provisioned GitHub repos where the branch is really master... or do we actually still want 'master' for new repos anyway?


costrouc (approved):
Nice addition :+1:

PR #503, opened 2021-04-19
AWS_DEFAULT_REGION respected, and availibility zones not needed

danlester:
AWS_DEFAULT_REGION was always set as a GitHub secret when `qhub init` auto-provisions a GitHub repo, but the region was never set in the qhub-config.yaml file. It was always left as a hardcoded us-west-2.

Additionally, availability_zones of us-west-2a and us-west-2b were hardcoded.

This PR:
-  respects AWS_DEFAULT_REGION and inserts that into qhub-config.yaml
- no longer requires availability_zones to be set in qhub-config.yaml
- if qhub-config.yaml does contain availability_zones those will be used

Note that availability_zones is still allowed in qhub-config.yaml:
1. for backwards-compatibility
2. in case the user does have a preference

However, I do not believe it is possible to use `qhub deploy` to migrate an existing qhub to newly-specified availability zones (or regions, for that matter?)

This new approach could therefore cause problems if AWS starts to return different availability zones, or even just in a different order. (Maybe there is a known solution in this situation - I haven't studied it yet.)

If this PR is approved, we need to create similar issues for the other clouds if they have similar data sources available.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -72,13 +72,13 @@ jobs:
         if: ${{ matrix.provider != 'local' }}
         uses: hashicorp/terraform-github-actions@master
         with:
-          tf_actions_version: 0.13.5
+          tf_actions_version: 0.14.9
           tf_actions_subcommand: 'fmt'
           tf_actions_working_dir: 'qhub-${{ matrix.provider }}-deployment/terraform-state'
       - name: 'Terraform Format'
         uses: hashicorp/terraform-github-actions@master
         with:
-          tf_actions_version: 0.13.5
+          tf_actions_version: 0.14.9
           tf_actions_subcommand: 'fmt'
           tf_actions_working_dir: 'qhub-${{ matrix.provider }}-deployment/infrastructure'
       - name: QHub Render Artifact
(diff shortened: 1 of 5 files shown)


danlester:
By the way, I really don't understand why `terraform fmt` is throwing this error, or rather why Jinja2 is dropping that whitespace!

costrouc:
> This new approach could therefore cause problems if AWS starts to return different availability zones, or even just in a different order. (Maybe there is a known solution in this situation - I haven't studied it yet.)

I think the way we can mitigate this is to sort the availability zones returned by AWS. We can't have deployed on zones that does not already exist.

>  However, I do not believe it is possible to use qhub deploy to migrate an existing qhub to newly-specified availability zones (or regions, for that matter?)

Terraform will do this i believe which will in the process destroy nearly everything and recreate it.

costrouc:
> By the way, I really don't understand why terraform fmt is throwing this error, or rather why Jinja2 is dropping that whitespace!

It had to do with cookiecutter. I was not removing the newline e.g.

```
a
{% if cookiecutter.b is defined %}
{{ b }}
{% endif %}
d
```

Will render if b=1 is defined to

```
a

1

c
```

Preserving the newline.

costrouc (approved):
Looks good to me. We certainly need to do this for other providers. Eventually I'd like to make the kubernetes version optional as well.

Issue #391, opened 2021-03-08
Allow users access to lifecycle hooks in kubespawner in qhub-config.yaml


tylerpotts:
## Description
@rsignell-usgs and @jkellndorfer Frequently work with teams where they would like to configure the home directory of new users (change `.profile` to add environment variables for example.) Currently the only way to do this on AWS is to manually create a kuberenetes pod with the shared drive attached and edit the files via terminal.

## Suggestion
Have an admin user group

tylerpotts:
Tagging @costrouc for his opinions

costrouc:
> Frequently work with teams where they would like to configure the home directory of new users (change .profile to add environment variables for example.) 

This can be done via pre exec hooks on jupyterlab start. `c.KubeSpawner.lifecycle_hooks` option in https://jupyterhub-kubespawner.readthedocs.io/en/latest/spawner.html. We currently do this here https://github.com/Quansight/qhub-cloud/blob/34074ce063d702e241f6bf412d195b708fac27b6/qhub/template/%7B%7B%20cookiecutter.repo_directory%20%7D%7D/templates/jupyterhub_config.py.j2#L69-L79. But I agree that we need to have an issue for allowing users more customization of the lifecycle hooks.

Renaming this issue to reflect the feature request and moving to backlog.

tylerpotts:
Potential ideas:

- Expose NFS filesystem behind a password protected web UI (IE freeNAS)

danlester:
A section in qhub-config.yaml to list files to be copied.
At render/deploy stage copy these to /etc/jupyterhub/homefiles or something
In Dockerfile for our Jupyterlab environment, insert a script to copy anything in /etc/jupyterhub/homefiles to the user's home folder.
This Dockerfile script is run somewhere in or around the start.sh script.

danlester:
Trying out a proof-of-concept of this - let me know if that's treading on anyone's toes.

danlester:
I've made some progress in the userskel branch. Will go through it with someone who knows more about permissions via nfs.

Also need to tidy up jupyterhub_config.py.j2 (some unused functions), but maybe worth waiting for helm/z2jh to be removed first.

PR #521, opened 2021-04-23
Create shared/.userskel folder to be copied to users' home

danlester:
There will be a new shared/.userskel folder ('user skeleton').

The contents of this will be copied to any user's home folder on spawn of a server (but not overwriting any existing files).

.userskel is admin-writable but only readable by other users.

This PR is a solution, at least for the inspiration behind, #391

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,7 +27,7 @@ gcp-jupyterhub
 modules
 quansight-terraform-modules
 config.yaml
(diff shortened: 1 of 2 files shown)


danlester:
If we like this feature it needs some documentation.

tylerpotts (approved):

PR #522, opened 2021-04-23
update docker images

leej3:
Supercedes #479


costrouc (approved):
Awesome work @leej3! I'm going to make sure these images make it into the next Qhub release.

PR #523, opened 2021-04-23
Remore ref. to missing doc

### LINKED ISSUE #516: Reference to missing file [documentation]
The deploy guides in `qhub/docs/source/04_how_to_guides` mention `./scripts/00-guided-install.sh` which doesn't exist in the repo.  Not sure what this should be replaced with.  @laisbsc do you happen to know?

laisbsc:
closes #516


laisbsc:
🙄 I feel like I will never understand git sometimes...

PR #525, opened 2021-04-23
Update 04_configuration.md

tylerpotts:
I had to guess and use previous knowledge to know to copy the url from the hyperlinked text for the Let's Encrypt server. This makes things a bit more explicit with the let's encrypt example.


laisbsc (approved):
Looks good to me. Thanks for the input @tylerpotts. All good to go!

PR #526, opened 2021-04-23
Node group fix and Pin k8s for azure

Adam-D-Lewis:
- Pins k8s version for azure deployment
- renames node resource group to enable multiple qhub deployments per azure account


Adam-D-Lewis:
Closes https://github.com/quansight/qhub/issues/524

costrouc:
Awesome thanks for fixing the issue with multiple azure deployments @Adam-D-Lewis!

PR #529, opened 2021-04-26
Improvements to qhub destroy and terraform-state sync

danlester:
To improve `qhub destroy`:

- Render terraform files before destroy (like deploy) by default
- Assume terraform-state/terraform.tfstate is old or not present (for remote backends)
- Import state from cloud

This is especially useful for integration tests because the state buckets are not currently being destroyed, making it difficult to start the test next time round.

AWS/GCP now working. Attempt at Digital Ocean import but fails with:
```
[terraform]: module.terraform-state.module.spaces.digitalocean_spaces_bucket.main: Importing from ID "qhubintgrtndo-dev-terraform-state"...
[terraform]: module.terraform-state.module.spaces.digitalocean_spaces_bucket.main: Import prepared!
[terraform]:   Prepared digitalocean_spaces_bucket for import
[terraform]: module.terraform-state.module.spaces.digitalocean_spaces_bucket.main: Refreshing state... [id=qhubintgrtndo-dev-terraform-state]

Error: error reading Spaces bucket "qhubintgrtndo-dev-terraform-state": RequestError: send request failed
caused by: Head "https://qhubintgrtndo-dev-terraform-state..digitaloceanspaces.com/": dial tcp: lookup qhubintgrtndo-dev-terraform-state..digitaloceanspaces.com: no such host
```

This looks as though the region (e.g. nyc3) is not being inserted into the URL.

Azure  - no attempt yet

### DIFF
--- a/qhub/cli/destroy.py
+++ b/qhub/cli/destroy.py
@@ -5,6 +5,7 @@ from ruamel import yaml
 
 from qhub.destroy import destroy_configuration
 from qhub.schema import verify
+from qhub.render import render_default_template, render_template
 
 logger = logging.getLogger(__name__)
 
@@ -12,6 +13,13 @@ logger = logging.getLogger(__name__)
 def create_destroy_subcommand(subparser):
     subparser = subparser.add_parser("destroy")
     subparser.add_argument("-c", "--config", help="qhub configuration", required=True)
+    subparser.add_argument("-i", "--input", help="input directory")
+    subparser.add_argument("-o", "--output", default="./", help="output directory")
+    subparser.add_argument(
+        "--disable-render",
+        action="store_true",
+        help="Disable auto-rendering before destroy",
+    )
     subparser.set_defaults(func=handle_destroy)
 
 
@@ -26,5 +34,11 @@ def handle_destroy(args):
         config = yaml.safe_load(f.read())
 
     verify(config)
+
+    if not args.disable_render:
+        if args.input is None:
+            render_default_template(args.output, args.config, force=True)
+        else:
+            render_template(args.input, args.output, args.config, force=True)
 
     destroy_configuration(config)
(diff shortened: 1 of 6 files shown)


danlester:
I have submitted an issue with the Digital Ocean provider: https://github.com/digitalocean/terraform-provider-digitalocean/issues/627

danlester:
Now complete, requires qhub-terraform-modules PR to be used: [Set terraform-state buckets to force_destroy=true](https://github.com/Quansight/qhub-terraform-modules/pull/55) 

The idea is to always attempt to import remote state of terraform-state module before doing anything (deploy or destroy).

--skip-remote-state-provision will still skip this syncing but if it is omitted then `qhub deploy` should be able to create the remote buckets etc if they don't exist, and simply import afresh and (probably) do nothing in `terraform apply` if they do exist. It always removes terraform-state/terraform.tfstate before doing the import.

When running `terraform import` it strips the color red from the output (because bold red 'Error: import resource already exists' is worrying), and always catches any non-zero exit code presuming that the error was because the import failed due to the resource not yet existing.

BTW verify_configuration_file_exists function wasn't really doing anything so removed it - was checking for the existence of a file in cwd called 'qhub-config.yaml' regardless of the provided input filename, and in any case always after the file was imported.

danlester:
Will run the qhub-integration-tests once this is merged.

costrouc (approved):

costrouc (approved):
@danlester thanks for going through the hard work on making our integration tests more repeatable. I also got to learn about [terraform import](https://www.terraform.io/docs/cli/import/index.html) which I previously didn't know about and think I will be using in the future! This looks like a great way to "find" resources that were created by terraform but lost the state.

danlester:
@costrouc Thanks for taking a look. Just to reiterate, this also performs the 'sync' of terraform-state before deploy by default, so it's not essential to use --skip-remote-state-provision in CI (but saves some time, and of course must not be used if there is a chance it could be running for the very first time).

PR #537, opened 2021-04-28
Set conda_envs for dashboards

### LINKED ISSUE #530: Allow selection of cdsdashboards conda_envs if >1 environment in qhub-config.yaml
In the jupyterhub_config.py.j2 template, we need to supply the conda_envs list for cdsdashboards:

https://cdsdashboards.readthedocs.io/en/stable/chapters/customization/huboptions.html#conda-environments

danlester:
Sets c.CDSDashboardsConfig.conda_envs in jupyterhub_config so that the user can choose the Conda Env when creating/editing the dashboard.

Closes #530

### DIFF
--- a/qhub/template/{{ cookiecutter.repo_directory }}/templates/jupyterhub_config.py.j2
+++ b/qhub/template/{{ cookiecutter.repo_directory }}/templates/jupyterhub_config.py.j2
@@ -15,6 +15,12 @@ c.JupyterHub.default_url = '/hub/home'
 # Force dashboard creator to select an instance size
 c.CDSDashboardsConfig.spawn_default_options = False
 
+c.CDSDashboardsConfig.conda_envs = [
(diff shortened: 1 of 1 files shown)


tylerpotts (approved):
LGTM

costrouc:
@danlester is this all that is needed? Does cdsdashboards need to know the path to the conda environment? It is at a non-standard location `/home/conda/environments/<name>`

danlester:
@costrouc It seems to find it if that's good enough :)

It is possible to use the full path there if required, but I presume some config is set in our images (or maybe some logic in jhsingle-native-proxy) to find these conda envs. Happy to look into it if it's a concern, or anyone comes across problems.

costrouc:
> @costrouc It seems to find it if that's good enough :)

Good enough with me :smile:

PR #543, opened 2021-04-29
Versioning and docs improvements

### LINKED ISSUE #429: Co-ordinate Docker image tags for upgrades (and qhub init)
New packages may be added to the Dockerfile and environment.yaml files during development, e.g. [here](https://github.com/Quansight/qhub-cloud/blob/d45f4c6168742367cd34538599a0fed14f783ceb/qhub/template/%7B%7B%20cookiecutter.repo_directory%20%7D%7D/image/Dockerfile.jupyterhub)

When it comes to release (and also during dev) the default docker images in use are referenced by hard-coded tags [here](https://github.com/Quansight/qhub-cloud/blob/d45f4c6168742367cd34538599a0fed14f783ceb/qhub/initialize.py#L32).

These tags need to be updated automatically before release as per #428 (which currently asks for it to be done manually).

One solution (used in cdsdashboards repo) is for Docker images to be built when the repo is tagged with a version number on main. The code itself already knows the current version number of qhub, which can be stored in a version.py file etc. So the code (and also docs) can know what tag should be used dynamically - it matches the version number if git tagging is done correctly.

However, this does not help existing installations which will have images directly referenced in existing qhub-config.yaml files. These may or may not be completely bespoke images per-deployment. The problem is that version numbers (e.g. for dask gateway) may not match across all locations - qhub/terraform, jupyterhub image, jupyterlab image etc.

danlester:
Now qhub/version.py contains the only reference to a version number (e.g. 0.3.6)

- terraform_modules: ref is now assumed to be `release-0.3.6`
- Docker images are assumed to be tagged `v0.3.6` in the default config, e.g. `quansight/qhub-jupyterhub:v0.3.6`
- The documentation refers to images by this tag
- GitHub/GitLab CI explicitly install that version of qhub

In practice, these will only exist:
- If the new simplified release guidelines have been followed to ensure correct tag names; and
- Because the qhub repo's GitHub workflow has been updated to tag images based on git tags

In addition, Docker images are given multiple tags - based on git sha, branch name, and tag (if any). So `quansight/qhub-jupyterhub:main` should refer to the latest images built from the contents of the main branch.

While this should make releases easier and much more predictable, it may make dev environments slightly harder to set up, for example always having to add a `terraform_modules: rev: main` override. There should perhaps be a way to override this via env var.

Closes #429


danlester:
Note qhub/VERSION has been removed.
In README.md, I have dropped the version number as displayed in the qhub CLI help message - its absence is unlikely to confuse anyone, and it saves changing it pre-release. The main docs do show the correct version number here.

danlester:
There are some ideas to make dev easier in issue #545 

One policy we might need to set is exactly when version numbers are incremented. e.g. do we upgrade main branch's version.py to 0.3.7-dev1 immediately after 0.3.7 release?

costrouc:
@danlester is it intentional that this PR is pushing to docker hub? 

image

danlester:
Now also checks QHUB_GH_BRANCH env var to produce development-friendly qhub-config.yaml files and terraform templates.

Closes #545

danlester:
Now also addresses #558 
test.yaml workflow should not set QHUB_GH_BRANCH for a tag

Closes #558

docs/ext/substitute.py
```diff
@@ -0,0 +1,16 @@
+"""
```
costrouc, line 1: Love this. Nice and clean way to add substitution in sphinx :smile:

costrouc (approved):

Issue #448, opened 2021-03-29
Timeout upon deploy [bug]


leej3:
## Describe the bug
When launching a cluster for the first time the deployment times out

## How can we help?

* What are you trying to achieve?
This is for the local dev deployment with a minikube cluster. The context is an inadequate internet connection. All of the containers cannot be downloaded quickly enough before the kubernetes time out.

* How can we reproduce the problem?
Using a minikube cluster (as part of the (developers contribution guide)[https://cloud.qhub.dev/en/latest/source/06_developers_contrib_guide/04_tests.html]) launch a cluster using `qhub deploy`.

* What is the expected behaviour?
Success and happiness

* And what is currently happening?

Not success or happiness... a timeout

```

INFO:qhub.deploy:All qhub endpoints will be under https://github-actions.qhub.dev
INFO:qhub.provider.terraform:terraform=/tmp/terraform/0.13.5/terraform init directory=infrastructure
INFO:qhub.provider.terraform:terraform init took 2.259 [s]
INFO:qhub.provider.terraform:terraform=/tmp/terraform/0.13.5/terraform apply directory=infrastructure targets=['module.kubernetes', 'module.kubernetes-initialization', 'module.kubernetes-ingress']
INFO:qhub.provider.terraform:terraform apply took 7.698 [s]
INFO:qhub.provider.terraform:terraform=/tmp/terraform/0.13.5/terraform output directory=infrastructure
INFO:qhub.provider.terraform:terraform output took 1.536 [s]
INFO:qhub.provider.terraform:terraform=/tmp/terraform/0.13.5/terraform apply directory=infrastructure targets=[]

Error: timed out waiting for the condition

  on .terraform/modules/qhub/modules/kubernetes/services/dask-gateway/main.tf line 21, in resource "helm_release" "dask-gateway":
  21: resource "helm_release" "dask-gateway" {


ERROR:qhub.deploy:b'\x1b[0m\x1b[1mmodule.kubernetes-initialization.kubernetes_namespace.main: Refreshing state... [id=new]\x1b[0m\n\x1b[0m\x1b[1mmodule.kubernetes-ingress.kubernetes_service.main: Refreshing state...
```

* Any error messages?
> If helpful, please add any screenshots and relevant links to the description.

## Your environment
Describe the environment in which you are experiencing the bug.

Debian derivative, x86, docker deployment as described in the developer contribution guide linked above.

## The fix?

Currently it seems that repeating the execution of the deployment command does eventually get there... not ideal but it works!

PR #549, opened 2021-04-30
[WIP] use bash for postStart in jupyterlab

### LINKED ISSUE #542: Jupyterlab Terminal reporting errors on start
## Describe the bug

When starting a new jupyterlab session the following is being reported. Though I can report that umask is working!

```shell
/bin/sh: 16: /etc/profile.d/term_settings.sh: shopt: not found
/bin/sh: 24: /etc/profile.d/term_settings.sh: shopt: not found
/bin/sh: 58: /etc/profile.d/term_settings.sh: Syntax error: Bad fd number
```

cc: @leej3

leej3:
This is an attempt to fix #542.

The combination of #448 and #548 is making this very difficult for me to test.

Also, currently the change isn't showing up on my local deployment... I imagine
the config script is baked into the docker image. I will explore further...

### DIFF
--- a/qhub/template/{{ cookiecutter.repo_directory }}/templates/jupyterhub_config.py.j2
+++ b/qhub/template/{{ cookiecutter.repo_directory }}/templates/jupyterhub_config.py.j2
@@ -111,7 +111,7 @@ def qhub_configure_profile(username, safe_username, profile):
     profile['kubespawner_override']['lifecycle_hooks'] = {
         "postStart": {
             "exec": {
-                "command": ["/bin/sh", "-c", (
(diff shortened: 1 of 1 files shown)


leej3:
I don't think this is necessary... and I don't think it fixes the problem. Setting SHELL=/bin/bash fixed the problem

costrouc:
Adding things here would have used a command to run `/bin/bash` after the jupyterlab session is started. This is a nice hook if we want to do some actions in parallel right after the user started the lab session.

leej3:
Good to know. Thanks. I’m sure we’ll think of something fun to put there!

PR #565, opened 2021-05-06
Pydantic correct validation, github auto-provision checks

### LINKED ISSUE #539: Yaml validation doesn't trap missing subfields
For example:
```
class Main(Base):
    terraform_modules: typing.Optional[TerraformModules]

class TerraformModules(Base):
    repository: str = "github.com/quansight/qhub-terraform-modules"
    rev: str = "main"
```

If the yaml has repository missing:
```
terraform_modules:
  rev: main
```

the missing subfield is not trapped in validation and we get an incomprehensible error later on. This happens whether or not TerraformModules is optional.

### LINKED ISSUE #557: Failure to autoprovision github repo does not throw error
Running something like this:

```
qhub init azure \
          --domain qhubaz.ideonate.com\
          --project dslqhubaz \
          --auth-provider password \
          --repository dslqhubaz \
          --repository-auto-provision
```

does not create the github repo but does not give a warning or error.( The repo needs the full URL for it to work.)

It should throw an error. Maybe similar issue for Auth0 and GitLab etc.

danlester:
Removed defaults from pydantic schema since they were causing missing values to remain missing, and the defaults were never used anyway - so confusing for developers to see them there. Defaults aren't used because Pydantic is being used for validation but not model initialization.
Closes #539 

GitHub auto-provision - returns errors if provision fails or secrets missing, warns if repo already exists.
Closes #557

### DIFF
--- a/qhub/deploy.py
+++ b/qhub/deploy.py
@@ -48,7 +48,7 @@ def guided_install(
     # don't have `terraform_state` key
     if (
         (not skip_remote_state_provision)
-        and (config.get("terraform_state", {}).get("type") == "remote")
+        and (config.get("terraform_state", {}).get("type", "") == "remote")
         and (config.get("provider") != "local")
     ):
         terraform_state_sync(config)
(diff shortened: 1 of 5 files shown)


costrouc (approved):
Always a fan of better error messages and less confusing default values.

PR #567, opened 2021-05-06
Github Worfkflow templates simplified

### LINKED ISSUE #556: Simplify GitHub Workflows
They currently use terraform explicitly, but can just use qhub deploy like the gitlab workflow does

danlester:
Github workflow only needs deploy now, no separate terraform workflows

Closes #556

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -1,6 +1,8 @@
 name: "Build Docker Images"
 
 on:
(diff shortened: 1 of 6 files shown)


danlester:
Related note about deleting some of our existing workflow files: https://github.com/Quansight/qhub/issues/573

costrouc (approved):

PR #576, opened 2021-05-11
Qhub linter

### LINKED ISSUE #379: Comment on PR upon change to qhub-config.yaml in github-actions
The current linter is run with

https://github.com/Quansight/qhub-cloud/blob/main/qhub/template/%7B%7B%20cookiecutter.repo_directory%20%7D%7D/.github/workflows/qhub-ops.yaml

Also validation is always run on the qhub-config.yaml even when the validate command is not directly called.

Add validation to 

 - github-actions have the following information https://docs.github.com/en/actions/reference/environment-variables

You'll have GITHUB_TOKEN set which you can use for api requests.

You can find the pull request number

https://github.com/actions/checkout/issues/58#issuecomment-545446510

Submit a comment via

This looks usefull https://docs.github.com/en/rest/reference/pulls#create-a-review-comment-for-a-pull-request

viniciusdc:
This approach aims to execute a specific file for linting called `linter.py`. Right now it only utilizes `qhub validate` (`verify(config)`), but we are able to put even more linting abilities to it in the future, without changing all the process.

close #379

### DIFF
--- a/qhub/cli/validate.py
+++ b/qhub/cli/validate.py
@@ -3,11 +3,15 @@ import pathlib
 from ruamel import yaml
 
 from qhub.schema import verify
+from qhub.provider.cicd.linter import comment_on_pr
 
 
 def create_validate_subcommand(subparser):
     subparser = subparser.add_parser("validate")
     subparser.add_argument("config", help="qhub configuration")
+    subparser.add_argument(
+        "--enable-commenting", help="Turn on PR commenting", action="store_true"
+    )
     subparser.set_defaults(func=handle_validate)
 
 
@@ -21,4 +25,8 @@ def handle_validate(args):
     with config_filename.open() as f:
         config = yaml.safe_load(f.read())
 
-    verify(config)
+    if args.enable_commenting:
+        # for PR's only
+        comment_on_pr(config)
+    else:
+        verify(config)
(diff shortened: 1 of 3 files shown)


viniciusdc:
@tylerpotts @dharhas or @costrouc what do you think? I still need to add the environment dependencies check (aka Distributed and Dask packages presence) but this needs more understanding of Pydantic.

danlester:
I've just rebased this to main and force-pushed (which normally I would only do for my own branch - sorry!)

Anyway I'll take a look at this PR now.

danlester (approved):

danlester:
This is great - thanks @viniciusdc

If anyone doesn't need it, maybe we will add an option to qhub-config to turn off this feature, but for now let's see how it helps people.

BTW I made a couple of changes to formatting, especially for the markdown blocks. In fact, we should really 'markdown-escape' them but I don't think we are in too much danger anyway.

viniciusdc:
> This is great - thanks @viniciusdc
> 
> If anyone doesn't need it, maybe we will add an option to qhub-config to turn off this feature, but for now let's see how it helps people.
> 
> BTW I made a couple of changes to formatting, especially for the markdown blocks. In fact, we should really 'markdown-escape' them but I don't think we are in too much danger anyway.

Oh thank you very much @danlester, no problems with the rebase, and thanks for reviewing this. For the Markdown escape, I noticed that too but was something difficult to handle directly in the strings (but I made a function to parse that if we need to use in the future)

PR #577, opened 2021-05-11
Move terraform modules into qhub repo

### LINKED ISSUE #566: Move terraform modules into qhub repo

danlester:
Closes #566 

qhub-terraform-modules taken from [this point](https://github.com/Quansight/qhub-terraform-modules/releases/tag/everythingmoved), and split into infrastructure and terraform-state 'modules' folders in the qhub cookie-cutter template.

@costrouc says he has one client using their own version of qhub-terraform-modules. My suggestion is to release this PR in a full version of qhub, advise the client not to upgrade immediately, and then re-add the option to specify the terraform_modules field in the qhub-config.yaml file to override the template's versions. This will need a slight reorg of the qhub-terraform-modules repo to match the template's version. The point is I don't want to change that repo until this PR is released.


danlester:
k8s tests currently failing because jupyterlab docker image isn't building. Will check if still a problem once merged into main.

pyproject.toml
```diff
@@ -1,8 +1,4 @@
 [tool.black]
 line-length = 88
 target-version = ['py36', 'py37', 'py38']
-exclude = '''
-/(
-  scripts
-)/
-'''
+exclude = '/\{\{ cookiecutter.repo_directory \}\}/'
```
aktech, line 4: What is this change about?

qhub/cli/deploy.py
```diff
@@ -15,7 +15,6 @@ def create_deploy_subcommand(subparser):
     subparser.add_argument(
         "-c", "--config", help="qhub configuration yaml file", required=True
     )
-    subparser.add_argument("-i", "--input", help="input directory")
```
aktech, line 18: Don't recall why this change?

qhub/render/__init__.py
```diff
@@ -92,16 +88,11 @@ def deep_merge(d1, d2):
         return d1
 
 
-def render_default_template(output_directory, config_filename=None, force=False):
+def render_default_template(output_directory, config_filename, force=False):
```
aktech, line 91: if there is only one function for rendering template, then probably worth naming it as just `render_template`?

qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/azure.tf
```diff
@@ -4,15 +4,15 @@ provider "azurerm" {
 
 
 module "registry" {
-  source              = "{{ cookiecutter.terraform_modules.repository }}//modules/azure/registry?ref={{ cookiecutter.terraform_modules.rev }}"
+  source              = "./azure/registry"
```
aktech, line 7: I guess `./azure/registry` -> `./modules/azure/registry`?

qhub/render/__init__.py
```diff
@@ -37,20 +39,14 @@ def patch_dask_gateway_extra_config(config):
...
 
     config["pip_install_qhub"] = pip_install_qhub
 
-    if "terraform_modules" not in config:
-        config["terraform_modules"] = {
-            "repository": "github.com/quansight/qhub-terraform-modules",
-            "rev": QHUB_GH_BRANCH and QHUB_GH_BRANCH or f"release-{__version__}",
-        }
-
     config["QHUB_GH_BRANCH"] = QHUB_GH_BRANCH
```
aktech, line 50: This doesn't seems to be imported, might just be a problem in main branch as well.

aktech (changes requested):
Looks good to me, apart from minor comments

qhub/render/__init__.py
aktech: I see, you've fixed this already.

danlester: Yes, my other PR was already merged, but needed to be merged after this one I think.. anyway I fixed it directly here.

qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/azure.tf
danlester: Well spotted. I will run the integration tests after merge anyway to pick up any issues. I tried Azure, but maybe not with the final version of this branch :)

qhub/render/__init__.py
danlester: Sure, have done now. Shorter name is better!

pyproject.toml
danlester: Just because some Python files are now within the tf modules templates, and they should be excluded from Black formatting. I don't think 'scripts' was needed here as an exclusion.

qhub/cli/deploy.py
danlester: You used to be able to supply an alternative cookie-cutter template location, but we don't believe it was useful or in use at this time, so decided to remove it to keep things simpler.

danlester:
Oh sorry, I marked everything as resolved, but maybe that was for you to do...

aktech (approved):
@danlester Thanks for the changes, I think this looks good to go now.

PR #580, opened 2021-05-13
Add GPU PATHs in the Dockerfiles for GCP

### LINKED ISSUE #572: GPU GCP Support
Opening this issue to track.

4 things here:

 - The bash_profile I suggested is not overwritten by the kubespawner environment as the .bash_profile is created on the post start hook (I have tested it as well)
 - I think its not a good idea to let the user write the PATH and LD lib path on the config because the installation of drivers is done by QHub via nvidia installer daemon set and not by the user, so user should not worry about putting the path to lib and hardcoding the path like that in the config is error prone and can have side effects.
 - This is cloud specific, that's why I suggest putting things in bash profile based on cloud provider
 - The most we could do, may be later is giving the user ability to configure bash_profile via qhub config

How does that sounds?

Good points. I agree with the you that the user should not have to worry about that.
  - I don't think we should ever manage files in the users home directory since this would mean that they could no longer modify that file .bash_profile.
  - I'm fine if we want to bake this logic into the jupyterlab docker image itself. The commit you showed will only work for when you open a terminal in jupyterlab not the ipykernel to my knowledge.  E.g adding a script to /etc/profile.d/...sh. Not sure if there are other approches... I think getting the paths right for the jupyterlab kernel will be tricky since it does not source bash profiles. We may have to modify the jupyter_config.py scriptEither way https://github.com/Quansight/qhub/issues/571 is a bug and should be fixed since right now users can't add any environment variables (so I see 571 as an orthogonal issue)

aktech:
This adds paths to NVIDIA libs in the Dockerfile, for the user to be able to access GPUs via CLI and notebook.

- [x] I have added this irrespective of we use GPUs or not, I don't see a problem with this at the moment as those paths would simply be unused in case the user don't use GPUs.
- [x] Added some docs for usage.
- [x] Checksum update for https://code-server.dev/install.sh, seems like the install script got updated recently

Fixes #572


costrouc:
Thanks @aktech. Have you tested this on GCP?

aktech:
> Thanks @aktech. Have you tested this on GCP?

Yep, I deployed the the internal quansight qhub with this.

costrouc:
Also in the notes could you specify that if they are not using the gcp provider in qhub but are using gcp. They will need to manually specify https://cloud.google.com/kubernetes-engine/docs/how-to/gpus#installing_drivers and run on their cluster.

costrouc:
Otherwise this looks great

costrouc (changes requested):
Requesting to add the docs on kubectl if you are deploying on a kubernetes cluster on gcp but not using the gcp qhub provider

aktech:
> if they are not using the gcp provider in qhub but are using gcp.

What do you mean by this? When they deploy QHub on an existing cluster on GCP?

aktech:
ping @costrouc

costrouc:
I mean the case when they do `qhub init local ...` but they are deploying on an existing gcp cluster. This is the case for AES and thought this would be helpful to have in the docs.

aktech:
Thanks @costrouc added the note about manually installing drivers.

costrouc (approved):

Issue #504, opened 2021-04-20
Regions and Availability Zones for all Clouds


danlester:
We have already fixed AWS as follows in issue #489

When using qhub init e.g. for AWS you would set
export AWS_DEFAULT_REGION=eu-west-2
for example.

That is set faithfully in GitHub secrets, but the qhub-config.yaml had a hardcoded region: us-west-2.

Availability zones also need to relate to the region.

The PR was: https://github.com/Quansight/qhub/pull/503

However, this behavior needs to be replicated on the other cloud providers: GCP, Azure, DO as appropriate.

PR #586, opened 2021-05-14
Cloud Regions simplified

danlester:
AWS_DEFAULT_REGION no longer needed.

For GCP, `zone` is removed from qhub-config.yaml, and availability_zones is optional - pulls default from Google if not specified.

To avoid clusters being torn down inadvertently, once set availability_zones cannot be changed (this was the same for the similar PR I did for AWS).


Adam-D-Lewis:
Looks good to me, @danlester

Adam-D-Lewis (approved):

danlester:
Closes #504

PR #385, opened 2021-03-05
pinning the qhub-dask in qhub-cloud

viniciusdc:
The metapackage approach for `qhub-dask` allow us to update our pinned versions for dask dependencies. fix #86

PR #588, opened 2021-05-18
Pinning the Qhub-dask in Qhub

### LINKED ISSUE #86: Use conda metapackages for environment consistency
We have been facing the problems described at [this medium article](https://medium.com/pangeo/managing-environments-5604d848367a) regarding Management of Virtual Environments.

We already have this problem and are enforcing it manually when creating new conda envs. This is typically the primary reason dask etc fails on new environments etc. The metapackage approach is the only real solution at this point and will also allow us to build a good linter (#119) that can check pull requests.

## Further discussion
There is probably a discussion in terms of whether the conda-store story changes the way we do this. What is the best way to integrate it with conda-store?

viniciusdc:
The metapackage approach for `qhub-dask` allows us to update our pinned versions for dask dependencies. close #86 
See for reference #385 

This is the current package dependencies `qhub-dask` inherits 

```yaml
- dask ==2.30.0    
- distributed ==2.30.1   
- dask-gateway ==0.9.0
```


danlester (changes requested):
@viniciusdc I think you also need to change the dependencies listed in DEFAULT_ENVIRONMENTS of qhub/initialize.py

viniciusdc:
> @viniciusdc I think you also need to change the dependencies listed in DEFAULT_ENVIRONMENTS of qhub/initialize.py

Sure, I will do. I missed that in my search. Other then that shouldn't `dashboard.yml` environmet contain those pacakges too (I could oly see the `ipykernel, ipywidgets >=7.6` there)?

danlester:
Well, I guess it depends on whether we think dask is something likely to be used in a dashboard... If they are relatively small installs, maybe put them in dashboard.yml too.

viniciusdc:
> Well, I guess it depends on whether we think dask is something likely to be used in a dashboard... If they are relatively small installs, maybe put them in dashboard.yml too.

Uhm I see, so I added it in this last commit, but my guess is that we really want to set this as a user choice right?

danlester (approved):

danlester:
Oh yes, that's fine - leave it as you have.

danlester:
Shall I leave it to you to resolve the documentation conflicts and merge?

PR #591, opened 2021-05-19
Secure dask dashboard behind forwardauth

danlester:
Uses forwardauth to protect dask dashboards behind JupyterHub auth using a new oauth client.

Note any JupyterHub user can access any dashboard.

### DIFF
--- a/qhub/template/{{ cookiecutter.repo_directory }}/templates/jupyterhub_config.py.j2
+++ b/qhub/template/{{ cookiecutter.repo_directory }}/templates/jupyterhub_config.py.j2
@@ -62,6 +62,8 @@ QHUB_USER_MAPPING = {{ cookiecutter.security.users }}
 QHUB_GROUP_MAPPING = {{ cookiecutter.security.groups }}
 QHUB_PROFILES = {{ cookiecutter.profiles.jupyterlab }}
 
+import escapism
+import string
 
 def qhub_generate_nss_files():
     passwd = []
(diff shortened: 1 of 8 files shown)


qhub/template/{{ cookiecutter.repo_directory }}/templates/jupyterhub_config.py.j2
```diff
@@ -142,11 +144,17 @@ def qhub_configure_profile(username, safe_username, profile):
...
     profile['kubespawner_override']['supplemental_gids'] = secondary_gids
     profile['kubespawner_override']['fs_gid'] = primary_gid
+
+    # Cost monitoring labels
+    k8s_safe_chars = set(string.ascii_lowercase + string.digits + '.-_')
+    k8s_safe_username = escapism.escape(username.lower().replace('@', '_'), safe=k8s_safe_chars, escape_char='-')
+    profile['kubespawner_override']['extra_labels'] = {
+        'owner': k8s_safe_username,
+        'team': user['primary_group']
+    }
```
aktech, line 154: Is this intentionally part of this pull request?

danlester: Well, sort of - they were really part of my Kubecost experiment, but think they could be useful for that and/or any other monitoring systems that people might want to try out in the meantime.

I think JupyterHub already adds another label with its own name for the owner anyway.

So, yes, I did include it on purpose but they aren't actually related to the dashboard auth.

aktech: OK, makes sense.

aktech (approved):

PR #604, opened 2021-05-25
Add support for GPUs on AWS

aktech:
- [x] Adds ability to spin up GPU nodes on AWS deployment
- [x] Adds daemonset to install nvidia device plugin 
- [x] Added docs and example
- [x] Tested with a deployment


danlester:
@aktech I have just updated the docs slightly (BTW please see how I put QHUB_VERSION in the docker tags for documentation purposes!)

Otherwise all looks good, if you are happy with my changes.

aktech:
Thanks @danlester that looks good to me.

PR #621, opened 2021-06-01
Terraform refresh before destroy to avoid failing destroy

aktech:
The terraform destroy fails on AWS if there is a significant delay between apply and destroy. This is why it was not caught in the integration tests.

The primary reason for this is in terraform >= 0.14, terraform doesn't refreshes the terraform state before generating a plan to destroy resources. This PR runs the refresh before destroy in order to cleanly destroy everything.


danlester (approved):
Fantastic - well done for working this out!

It make sense, and could explain some of the problems with GCP too.

PR #629, opened 2021-06-08
Instruct user to activate docker group addition.

toonarmycaptain:
Assimilate the command docker will prompt user to make into the useradd to docker step.

Alternative is for user to logout/in/reboot.


toonarmycaptain:
Everywhere I've followed these steps, docker has prompted me to do this when proceeding with 1inikube start`, so it might as well be in the docs. 
The [Docker post-installation](https://docs.docker.com/engine/install/linux-postinstall/) docs suggest in step three you can reboot/log out and log back in as an alternative, but it might be easier to just add the command to the docs than include the explanation.

danlester (approved):
Thank you!

PR #631, opened 2021-06-08
Update 1_project_setup_tutorial.md to remove reference to jupyter.<domain>

tylerpotts:
The DNS name was previously required to be prepended with `jupyter.<domain>`. This is no longer a requirement, and running `qhub init --domain <domain_name>` will deploy qhub to <domain_name>


danlester (approved):
Very helpful!

PR #632, opened 2021-06-09
Allow explicit ci-provider=none

danlester:
Can explicitly provide --ci-provider=none to qhub init (don't have to completely omit the parameter).

This makes integration tests easier.


viniciusdc (approved):
LGTM

PR #637, opened 2021-06-10
Cypress tests - simplify for Minikube

danlester:
Docs for running Cypress locally.
Run tests.yaml workflow on every branch.


costrouc (approved):
Looks like this will fix CI. Going to approve since it only touches tests.

PR #645, opened 2021-06-10
Change subned cidr mask

tylerpotts:
Changes the network mask for the QHub subnets on AWS from `x.x.x.x/24` to `x.x.x.x/20` to increase the number of available IP addresses

### DIFF
--- a/qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/modules/aws/network/main.tf
+++ b/qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/modules/aws/network/main.tf
@@ -12,7 +12,7 @@ resource "aws_subnet" "main" {
   count = length(var.aws_availability_zones)
 
   availability_zone       = var.aws_availability_zones[count.index]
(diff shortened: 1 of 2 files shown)


costrouc:
Waiting for Dan's cypress tests to be merged and then going to make sure that the CI passes. Thanks @tylerpotts

brl0:
I may be mistaken, but I think this change should have no effect on what is being tested since it will only affect AWS deployments.  

This change worked in our test deployment, and we are preparing to try again at making our production deployment with this change.

I do wonder if it would be better to create a variable like `var.vpc_cidrsubnet_newbits` to make it easier to override this value if appropriate in the future.

That said given our current situation I would be happy to see this change merged.

cc @iameskild

costrouc:
> I may be mistaken, but I think this change should have no effect on what is being tested since it will only affect AWS deployments.

Agreed this change won't affect the CI. Just like to see it green :smile:

> I do wonder if it would be better to create a variable like var.vpc_cidrsubnet_newbits to make it easier to override this value if appropriate in the future.

Great idea! Added it to this PR

costrouc (approved):
Look great thanks @tylerpotts and @brl0!

PR #646, opened 2021-06-10
ADD prefect optional field

### LINKED ISSUE #114: cookiecuter does not check if prefect is true/false only is defined
https://github.com/Quansight/qhub/blob/master/qhub/template/%7B%7B%20cookiecutter.repo_directory%20%7D%7D/infrastructure/variables.tf#L89

viniciusdc:
This should fix schema validation error for `prefect: true` new field in the `qhub-config.yml` file.

This seems to be the current error message when attempt to deploy QHub with prefect in it:

```bash
Problem encountered: 1 validation error for Main
prefect
  extra fields not permitted (type=value_error.extra)
```

See [discussion](https://github.com/Quansight/qhub/pull/641#issuecomment-859101173) 

Closes #114

### DIFF
--- a/qhub/schema.py
+++ b/qhub/schema.py
@@ -280,6 +280,7 @@ class Main(Base):
         TerraformModules
     ]  # No longer used, so ignored, but could still be in qhub-config.yaml
     certificate: Certificate
+    prefect: typing.Optional[bool]
     cdsdashboards: CDSDashboards
     security: Security
     default_images: DefaultImages


costrouc (approved):
LGTM. Side note we need to establish a convention for features. I would prefer

```
prefect:
   enabled: true
   ...
```

This is how cdsdashboards is configured but not important at the moment.

costrouc:
Related to mention above https://github.com/Quansight/qhub/issues/633

PR #650, opened 2021-06-15
Make load balancer ip and annotations configurable

costrouc:
Closes: https://github.com/Quansight/qhub/issues/638

Tried to do the minimal ammount to support load balancer
configuration. One downside I see with this PR is we are introducing
another 2 configuration variables to the qhub_config.yaml. On the
other hand this has been extremely easy to implment thanks to having
the qhub modules in the same directory.

cc: @leej3

### DIFF
--- a/qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/kubernetes.tf
+++ b/qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/kubernetes.tf
@@ -164,6 +164,18 @@ module "kubernetes-ingress" {
 
   node-group = local.node_groups.general
 
+{% if cookiecutter.load_balancer_ip is defined %}
+  load-balancer-ip = "{{ cookiecutter.load_balancer_ip }}"
+{% endif %}
+
+{% if cookiecutter.load_balancer_annotations is defined %}
+  load-balancer-annotations = {
+{% for key, value in cookiecutter.load_balancer_annotations.items() %}
+    "{{ "%s" | format(key) }}" = "{{ value }}"
+{% endfor %}
+  }
+{% endif %}
+
 {% if cookiecutter.certificate.type == "lets-encrypt" %}
   enable-certificates = true
   acme-email          = "{{ cookiecutter.certificate.acme_email }}"
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8, 3.9]
     steps:
       - name: 'Checkout Infrastructure'
         uses: actions/checkout@main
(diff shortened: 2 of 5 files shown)


qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/kubernetes.tf
```diff
@@ -158,6 +158,18 @@ module "kubernetes-ingress" {
...
   node-group = local.node_groups.general
 
+{% if cookiecutter.load_balancer_ip is defined %}
+  load-balancer-ip = "{{ cookiecutter.load_balancer_ip }}"
+{% endif %}
+
+{% if cookiecutter.load_balancer_annotations is defined %}
+  load-balancer-annotations = {
+{% for key, value in cookiecutter.load_balancer_annotations.items() %}
+    {{ "%-10s" | format(key) }} = "{{ value }}"
```
leej3, line 168: Should be:
`"{{ "%-10s" | format(key) }}" = "{{ value }}"`

costrouc: Oh that's right! Yeah I didn't consider the keys that kubernetes uses in annotations. Thanks for the catch!

costrouc: Fixed.

docs/source/02_get_started/04_configuration.md
```diff
@@ -499,6 +500,24 @@ storage:
...
+
+The kubernetes load balancer is used for exposing Qhub. Often times
+the load balancer may need to be configured with a static IP or
+additional annotations to use an internal vpc. QHub optionally
+supports these features.
+
+```yaml
+load_balancer_ip: "10.11.12.13"
+load_balancer_annotations:
+  "networking.gke.io/load-balancer-type": "Internal"
```
brl0, line 513: Not that it really matters much, but might be a bit more tidy to organize these under a common `load_balancer` key.

leej3: Agreed. There might be an increased number of this sort of tweak now that the terraform modules are integrated. A quick brain storm on how to organize these as they are added  would be valuable in the long run.
@costrouc I'm happy to deploy off any branch/fork so there's no rush on our deployment's account.

costrouc: Yeah I agree too. For me it was controversial to add this in that I'm hesitant to add more keys to the qhub-config.yaml. I would prefer to have it all under a single key. Will make that change

leej3:
I have chatted with @viniciusdc about this in the process of implementing #686. In determining the modifications to the schema we thought of something like the following where ip/annotations is a "dns_config" class in the schema. Does this seem reasonable

```
optional_features:
  load_balancer:
    enable: true
    dns_config:
      ip : "10.247.11.22"
      annotations: "networking.gke.io/load-balancer-type": "Internal"
  clearml:
    enable: true
    ...

```

iameskild:
we've marked this as Stale. if y'all want to continue working on this respond before 11/22.

viniciusdc:
This will be included in a major PR soon, will update later today

PR #672, opened 2021-06-21
Fix helm upgrade: remove helm_repository data source

aktech:
`helm_repository` data source is deprecated now.


danlester (approved):
Have approved, let's try it in AWS int test

PR #681, opened 2021-06-23
Add ClearML DNS only when its enabled

### LINKED ISSUE #680: ClearML optional rendering
@aktech,

We are temporarily using the main branch of qhub that contains some fixes we need.
Since the recent merge of #640 adds ClearML, I noticed that section of the terraform module is being rendered in our repo.  I am guessing (but may be wrong) that the only reason it did not actually deploy is we did not apply the app label to any of the nodes.
I think this section should probably only be rendered if ClearML is actually enabled in the qhub config file.

Additionally, it looks like if the dns was automatically provisioned, it would probably add the ClearML dns records even if it were not enabled, it looks like it just needs a condition added to check that.

cc @iameskild

aktech:
Fixes #680 

cc @brl0

### DIFF
--- a/qhub/deploy.py
+++ b/qhub/deploy.py
@@ -97,10 +97,12 @@ def guided_install(
         zone_name = ".".join(zone_name)
         if config["provider"] in {"do", "gcp", "azure"}:
             update_record(zone_name, record_name, "A", ip_or_hostname)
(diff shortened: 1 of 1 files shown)


danlester (commented):
Would it be safer to use .get("enabled", **False**)?

danlester (approved):

PR #684, opened 2021-06-23
Additional config validation checks

### LINKED ISSUE #209: Additional config validation checks
Added the following validation checks to QHub:
- [] multiple users with the same `uid`
- [] multiple JupyterHub profiles set as default
- [] multiple environments with the same name

viniciusdc:
fix #209 

This PR adds new validators under Pydantic `schema.py`: 
- `check_default` to guarantee that only one instance under Jupyterlab has defined as `default: true`;
- `validate_userids` to guarantee that there is no repeated user id under the users profiles.

The `yaml.safe_load` function inside `qhub render` and [`qhub deploy`](https://github.com/Quansight/qhub/blob/426d327246a5e40a21d8244e2728c917b20386d8/qhub/cli/deploy.py#L55) currently checks the last problem commented in that issue 
> multiple environments with the same name

### DIFF
--- a/qhub/schema.py
+++ b/qhub/schema.py
@@ -1,9 +1,10 @@
 import enum
+from os import pread
 import typing
 
 import pydantic
-
-from qhub.utils import namestr_regex
+from pydantic import validator, ValidationError
+from qhub.utils import namestr_regex, check_for_duplicates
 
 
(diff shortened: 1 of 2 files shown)


danlester (approved):
Works well, thanks!

PR #685, opened 2021-06-23
More AWS Tags

danlester:
Extra tags added to AWS resources so they can be found and destroyed more easily.


aktech (approved):
Looks good apart from the terraform fmt issues in the CI.

PR #687, opened 2021-06-23
Add optional features

leej3:


leej3:
Leads to undesirable levels of indentation

PR #694, opened 2021-06-25
AWS qhub force-destroy command

### LINKED ISSUE #678: AWS Force Destroy
Is there a way to destroy AWS resources automatically when terraform up/down has already failed?

danlester:
A `qhub force-destroy -c qhub-config.yaml` command (only works on AWS for now) that attempts to completely tear down a (potentially failed) QHub environment including remote terraform-state. Uses only AWS APIs directly (no terraform).

It could do with some more 'waiting' for earlier resources to definitely be removed, so at the moment is likely to have to be run multiple times before it picks up everything that needs to be deleted.

Closes #678


aktech:
Thanks for working on this. This has been a huge pain. I haven't looked at it closely, but I have few general comments:

- I think its definitely a good idea to do this if terraform is letting us down.
- Do we know the reason why terraform fails to do this, some docs around that would be very useful, so that if terraform fixes it, we can always use them.
- How do you feel about using the `--force` flag in the `qhub destroy`? My assumption is the motivation for a new command is it is using a differently technology to destroy resources instead of terraform.

danlester:
Amit, thanks for your thoughts!

> Do we know the reason why terraform fails to do this, some docs around that would be very useful, so that if terraform fixes it, we can always use them.

In the 'production release' case, my guess is that it's either just timeouts, or otherwise resources that have been changed when they shouldn't have been.

I think the biggest pain from my point of view is just when 'things go wrong' in dev environments. Not necessarily terraform at fault, but it doesn't seem to be able to recover sometimes.

In short, other than looking at timeouts more closely (which seem to be in the provider, not globally in terraform), I'm not 100% clear on whether terraform could do more to help us. Ultimately we should also be aware we are using an alpha kubernetes provider...

> How do you feel about using the --force flag in the qhub destroy? My assumption is the motivation for a new command is it is using a differently technology to destroy resources instead of terraform.

My feeling, having used force-destroy a few times is that it is very useful for dev use, but not ready for official use yet.

So maybe really it needs to be pulled out into a completely separate script, for devs to use - or users in a way that frames it as trying to be a helpful extra rather than a flimsy part of qhub itself.

There is still a lot that goes wrong in this script (which could be made more robust) and in practice I have to run it multiple times, sometimes commenting bits out... but it provides a way to get all relevant resources listed and a formula for running through them to delete in the right order.

I think I will pull it out into a 'scripts' folder unless you disagree! It doesn't use much from qhub other than reading in qhub-config.yaml, but doesn't validate it in case the schema is mismatched anyway.

danlester:
OK have moved it into a scripts folder and feel more comfortable with this being merged to see if it helps us...

For the integration tests especially, it would still be useful to address the timeouts issue.

danlester:
Flake8 is now fixed but I didn't rerun the tests.

aktech (approved):
Thanks @danlester for the explanation, this looks good to get in.

aktech:
would be good to run the CI, before merge

PR #716, opened 2021-07-01
Include namespace in conda-store PVC

### LINKED ISSUE #708: [bug] dask-scheduler pod unschedulable due to persistent volume claim "conda-store-dev-share" not found
https://github.com/Quansight/qhub/blob/ee21839c811d8b548ec5d013e9cba40dc6786ff7/qhub/render/__init__.py#L25

I'm using qhub version 0.3.11, deploying to GCP.

In my qhub config file I have changed the namespace from "dev" to "qhub" before deployment.

This deploys and works fine with the vanilla examples in the docs. However, when I add nodegroup selector configuration for the dask workers as described [here](https://docs.qhub.dev/en/stable/source/05_reference/02_system_maintenance.html#setting-specific-dask-workers-to-run-on-a-nodegroup) then redeploy, the dask-scheduler pod becomes unschedulable, due to persistent volume claim "conda-store-dev-share" not found.

I have looked at the persistent volume claims and there is "conda-store-qhub-share" present, so I am guessing this is because the config rendering has "conda-store-dev-share" hard-coded at the line linked above. Perhaps this needs to incorporate the namespace?

alimanfoo:
This PR builds the name for the conda-store PVC using the `namespace` variable from the config, in the case that the user has changed it from the default value of "dev".

Should resolve #708.

Apologies I don't have any idea how to test this, but submitting in case it's useful.

### DIFF
--- a/qhub/render/__init__.py
+++ b/qhub/render/__init__.py
@@ -20,9 +20,10 @@ def patch_dask_gateway_extra_config(config):
     directory.
 
     """
+    namespace = config["namespace"]
     conda_store_volume = {
         "name": "conda-store",
-        "persistentVolumeClaim": {"claimName": "conda-store-dev-share"},
+        "persistentVolumeClaim": {"claimName": f"conda-store-{namespace}-share"},
     }
     extra_pod_config = {"volumes": [conda_store_volume]}
 


aktech (approved):
Looks good to me. Thanks @alimanfoo

PR #721, opened 2021-07-06
Secure ClearML behind forwardauth

### LINKED ISSUE #666: Secure ClearML behind Auth
We recently integrated ClearML into QHub in this PR: #640 At the moment the ClearML is publicly accessible, it will be great to secure it behind forward auth, the way we secure dask-gateway.

aktech:
Fixes #666 

- [x] Fixes ClearML TLS issue
- [x] Add forwardauth to ClearML

cc @danlester

### DIFF
--- a/qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/modules/kubernetes/services/clearml/variables.tf
+++ b/qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/modules/kubernetes/services/clearml/variables.tf
@@ -24,5 +24,4 @@ variable "external-url" {
 
 variable "tls" {
   description = "TLS configuration"
-  type        = object({})
 }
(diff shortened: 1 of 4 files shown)


qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/modules/kubernetes/services/clearml/variables.tf
```diff
@@ -24,5 +24,4 @@ variable "external-url" {
 
 variable "tls" {
   description = "TLS configuration"
-  type        = object({})
```
danlester, line 27: Should there be a different type specified then?

danlester (approved):
Looks fine to me - just wondered if there should be a different type for the TLS variable though.

qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/modules/kubernetes/services/clearml/variables.tf
aktech: The problem here is you need to have explicit keys for the object, you can't have arbitrary keys (like say `**kwargs` in python) and even the optional keys in the typing is an experimental feature in terraform.

aktech: and we don't know the exact keys beforehand:

https://github.com/Quansight/qhub/blob/e3eb1d309714f4d53a620b9e4ec453462c7822b9/qhub/template/%7B%7B%20cookiecutter.repo_directory%20%7D%7D/infrastructure/modules/kubernetes/services/meta/qhub/locals.tf#L2-L6

PR #723, opened 2021-07-07
Reduce minimum nodes in user and dask node pools to 0 for Azure / GCP

tarundmsharma:
Revert changes: Update min_nodes for some cloud providers #709


viniciusdc:
@costrouc Could you have a look at this too? As you've commented about some special circumstances for the cloud providers... Those changes should revert the min nodes configuration to the original structure (before 0.3.11 release)

viniciusdc (approved):

tonyfast:
@costrouc i added you as a reviewer to this. @viniciusdc has already approved, what do you think we should do?

iameskild:
Reading through the Azure AKS node pool docs, it appears that the minimum number of nodes in a user pool can be set to 0.

[See the third bullet point under limitations](https://docs.microsoft.com/en-us/azure/aks/use-multiple-node-pools#limitations):
> System pools must contain at least one node, and user node pools may contain zero or more nodes.

iameskild:
Reading the docs for the GCP node pool and auto-scaler, it also appears that we can scale user / dask node pools to 0.

[See GKE auto-scaler minimum and maximum node pools](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-autoscaler#minimum_and_maximum_node_pool_size):
> Note: If you specify a minimum of zero nodes, an idle node pool can scale down completely. However, at least one node must always be available in the cluster to run system Pods.

costrouc (approved):
Looking at the docs it seems that azure and gcp both support scaling down to zero. Thanks!

Issue #619, opened 2021-05-31
Pin QHub dependencies 


aktech:
The QHub CLI was broken most likely due to the unpinned version of library `ruamel.yaml`. There are a bunch of libraries in the `setup.py`, which should be pinned to have more stability.

https://github.com/Quansight/qhub/blob/51ed41bd1cc1bb13cb0e3fdb729ef6837c1ffcd3/setup.py#L34-L58

costrouc:
Agreed. Especially for a devops focused tool I think we should favor pinning for stability.

viniciusdc:
Should we pin that directly or give it a lower bound? @costrouc

tarundmsharma:
@costrouc, can you please help me with the pin versions.

tarundmsharma:
https://github.com/Quansight/qhub/pull/724

PR #724, opened 2021-07-07
Pin QHub dependencies #619

tarundmsharma:
Pin QHub dependencies for ruamel.yaml, cloudflare, auth0-python, pydantic, pynacl, bcrypt, twine, pytest, diagrams and jhub-client


viniciusdc:
@tarundmsharma It seems that the local testing is failing, would you mind checking if this environment is solvable and trying to do a local deployment of Qhub using those pinned packages? Just to be sure that there aren't any inconsistencies with the packages that `qhub` now utilizes

iameskild:
This might not be a sustainable long-term solution. If we need to pin some of these packages, perhaps we can try upper and lower bounds instead. Close this PR for now.

PR #730, opened 2021-07-10
Add lab extensions jupyter-resource-usage and nbgitpuller

alimanfoo:
In our QHub deployment we've found it useful to add a couple of jupyterlab extensions. The first is the `jupyter-resource-usage` extension which shows memory used/available on the VM running lab. The second is the `nbgitpuller` package which allows distributing URLs that load a given notebook into the JupyterHub service. Thought I'd suggest these as generally good additions to include in the lab image.

### DIFF
--- a/qhub/template/{{ cookiecutter.repo_directory }}/image/jupyterlab/environment.yaml
+++ b/qhub/template/{{ cookiecutter.repo_directory }}/image/jupyterlab/environment.yaml
@@ -24,6 +24,8 @@ dependencies:
  - ipywidgets >=7.6
  - ipyleaflet >=0.13.5
  - pyviz_comms >=2.0.1
+ - jupyter-resource-usage
+ - nbgitpuller
 
  # cds dashboards
  - cdsdashboards-singleuser >=0.5.7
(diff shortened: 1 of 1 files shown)


costrouc (changes requested):
Thanks @alimanfoo for these additions! I think there are worthwhile to add to the default image. We certainly see a lot of issues with kernels dying due to out of memory errors and having an easy way for them to see their usage would be helpful. Also nbgitpuller would be helpful as well for like you said automatically downloading repos.

Could you fix the pin for `  - jupyter-resource-usage >=0.6.0`. The changelog indicates that jupyterlab 3 support was added in that version https://github.com/jupyter-server/jupyter-resource-usage/blob/master/CHANGELOG.md#060.

Otherwise looks great to me.

alimanfoo:
> Could you fix the pin for ` - jupyter-resource-usage >=0.6.0`. The changelog indicates that jupyterlab 3 support was added in that version https://github.com/jupyter-server/jupyter-resource-usage/blob/master/CHANGELOG.md#060.

Thanks @costrouc, no problem, done.

costrouc (approved):

PR #731, opened 2021-07-12
Pin jupyterhub package in lab image to match the hub image

alimanfoo:
The [JupyterHub docs](https://jupyterhub.readthedocs.io/en/stable/admin/upgrading.html?highlight=version#upgrade-jupyterhub-packages) recommend that the version of the jupyterhub package must match between the hub environment and the notebook user environments. 

Currently qhub pins the jupyterhub package to version 1.4.1 in the jupyterhub image environment, but does not pin the jupyterhub package in the jupyterlab environment. This PR suggests pinning also in the jupyterlab environment, to avoid version mismatch between the images.

### DIFF
--- a/qhub/template/{{ cookiecutter.repo_directory }}/image/jupyterlab/environment.yaml
+++ b/qhub/template/{{ cookiecutter.repo_directory }}/image/jupyterlab/environment.yaml
@@ -12,7 +12,7 @@ dependencies:
  - jupyterlab >=3
  - jupyter_client
  - jupyter_console
- - jupyterhub
+ - jupyterhub==1.4.1
  - nbconvert
  - nbval
 


danlester:
Thanks for this suggestion, and sorry that it appeared to get lost... but we have now pinned JupyterHub at 1.5.0 for the jupyterlab Docker image in the `main` branch as [seen here](https://github.com/Quansight/qhub/blob/ad8295927324a46ebe6773c43f1055ffed2e03b9/qhub/template/%7B%7B%20cookiecutter.repo_directory%20%7D%7D/image/jupyterlab/environment.yaml#L15).

This coincides with upgrading the jupyterhub package (and helm chart) on the JupyterHub side too.

Of course, you can build and use your own versions of the Docker images as needed instead.

So thanks again, and other than now having a higher version number, this suggestion has inspired our separate code update and I'll close this PR here without merging.

PR #741, opened 2021-07-23
External Container Registry

danlester:
Allow user to specify a private external AWS ECR as a replacement for Docker Hub for selected images, on any cloud.
Add to qhub-config.yaml:
```
external_container_reg:
  enabled: true
  access_key_id: <AWS access key id>
  secret_access_key: <AWS secret key>
  extcr_account: 892486800165
  extcr_region: eu-west-2
```
If this section is completely omitted, qhub should never complain (and assume enabled=false).


danlester:
Merging now to ensure my codebase doesn't diverge too much...

PR #751, opened 2021-08-09
Qhub extensions

danlester:
A mechanism for adding arbitrary service via qhub-config.yaml


danlester:
This is now part of keycloak branch #780

PR #797, opened 2021-08-31
Add Traefik Dashboard

Adam-D-Lewis:
- [x] Configures Traefik to produce data for Prometheus to Scrape
- [x] Configures Prometheus to scrape data from Traefik
- [x] Adds a grafana dashboard displaying relevant traefik data


qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/modules/kubernetes/services/monitoring/main.tf
```diff
@@ -5,6 +5,37 @@ resource "helm_release" "kube-prometheus-stack-helm-deployment" {
...
+      kubernetes_sd_configs:    
+        - role: service    
+          
+      relabel_configs:    
+      - source_labels: [__meta_kubernetes_service_label_app_kubernetes_io_component]    
+        action: keep    
+        regex: traefik-internal-service    
+      - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_scrape]    
+        action: keep    
+        regex: true
```
Adam-D-Lewis, line 26: As set up now, this will scrape only services with 2 labels `app.kubernetes.io/component: traefik-internal-service` and `prometheus.io/scrape: true`.  

It is possible to instead scrape all services with the label `prometheus.io/scrape: true`, but they'll all be included under the same job.  I'm not sure if that's the "right way" to do things in Prometheus given their definition of job which I've pasted below (from [here](https://prometheus.io/docs/concepts/jobs_instances/)):
>In Prometheus terms, an endpoint you can scrape is called an instance, usually corresponding to a single process. A collection of instances with the same purpose, a process replicated for scalability or reliability for example, is called a job.

Adam-D-Lewis: @costrouc

Adam-D-Lewis:
I also had to edit the traefik dashboard a bit since our traefik container (not pod) is called `qhub` instead of `traefik`.  Additionally, some of the prometheus metrics queried did not include the namespace label, and the dashboard assumed they did.

qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/modules/kubernetes/services/monitoring/main.tf
costrouc: > As set up now, this will scrape only k8s services with both labels app.kubernetes.io/component: traefik-internal-service and prometheus.io/scrape: true (see relabel configs).

I'm not a fan of this since for each new service we'll have to add a new job for scraping. 

> It is possible to instead scrape all services with the label prometheus.io/scrape: true, but they'll all be included under the same prometheus job. 

This is exactly what I'd like to have. I'd like to not have to require development to qhub to require additional modifications to the monitoring configuration.

costrouc:
Asside from requiring two labels this PR looks good to me. I'd like to have anything with `prometheus.io/scrape: true` be scraped.

tylerpotts:
Combined the 4 return code graphs into one

image

Adam-D-Lewis:
I'm testing out adding a sleep function before cypress runs, but the tests are still flaky (passed once, failed once) when sleeping 60 seconds before cypress runs.

PR #800, opened 2021-09-02
Update `remove_existing_renders` to only delete qhub related files/directories

### LINKED ISSUE #767: [bug] Running QHub within home directory will delete files in home directory
## Describe the bug

If a user runs the typical install process from their home directory.

```shell
conda create -n qhub -c conda-forge qhub==0.3.12
conda activate qhub

qhub init ...
qhub deploy ...
```

This will delete file within the home folder. This is due to the logic in https://github.com/Quansight/qhub/blob/744060aeaf19c194d39fd418c9834790c21f9f00/qhub/render/__init__.py#L139.

## Acceptance Criterial

 - remove all `os.remove` and `shutil.rmtree` logic in QHub this is too much of a risk for qhub.

iameskild:
Closes #767 

The approach taken will only remove files/directories found in the `deletable_dirs` list. In the linked issue, the acceptance criteria specifies removing all `os.remove` and `shutil.rmtree` logic so my current solution employs `Path` object methods (which essentially do the same thing). 

I have also added a `verbosity` flag (currently set to `0` - none) if and when we would like to output the list of files/directories being deleted. 

All feedback is welcome :)

### DIFF
--- a/qhub/render/__init__.py
+++ b/qhub/render/__init__.py
@@ -3,7 +3,7 @@ import collections
 import functools
 import json
 import os
-from shutil import copyfile
+from shutil import copyfile, rmtree
 from gitignore_parser import parse_gitignore
 
 from ruamel import yaml
@@ -133,7 +133,7 @@ def render_template(output_directory, config_filename, force=False):
 
     remove_existing_renders(
         dest_repo_dir=output_directory / repo_directory,
-        verbosity=0,
+        verbosity=2,
     )
 
     generate_files(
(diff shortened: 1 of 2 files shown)


qhub/render/__init__.py
```diff
@@ -144,60 +144,35 @@ def render_template(output_directory, config_filename, force=False):
...
-    else:
+    home_dir = pathlib.Path.home()
+    if pathlib.Path.cwd() == home_dir:
+        raise ValueError(f"Deploying QHub from the home directory, {home_dir}, is not permitted.")
+
+    deletable_dirs = ["terraform-state", ".github", "infrastructure", "image"]
+
+    for deletable_dir in deletable_dirs:
+        deletable_dir = dest_repo_dir / deletable_dir
+        if deletable_dir.exists():
```
costrouc, line 161: Ah. Sorry I shouldn't have mentioned that `os.remove` and `shutil.rmtree` shouldn't be used. I meant that more generally that long term i'd like to minimize the number of times that we make this call! Since it's a scary operation.

I'm fine with using `shutil.rmtree` here and I think that will dramatically reduce the code here.T his should remove all the walking in the code.

Also `deletable_dirs` should include `.gitlab-ci.yml`.

costrouc (changes requested):

qhub/render/__init__.py
iameskild: Got it! Thanks for the feedback. And I agree, using `shutil.rmtree` would be preferred.

costrouc (approved):
Thanks @iameskild this looks great to me and worked when I tested it locally.

PR #823, opened 2021-09-16
Aws ci

Adam-D-Lewis:
Use AWS Spot Instances for CICD.  This could be a pain if our instances often get pre-empted, but I think we should try it and see how often the instances are pre-empted.  When instances do get preempted, I imagine the error message will make it clear that is the case so you won't be left wondering if your code has an error or if the CI is failing.

FYI, the costs associated with this option are:
- AMI storage (should be a few dollars per month)
- EBS (32GB while instances are running)
- EC2 costs for **spot** instances (t3a.xlarge instance costs ~$0.05/hr at the moment billed by the second)
- Data ingress/egress costs (downloading minikube, github repo, etc.)


.cirun.yml
```diff
@@ -1,15 +1,16 @@
-# Self-Hosted Github Action Runners on DigitalOcean via Cirun.io
+# Self-Hosted Github Action Runners on AWS via Cirun.io
 # Reference: https://docs.cirun.io/reference/yaml.html
 runners:
   - name: run-k8s-tests
     # Cloud Provider: DigitalOcean
```
aktech, line 5: Comment: AWS

.cirun.yml
```diff
@@ -1,15 +1,16 @@
...
-    cloud: digitalocean
+    cloud: aws
     # Cheapest VM on DigitalOcean
```
aktech, line 7: AWS and not the cheapest

.cirun.yml
```diff
@@ -1,15 +1,16 @@
...
-    instance_type: g-4vcpu-16gb
-    # Ubuntu-20.4  image"
-    machine_image: docker-20-04
-    region: nyc1
+    instance_type: t3a.xlarge
+    # Ubuntu-20.4 image"
```
aktech, line 9: It's a custom image

.github/workflows/kubernetes_test.yaml
```diff
@@ -18,19 +18,41 @@ jobs:
...
+
+          sudo systemctl enable docker
+          sudo systemctl restart docker
+          echo 'systemctl enable docker done' 
+          df -h 
+          df -h /var/lib/docker
+      # groupadd: group 'docker' already exists
+      # Password: su: Authentication failure
+      # systemctl enable docker done
+
```
aktech, line 43: This step is not required, I already installed docker in the custom AMI.

Adam-D-Lewis:
All good points, @aktech.  Thanks for your review.  I've made the suggested changes.  Could you re-review when you get a chance please?

aktech (approved):
Thanks @Adam-D-Lewis Looks great to me, we can merge this when CI is green.

PR #832, opened 2021-09-30
Increase default subnets in AWS VPC

### LINKED ISSUE #828: [bug] Unable to spin up 200+ dask workers on AWS QHub
## Describe the bug
A clear and concise description of what the problem is.

When requesting a large number of dask workers, some fraction of worker pods never spin up. During some reason testing, I requested 250 workers and was only ever to get 190 (or so) workers to spin up; 190 (or thereabouts) seems to be the limit of dask workers nodes we can spin up at one time. For the node-group in question, we have set the max number of workers to the default 450 in the `qhub-config.yaml`.  

Checking the status of these nodes on AWS reveals an interesting error message:
```
runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:docker: network plugin is not ready: cni config uninitialized
```

## How can we help?
Help us help you.
* What are you trying to achieve?
* How can we reproduce the problem?
* What is the expected behaviour?
* And what is currently happening?
* Any error messages?
> If helpful, please add any screenshots and relevant links to the description.

Here are a few screenshots of these error messages:

Screen Shot 2021-09-20 at 16 30 50
Screen Shot 2021-09-20 at 16 31 55

## Your environment
Describe the environment in which you are experiencing the bug.
> Include your conda version (use `conda --version`), k8s and any other relevant details.
```
qhub - 0.3.12 (installed from commit 0dff706)
k8s - 1.19
```

iameskild:
Closes #828 

This will enable AWS subnets to handle `2**(32-8-4) = 1048576` IP addresses.

### DIFF
--- a/qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/modules/aws/network/variables.tf
+++ b/qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/modules/aws/network/variables.tf
@@ -35,6 +35,7 @@ variable "aws_availability_zones" {
 variable "vpc_cidr_block" {
   description = "VPC cidr for subnets to be inside of"
   type        = string
+  default     = "10.0.0.0/8"
 }
 
 variable "vpc_cidr_newbits" {


iameskild:
@costrouc sorry I didn’t see your (slack) comment above before submitting my PR. I added a default `vpc_cidr_block` for AWS deployments, would you still prefer the updating the default `vpc_cidr_block` for all deployments?

costrouc:
@iameskild from the issue I think that we will have to change the numberof bits for the subnet for now. Like @Adam-D-Lewis found that we can't have a vpc cidr larger than 16 😦

PR #833, opened 2021-10-04
Add `pre-commit` to repo

iameskild:
I have forgotten to run `black` and `flake8` before committing my changes only for the CI to fail with formatting issues. To get around this, I propose using `pre-commit` hooks. 

I have also updated the "Developer docs" sections to reflect how the package might be installed and how to get started with `pre-commit`. Lastly, I want to confirm that we will no longer need to run `pip install qhuib-ops` in order to start QHub development.


costrouc (approved):
Thanks for adding the pre-commit hook @iameskild !

PR #846, opened 2021-10-09
Added kubernetes version using azure python sdk for azure provider

nayaksuraj:


nayaksuraj:
@costrouc Please review and suggest if any optimisation or I missed anything.

PR #873, opened 2021-10-19
Documentation cleanup

HarshCasper:
## Description

This PR:

- Make `qhub` compatible with Vale and add more vocabulary text
- Fix all broken links on the documentation website
- Add a GHA to check for broken links for every PR
- Add Google Documentation style guide to be followed by QHub docs


HarshCasper:
Hi @costrouc 

I have migrated Vale-specific configurations to the `tests/` directory. I have also added some minimal docs to run Vale locally. It was also a good perspective from the docs side of how we can use Vale to find issues in our docs and quickly fix them up. There is still (a lot of) noise but it would be hopefully improved in the coming days.

PR #885, opened 2021-10-27
QHub development command to make development/testing easier

costrouc:
The goal of this PR is to make it such that there is a simple way to develop QHub. Currently the development docs talk about manual steps about spinning up minikube, setting up a load balancer, running the qhub init, deploy commands. This also does not address the fact that we are not building the local dockerfile images for testing.

This PR hopes to address this while also making the command line output much more human friendly.

See this asciinema for the experience https://asciinema.org/a/WQqBdcpXvkrboFRcdILYK2hbi

The `qhub develop` command does several things:
  - it run QHub with the current state of the qhub repository that you are in
  - It builds the docker images on the minikube cluster
  - It deploys QHub using the docker images in the repository also with any changes to the terraform modules

Other miscellaneous things this PR does:
  - all qhub exceptions are subclasses of QHubError
  - all exceptions are caught in the cli. If it is a subclass of QHubError then the error message is printed and it exits. If it is not a subclass of QHubError then it is treated as a bug
  - further testing on the providers now

If you don't delete the minikube cluster between changes complete deployment is usually around 40 seconds.

### DIFF
--- a/docs/source/dev_guide/testing.md
+++ b/docs/source/dev_guide/testing.md
@@ -300,3 +300,26 @@ QHub uses [Vale](https://github.com/errata-ai/vale), a syntax-aware linter to li
 - In case of false positives, make appropriate changes to the Vale configurations hosted in the `tests/vale/styles` directory.
 
 Vale runs on the GitHub Actions CI to automatically validate the documentation language. By default, Vale only checks the modified documentation to ensure that Vale doesn't generate a lot of noise over the Pull Requests.
+
+# Nix Development
+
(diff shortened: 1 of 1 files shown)


danlester:
@costrouc I started to take a look at this PR.

One issue is that `pip install -e .` doesn't seem to work for this branch. I get ` ModuleNotFoundError: No module named 'setuptools'`. I haven't investigated, but don't see any major change to setup.py from main branch.

Please could you do a rebase to the latest main and then I can take a closer look? Thank you.

Issue #892, opened 2021-11-02
[enhancement] Add better metadata to the Docker images


trallard:
## Description
As part of the Docker over-haul currently going on it would be good to add better metadata for the Docker images, to facilitate findability, reproducibility and trust
Right now the Docker images have a single label `maintainer` 

## Suggestion
Add more `OCI` compliant annotations to the Docker images see<https://github.com/opencontainers/image-spec/blob/main/annotations.md>

## Context
This adheres to industry-standard best practices and increases the end-user confidence

iameskild:
Hey @trallard, thanks for the suggestion! I'm only faintly aware of OCI, but I started with adding the following labels:
```
LABEL org.opencontainers.image.authors="Quansight" \
      org.opencontainers.image.source="https://github.com/Quansight/qhub"
```

Are there any others that you would recommend? Perhaps a description for each of image in question?

trallard:
I would suggest adding the following on top of source and maintainer:

- description 
- created (previous build-date from label schema) - this can be updated through GH actions
- license 

Also it is advisable that the `author` label has a contact option (email, github handle, etc)

iameskild:
@trallard @HarshCasper I just realized that we already set OCI-compliant labels :)

https://github.com/Quansight/qhub/runs/4111162873?check_suite_focus=true#step:5:32

Who should we add as `author`? @costrouc?

trallard:
Maybe the QHub development team? - do we have a qhub email or something?

trallard:
@iameskild  are you still working on items related to this or shall we close the issue?

iameskild:
Hey @trallard, I believe that we can. Here are the labels currently being applied:

```
      "org.opencontainers.image.title": "qhub",
      "org.opencontainers.image.description": "QHub deployment tool",
      "org.opencontainers.image.url": "https://github.com/Quansight/qhub",
      "org.opencontainers.image.source": "https://github.com/Quansight/qhub",
      "org.opencontainers.image.version": "docs-update-cleanup",
      "org.opencontainers.image.created": "2021-12-13T18:02:27.414Z",
      "org.opencontainers.image.revision": "7bcfa7e72875102025276298168c8a95e2518f5c",
      "org.opencontainers.image.licenses": "BSD-3-Clause"
```

I didn't find "email" as a predefined OCI key but we can add another field/key if you would like to include it. Otherwise this issue can be closed.

trallard:
that should be enough - will go ahead and close this issue

Issue #888, opened 2021-10-29
[enhancement] Use Docker Slim for smaller Docker images


HarshCasper:
## Background

During the setup and installation process, I suffered a lot with the large Docker container sizes. We can slim the Docker container size using `DockerSlim`, a command-line tool that allows us to reduce the size of the Docker images and improve the time required to download the image. This can allow us to reduce the CI time required to download the images and make it lightweight to use locally as well without adding any extra overhead, by flattening the output container layer structure.

## How does Docker Slim work?

Docker Slim uses both static and dynamic analysis to work. It intelligently determines the files, libraries and executables that are required for the container to work and remove the extra overhead in the process. 

During the container-run process, Docker Slim probes the functions running in the container and records these dependencies while composing the output container.

It does not modify the original input container; Rather it produces an output container with a `.slim` attached to the original input container image name. 

## Testing

| Name                        | Original | Slim    | Minified |
| --------------------------- | -------- | ------- | -------- |
| quansight/qhub-jupyterhub   | 702 MB   | 73.3 MB | 9.58x    |
| quansight/qhub-jupyterlab   | 2.53 GB  | 5.44 MB | 465.08x  |
| quansight/qhub-dask-worker  | 753 MB   | 3.68 MB | 204.86x  |
| quansight/qhub-dask-gateway | 179 MB   | 30 MB   | 5.94X    |
| quansight/qhub-conda-store  | 2.98 GB  | 27.3 MB | 108.91X  |

## Linting

Using the Lint option, we analyze the Dockerfiles and analyze the container instructions. It is a good way to analyze and refactor our Dockerfiles to make them more readable and remove anti-patterns.

#### `Dockerfile.jupyterhub`
<details>
  <summary>slim.report.json</summary>

```sh
{
  "version": "1.0",
  "engine": "linux|Transformer|1.37.0|70cc8acfcb733161ce7e685b81ad6c172643c222|2021-09-23_09:23:43AM",
  "containerized": false,
  "host_distro": {
    "name": "Red Hat Enterprise Linux",
    "version": "8.2",
    "display_name": "Red Hat Enterprise Linux 8.2 (Ootpa)"
  },
  "type": "lint",
  "state": "done",
  "target_type": "dockerfile",
  "target_reference": "Dockerfile.jupyterhub",
  "hits_count": 2,
  "nohits_count": 23,
  "errors_count": 0,
  "hits": {
    "ID.10001": {
      "source": {
        "id": "ID.10001",
[...]

Issue #872, opened 2021-10-19
[enhancement] Edit QHub's timezone


pierrotsmnrd:
## Description

I noticed that my JupyterLab instances were set up with the `America/Chicago` timezone, as defined [here](https://github.com/Quansight/qhub/blob/a5f3dde62abb8e0831d7a7d8fd3ec559cea128e3/qhub/template/%7B%7B%20cookiecutter.repo_directory%20%7D%7D/image/Dockerfile.jupyterlab#L21).

I see multiple use cases where this can lead to bad user experience. For instance, a user having made a notebook that displays a time and date _on his own machine_ will have to edit his code to adapt to the different timezone _on QHub_.
It's even more problematic if the notebook is actually meant to be a dashboard shared with people.

## Suggestion

By giving enough flexibility to set up a unique timezone, it removes the unexpected surprise for Users.

- First option : Have the Admins set up  _one unique_ timezone for _all_ the components of QHub. 
This is convenient for companies spread over only one timezone. It's also easier to implement (I guess)

- Second option : Have the users be able to set up their own timezone. 
For a full-remote company like us, spread over many timezones, it makes sense : If I run a codebase showing the current time in France, I want it to show the current time at your place too.

## Context

- With the Prefect integration, a focus on timezones coherence becomes necessary : we don't want a task to be run at the wrong time. 

- Having to modify code to adapt to a timezone is painful and should be unnecessary.

trallard:
I have noticed the same and I found it an odd decision to make `Chicago` the default tz identifier.  
I am +1 on adopting a universal tz such as `UTC`  and explicitly call it in the docs (many projects use UTC)

Plus giving the option to change the default timezone as needed

viniciusdc:
c.c. @costrouc what are your thoughts on this?

iameskild:
I think customizing the JuptyerLab image with a specific TZ is possible only if we allow users to customize their own images: https://github.com/nebari-dev/governance/issues/22

pavithraes:
Customizing the timezone can be handled by `jhub-apps`, moving this issue to the relevant repo.

We can track changing the default timezone in the docker-images repo :)

Issue #858, opened 2021-10-14
[enhancement]  Add the black labextension


rsignell-usgs:
## Description
Would be cool to have the Black labextension available to Qhub users:
https://coiled.io/blog/code-formatting-jupyter-notebooks-with-black/

iameskild:
Hi @rsignell-usgs, could you try adding it to one of your conda environments (in the `qhub-config.yaml`) and redeploying? Although I haven't tried this with a qhub deployment yet, testing it in local `jupyter lab` session seemed to work :)

trallard:
@iameskild or @viniciusdc  I think this should be easy to add to the new Docker images now that they are in `nebari-dev` can either of you take this on?

iameskild:
@trallard @rsignell-usgs I've raised a PR to add `jupyterlab_code_formatter` :)
https://github.com/nebari-dev/nebari-docker-images/pull/26

PR #904, opened 2021-11-05
Add support for CentOS

### LINKED ISSUE #884: [Enhancement] Add support for CentOS
## Description
It would be great to have support for CentOS with the current Ubuntu.

Ideally we should be able to chose between CentOS and Ubuntu on the JupyterLab interface.

## Suggestion
What are your suggestions on how to solve this problem? What solution(s) would make you happy?

- [ ] TODO

## Context
A Plenty of users/orgs have a hard dependency on CentOS. This is in general would be a great feature to have, i.e. being able to create and uses images for multiple operating system.

iameskild:
Closes #884 #872 #892 #858 #888 

## Changes:

As part of the work to add CentOS JupyterLab and Dask-Worker profiles (Docker images), the JupyterHub, JupyterLab and Dask-Worker images have been refactored. 
- There is now a base image for each respective linux distribution. These bases images are then used when building all three images list above; this pattern was inspired by the [Pangeo's Docker image build](https://github.com/pangeo-data/pangeo-docker-images). However the underlying motivation for the use of base-images is due to the number of duplicate layers those three images already had in common, namely installing miniconda (and mamba) and other basic linux packages.
  - To reduce the number of Dockerfiles that need to be maintained (at least slightly), the JupyterLab and Dask-Worker (and yes, technically the JupyterHub) Dockerfiles can be built from either base image (`qhub-ubuntu-base` or `qhub-centos-base`) by passing the a `--build-arg BASE_IMAGE=<>` during build time.
  - The `Build Image Workflow` has been modified to accommodate the new build process. The default `ubuntu` images are still built with the same tags (no changes to how you would pull these images down). And to pull down `centos` images, simply append `centos` to the tag (ie. `qhub-dask-worker:v0.3.13centos`)
- To decrease build time, base conda packages + environment are now installed with mamba; this decreases the build time by a factor of about x3.5.
- To decrease the JupyterLab image size, Linux packages that are used by [conda GUI packages have been removed](https://docs.anaconda.com/anaconda/install/linux/). This, and other modifications, decrease the JupyterLab image size by ~250MB
  - It is my understanding that these GUI-related packages are unused; we don't run conda-specific GUI packages, do we?
  - Reducing this image any more than this might be difficult given the number of packages (conda and otherwise) that it needs. Thank you to @HarshCasper for exploring and testing the use of Docker Slim, unfortunately he found that there are limitations that prevent us from going forward with it.
    - #888 

Other changes:
- Pin `TZ` (timezone) to `UTC` 
  - #872 
  - I found it difficult to accommodate "local" timezones
- Docker images already follow OCI metadata conventions 
  - #892 
  - See [here](https://github.com/Quansight/qhub/runs/4111162873?check_suite_focus=true#step:5:32)
- Install `jupyterlab_code_formatter` 
  - #858 

- [ ] Documentation Update
- [x] Build related changes

### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

Although if you'd like we can add tests similar to how [pange-docker-images](https://github.com/pangeo-data/pangeo-docker-images/blob/master/tests/test_all.py) does?

## Changes to Include (or Add in the Future)
- Recommendations from @HarshCasper and @trallard to add [hadolint](https://github.com/hadolint/hadolint) - Docker Image linting
- Add `qhub init --distro` flag that automatically populates the profiles with centos JupyterLab and Dask-Worker images (??)
- Add docs to explain how to add CentOS profiles


pierrotsmnrd:
Hi @iameskild  :) 

Regarding the change to add `jupyterlab_code_formatter`:
- I've tried on my  side, as you did, by adding it in the Jupyter_Hub_ environment.yaml, without success.
- I've tried putting it in the Jupyter_Lab_ environment.yaml, without success either
- I've tried putting it in the Jupyter_Lab_ environment.yaml and also adding it in my conda env, didn't work 

Have you managed to make it work ? 
Regarding the mentionned issue #858 I think the goal is to have the button in the toolbar like shown [in this link](https://coiled.io/blog/code-formatting-jupyter-notebooks-with-black/)

iameskild:
Hey @pierrotsmnrd, thanks for the feedback :) 
I was able to get it to work locally (not on qhub, that package was added last night and I haven't had a chance to test it yet). I will say that in order for me to get it to work, I had to restart my Jupyterlab server.

iameskild:
This PR has become a bit too unwieldily. I will be breaking it up into smaller, more manageable PRs.

PR #922, opened 2021-11-15
Setup/Add integration/deployment tests via pytest

aktech:
## Changes:

Sets up a simple framework for writing integration/deployment tests for the deployed QHub on minikube.
Currently added basic tests for Dask Gateway.

## Types of changes

- [ ] Documentation Update
- [x] Build related changes

## Testing

Added tests for running via pytest.

### Requires testing

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/environment-dev.yaml
+++ b/environment-dev.yaml
@@ -16,3 +16,7 @@ dependencies:
   - pytest
   - diagrams
   - jhub-client
+
+  # deployment/integration tests dependencies
+  - dask-gateway
+  - python-kubernetes
--- a/.github/workflows/kubernetes_test.yaml
+++ b/.github/workflows/kubernetes_test.yaml
@@ -148,7 +148,6 @@ jobs:
 
       - name: Deployment Pytests
         run: |
-          pip install dask-gateway kubernetes pytest dask-gateway
           pytest tests_deployment/ -v
 
       ### CLEANUP AFTER TESTS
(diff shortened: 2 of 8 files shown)


environment-dev.yaml
```diff
@@ -16,3 +16,7 @@ dependencies:
   - pytest
   - diagrams
   - jhub-client
+
+  # deployment/integration tests dependencies
+  - dask-gateway
+  - python-kubernetes
```
danlester, line 22: I think this should be `kubernetes` not `python-kubernetes` (which is very old, and doesn't work with the tests).

But also, note that `pip install -e .[dev]` does not make use of this environment-dev.yaml file, so really "kubernetes" and "dask-gateway" need to be added to the dev list in setup.py

danlester (changes requested):
By the way, I still think it would be worthwhile adding a Dask-based Jupyter notebook to be run during the Cypress stage :) But maybe another time...

environment-dev.yaml
aktech: The correct pip package is `kubernetes`, but the conda's `kubernetes` package didn't work for me yesterday and `python-kubernetes` seems to work, either ways I'll update the dev dependency.

danlester (approved):
Oh yes, I wasn't thinking about Conda. Either way, the new setup.py looks better for pip.

Issue #847, opened 2021-10-11
Split infrastructure into components


danlester:
`qhub deploy` currently splits the `infrastructure` deployment into stages using Terraform resource targeting.

This is 'wrong' according to Terraform, and we should consider splitting it up into separate configurations instead and rolling them out in order.

If we had clear separate configs then this would also mean we can `destroy` in reverse order, and I think this might help with our timeout problems on tear down. (We could technically do this with target anyway, but only if all remaining targets are listed explicitly - our final `deploy` step just runs everything, including the initially-targeted items too.)

From HCL docs:

> Once Terraform has selected one or more resource instances that you've directly targeted, it will also then extend the selection to include all other objects that those selections depend on either directly or indirectly.
> 
> This targeting capability is provided for exceptional circumstances, such as recovering from mistakes or working around Terraform limitations. It is not recommended to use -target for routine operations, since this can lead to undetected configuration drift and confusion about how the true state of resources relates to configuration.
> 
> Instead of using -target as a means to operate on isolated portions of very large configurations, prefer instead to break large configurations into several smaller configurations that can each be independently applied. Data sources can be used to access information about resources created in other configurations, allowing a complex system architecture to be broken down into more manageable parts that can be updated independently.

danlester:
Note I have attempted to improve `qhub destroy` using terraform targets in the meantime: https://github.com/Quansight/qhub/pull/948

costrouc:
Going to close this issue since it is implemented in #1003. We should definitely look into using terragrunt. Since right now the outputs from each module are being captured in python and being explicitly passed in via a var-files.

PR #948, opened 2021-11-30
qhub destroy using targets

danlester:
# Problem 

There are many reasons why `qhub destroy` fails, including timeouts in terraform which are hard to set globally.

Sometimes the Kubernetes cluster starts to be destroyed before Terraform has formally had a chance to destroy some of the software deployed within Kubernetes, resulting in the error that the cluster is not accessible to destroy that software.

Since the Keycloak provider needs to make API calls to Keycloak within the cluster, terraform command can fail terminally if the provider is configured with a Keycloak URL that no longer exists.

# Other Solutions

A more robust definition system as proposed in [Split infrastructure into components](#847) might help since it would allow destroy to be done in stages, so we can be sure one stage is complete before we advance to the next.

Maybe a move to Terraform CDK would also change this.

But these are long-term ideas.

# Current Solution

In this PR I have changed `qhub destroy` just to run the reverse of the targetted `qhub deploy` stages that we already have. One difficulty is in listing out all items that need to be destroyed at each stage - since deploy's final stage is just 'everything else', we need to maintain a long list of all items. I don't believe my list is complete, but it doesn't really matter - the destruction of Kubernetes should destroy everything anyway, and removing most items from the cluster in a separate stage gives breathing room so there is less to remove during the cluster destroy.

The main thing that we need to avoid telling Terraform to destroy e.g. "Kubernetes" and "keycloak-configuration" in the same stage, which of course happens when we just run a straight `terraform destroy` with no targetting.

An alternative approach to all this may have been to strengthen the 'depends_on' tree but this is split across multiple files and is still likely to run into the timeout problem.

Note it is not possible to 'refresh state' once Keycloak is inaccessible. We should consider removing the `terraform refresh` command from the destroy procedure anyway.

It is possible to run the old style of `qhub destroy` without targets  by using the flag `--full-only`.

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

### DIFF
--- a/qhub/cli/destroy.py
+++ b/qhub/cli/destroy.py
@@ -23,6 +23,11 @@ def create_destroy_subcommand(subparser):
         action="store_true",
         help="Disable auto-rendering before destroy",
     )
+    subparser.add_argument(
+        "--full-only",
+        action="store_true",
+        help="Only carry out one full pass instead of targeted sections",
+    )
     subparser.set_defaults(func=handle_destroy)
 
 
@@ -43,4 +48,5 @@ def handle_destroy(args):
     destroy_configuration(
         config,
         args.skip_remote_state_provision,
+        args.full_only,
     )
--- a/qhub/destroy.py
+++ b/qhub/destroy.py
@@ -7,10 +7,10 @@ from qhub.state import terraform_state_sync
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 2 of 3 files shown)


iameskild (approved):
LGTM

PR #957, opened 2021-12-03
No Users group by default

### LINKED ISSUE #921: Remove the default 'users' group?
## Description

Before Keycloak was added to QHub (now available in `main` branch), users and groups were specified in the qhub-config.yaml file. Every user had a `primary_group` key (plus optional `secondary_groups`). Every QHub installation was expected to have a group named `users`, and typically every user would be a member of this group. This wasn't enforced programmatically.

A uid had to be supplied for each user, and a primary gid was important under that system, due to Linux file permissions rules.

The new Keycloak functionality also includes a change that gives every user the name `jovyan` on the shared NFS disk, so every user has the same uid/gid and all files are owned by that user. Additionally, shared folders are mounted only corresponding to groups the user belongs to, so files can be shared privately amongst group members. (Previously, a shared folder was available for members of each group, but access was managed by Linux permissions.)

The uid and primary gid are now determined by the Docker image for JupyterLab.

As it stands in the new Keycloak functionality, a group named `users` will always be created, and it is set as a default group so all new Keycloak users will be members of it. Thus, the `users` folder will be available to everyone.

However, that requires extra configuration to be maintained. And it is no longer essential for users to be a member of at least one group.

My suggestion is to remove the automatic creation of the `users` group, so it is possible for a user to be a member of no groups.

Existing QHub installations that already have a `users` group would likely import that group when they upgrade to Keycloak. We would add documentation explaining that it can be set as a 'default group' in Keycloak if they want all new users to be added to that group automatically. 

Note we would still add the `admin` group automatically for all QHub installations, new and old.

## Value/benefit

This proposal removes unnecessary configuration and the legacy requirement for users to be a member of one group for Linux permissions purposes. For admins, it increases flexibility over groups and file sharing.

danlester:
Fixes #921 

## Changes:

- By default, there will be no `users` default group any more.
- There is now a `security.shared_users_group` field in `qhub-config.yaml` - if true, the `users` default group will be created. This will mean there is a shared folder for everyone to use.
- Added `--shared-users-group` flag to `qhub init` which will add `security.shared_users_group: true` to the `qhub-config.yaml` file. If the flag is omitted, the user will be asked if they want this `users` group or not (unless `--disable-prompt` passed, in which case shared_users_group defaults to false.
- Only symlink `shared` folder if there are any shared folder mounts available to the JupyterLab session.
- `qhub upgrade` from earlier versions will add `security.shared_users_group: true` if there was a `users` group specified in the incoming `qhub-config.yaml` file.

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No


danlester:
This is controlled by the `REQUIRE_USERS_GROUP="TRUE"` env var before running the backend. If TRUE, users cannot be removed from the `users` group, otherwise they can.

danlester:
Closed accidentally

viniciusdc:
HI Dan, thanks for the PR. I will test it! :+1:

viniciusdc:
Hi @danlester, sorry for the late feedback on this. I totally agree of the user group removal, but I didn't understand the optional user/shared settings... how that will interact with the recent changes proposed in #1003

danlester:
Yes, it's possible this is all outdated now.

Let's see once the other PR is merged.

PR #983, opened 2022-01-03
Add note to end of deployment

iameskild:
Fixes | Closes | Resolves #

> Please remove anything marked as optional that you don't need to fill in. Choose one of the keywords preceding to refer to the issue this PR solves, followed by the issue number (e.g Fixes # 666). If there is no issue, remove the line. Remove this note after reading.

## Changes:

-

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes
- [x] Other (please describe): small note to improve the user-experience during deployment

### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/qhub/deploy.py
+++ b/qhub/deploy.py
@@ -143,6 +143,14 @@ def guided_install(
     logger.info("Running Terraform Stage: FULL")
     terraform.apply(directory="infrastructure")
 
+    keycloak_docs_link = (
+        "https://docs.qhub.dev/en/stable/source/installation/login.html"
+    )
+
+    logger.info(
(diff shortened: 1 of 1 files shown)


danlester:
Thanks for this. Only thing is that at the moment `keycloak` and `keycloak.initial_root_password` are optional in the schema.

Maybe they shouldn't be, although an admin might like to remove initial_root_password from the YAML file after initial deployment, especially if they used a password they used elsewhere etc.

So anyway, the point is that it is possible your log message will throw an error if those fields are omitted.

Having said that, I wouldn't want to bet that everything functions correctly if you omit the fields and do `qhub deploy` anyway... It uses 'password' as a default where initial_root_password is not provided.

danlester (commented):
I commented on the main conversation thread.

iameskild:
Thanks for the info @danlester! The main reason I wanted to add this is due to the confusion around where the admin should go to change the `root` user password. I have added a basic check to ensure that this doesn't error out. 

(where is the main conversion thread?)

danlester:
Oh, I just added my original comment on the main PR, and then realised that maybe I was supposed to make a review, so I then just made a review with "I commented on the main conversation thread" in it. But it turns out GitHub just adds that to the same place anyway. So, basically just ignore my small second comment above!

Anyway, I still think the code doesn't quite take into account the schema. It is possible to have `security.keycloak` but not `security.keycloak.initial_root_password` in the YAML file.

I think you just need to do something like:

```
initial_password = config["security"].get("keycloak", {}).get("initial_root_password", "password")
```
And then just use `initial_password` to make up your log comment, which doesn't particularly need an `if` statement any more.

PR #984, opened 2022-01-05
Tweak node pool usage

leej3:
Allocates some stray cleaml resources to the clearml node-pool and targets the forwardauth deployment resource to the general node pool.

### DIFF
--- a/qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/modules/kubernetes/services/clearml/chart/values.yaml
+++ b/qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/modules/kubernetes/services/clearml/chart/values.yaml
@@ -226,6 +226,8 @@ redis: # configuration from https://github.com/bitnami/charts/blob/master/bitnam
   master:
     name: "{{ .Release.Name }}-redis-master"
     port: 6379
+    nodeSelector:
+      app: "clearml"
     persistence:
       enabled: true
       accessModes:
@@ -240,6 +242,8 @@ mongodb: # configuration from https://github.com/bitnami/charts/blob/master/bitn
     registry: docker.io
     repository: bitnami/mongodb
     tag: 3.6.21-debian-9-r71
+  nodeSelector:
+    app: "clearml"
   architecture: standalone
   auth:
     enabled: false
--- a/qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/kubernetes.tf
+++ b/qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/kubernetes.tf
@@ -421,6 +421,7 @@ module "forwardauth" {
   namespace    = var.environment
   external-url = var.endpoint
 
+  node-group = local.node_groups.general
   jh-client-id      = local.forwardauth-keycloak-client-id
   jh-client-secret  = random_password.forwardauth-jhsecret.result
   callback-url-path = local.forwardauth-callback-url-path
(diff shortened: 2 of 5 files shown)


viniciusdc (commented):
LGTM, I will present that in the next meeting with others and we can merge once they have a look as well. Thanks @leej3 for the contribuitions!!!

leej3:
Sounds good. Thanks. One thing to consider is whether to target some of the clearml pods using affinity as we have done in the case of forward-auth. On our deployment the clearml nodes use a lot of resources and  my understanding is that the clearml server needs quite a lot of resources (or at least we haven't explored what the minimum resources are). It may be more sensible to run all auxiliary pods somewhere other than the clearml node pool to be more efficient...

qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/modules/kubernetes/services/clearml/chart/values.yaml
```diff
@@ -226,6 +226,8 @@ redis: # configuration from https://github.com/bitnami/charts/blob/master/bitnam
   master:
     name: "{{ .Release.Name }}-redis-master"
     port: 6379
+    nodeSelector:
```
costrouc, line 229: How is the node being labeled "app: clearml"?

costrouc: Or this an assumption that the node will have this label?

leej3: > How is the node being labeled "app: clearml"

good point. It's a default value set in the variable.tf file that propagest to the chart's values.yaml (see [here](https://github.com/Quansight/qhub/blob/29f398af3542143751d52252b8a2204cd9c1e5cf/qhub/template/%7B%7B%20cookiecutter.repo_directory%20%7D%7D/infrastructure/modules/kubernetes/services/clearml/main.tf#L8))). 


> Or this an assumption that the node will have this label?

It is a requirement. We manually set this label for our deployment. We are not sure if this is automated when QHub deploys clearml when using the cloud providers (as described [here](https://docs.qhub.dev/en/stable/source/admin_guide/clearml.html?highlight=clearml#create-a-node-group)).

leej3:
Additionally, instead of targeting all clearml pods to the general pool, we want to be able to separately target the services to different node pools. The agent should run on a large node but all other services (including kube-system components) should be targeted to a general pool with nodes with less resources. @viniciusdc and myself will take a look. We'll submit this separately

qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/modules/kubernetes/services/clearml/chart/values.yaml
leej3: that drew our attention to missing redis/mongdb dynamic value setting. fixed now.

viniciusdc:
c.c @viniciusdc

costrouc:
@viniciusdc could you update this PR when you test clearml? I think this may already be fixed.

viniciusdc:
Based on a comparison between the main.tf files from [this PR](qhub/template/{{ cookiecutter.repo_directory }}/infrastructure/modules/kubernetes/services/clearml/main.tf) and the [current present on Qhub](https://github.com/Quansight/qhub/blob/main/qhub/template/stages/07-kubernetes-services/modules/kubernetes/services/clearml/main.tf), this was not fixed yet. So we need to move these changes to 0.4.0 standards then we can merge this. (after a new review).

As this will be part of #1217, I will move this to the same milestone.

viniciusdc:
I will be closing this due to the changes made for v0.4.0 and #1292. These changes will then be included once a PR for #1281 is merged. Thanks, @leej3 for pointing out the issue.

PR #993, opened 2022-01-14
Split kubernetes and kubernetes-initialization

### LINKED ISSUE #992: Kubeconfig state unavailable, Terraform defaults to localhost [bug]
Seen most often in the upgrade from 0.3.12 to main (0.4 candidate) of an AWS cluster, Terraform kubernetes provider sometimes tries to access the Kubernetes cluster at localhost instead of the EKS host.

The problem is with these two QHub Terraform modules: `kubernetes` and `kubernetes-initialization`.

The `kubernetes` module creates the cluster itself, and then `kubernetes-initialization` creates the namespace inside the cluster (plus some secrets).

`kubernetes-initialization` is configured like this:
```
provider "kubernetes" {
  host                   = module.kubernetes.credentials.endpoint
  cluster_ca_certificate = module.kubernetes.credentials.cluster_ca_certificate
  token                  = module.kubernetes.credentials.token
}
```

At the moment, there is a `terraform apply` step run targeting both these modules at the same time. The problem is that, if the state of the `kubernetes` module is not known (e.g. not refreshed in the right way at the right time), there is no way we can expect the `kubernetes-initialization` to have the appropriate provider configured correctly - because that depends on outputs of the earlier module.

We do set `kubernetes-initialization` to `depends_on` the `kubernetes` module, so `terraform apply` shouldn't try to create the namespace before the cluster exists, for example, but the provider needs to be configured at the start of the whole `terraform apply` step, otherwise it's as though we never passed through the settings such as `host` so it defaults to localhost.

The best description of this nuance is [probably here](https://github.com/hashicorp/terraform-provider-kubernetes/issues/1307#issuecomment-873089000).

Anyway, I will try splitting out the two modules into separate steps. That means an extra `terraform apply` call of course, and it might be better to see if we can instead push `kubernetes-initialization` into a joint step with the following module (`kubernetes-ingress`) if we set up appropriate `depends_on` hierarchies.

But in general, we need a proper multistage definition as per issue #847 .

danlester:
Fixes #992

### DIFF
--- a/qhub/deploy.py
+++ b/qhub/deploy.py
@@ -68,6 +68,15 @@ def guided_install(
     if not full_only:
         targets = [
             "module.kubernetes",
+        ]
+
+        logger.info(f"Running Terraform Stage: {targets}")
+        terraform.apply(
+            directory="infrastructure",
+            targets=targets,
+        )
+
+        targets = [
             "module.kubernetes-initialization",
         ]
 
--- a/qhub/destroy.py
+++ b/qhub/destroy.py
@@ -66,10 +66,15 @@ def destroy_configuration(config, skip_remote_state_provision=False, full_only=F
                     "targets": ["module.kubernetes-ingress"],
                 },
                 {
+                    "name": "Kubernetes Init",
+                    "targets": [
+                        "module.kubernetes-initialization",
+                    ],
+                },
+                {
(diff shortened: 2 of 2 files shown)


danlester:
I think this is superseded by PR #1003

PR #1017, opened 2022-02-03
Fix issue with traefik issueing certificates with letsencrypt acme

### LINKED ISSUE #1015: [bug] Fix traefik certificates from letsencrypt and tlsstore
## Description

With the recent changes made for #1003, we refactored the way certs are stored and passed over the qhub configuration (terraform modules, services..). This issue aims to fix any related problems that showed up regarding those changes.

costrouc:
The core of the issue was that for entrypoints a default certResolver
was not being specified. Additionally had to read some docs within
traefik https://doc.traefik.io/traefik/routing/entrypoints/#tls that
state that is a `tls = {}` section is specified within the crd
ingressroute the tls configuration is overwritten.

```
If a TLS section (i.e. any of its fields) is user-defined, then the
default configuration does not apply at all.
```

Fixes #1015 

## Changes:

- changes related to traefik ingress

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

## Testing

Tested locally using minikube and using digitalocean

### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

### DIFF
--- a/docs/source/admin_guide/index.md
+++ b/docs/source/admin_guide/index.md
@@ -15,6 +15,7 @@ jupyterhub.md
 keycloak.md
 clearml.md
 prefect.md
+traefik.md
 custom-helm-charts.md
 faq.md
 ```
--- a/docs/source/admin_guide/traefik.md
+++ b/docs/source/admin_guide/traefik.md
@@ -1,15 +1,15 @@
 # Traefik
 
-## Custom TLS Certificate
+## Custom TLS certificate
 
-### Creating the Certificate
+### Creating the certificate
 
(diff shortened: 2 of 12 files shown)


costrouc:
Still need to test with custom secret for compatibility. So do not merge until this is tested.

costrouc:
Tested and this appears to be working great with all three methods of creating certificates. This also appears to fix the dask-gateway issue when visiting the link. With the exception that self signed certificates are not trusted. This is due to issue https://github.com/thomseddon/traefik-forward-auth/issues/122.

viniciusdc (approved):

PR #1029, opened 2022-02-04
Fix input variable support for multiple types

### LINKED ISSUE #1028: [bug] Invalid value for input variable 'authentication'
## Describe the bug

0.4.0 test tracker: #976 

During deployment, qhub `0.4.0.dev64+g25b2eaa` on local `minikube` and `aws`, me and @iameskild noticed the following error during deployment when using Auth0/Github based auth:

```
INFO:qhub.provider.terraform:terraform apply directory=stages/06-kubernetes-keycloak-configuration targets=[]
INFO:qhub.provider.terraform: terraform at /tmp/terraform/1.0.5/terraform
[terraform]: ╷
[terraform]: │ Error: Invalid value for input variable
[terraform]: │ 
[terraform]: │   on /tmp/tmp_hmra46m.tfvars.json line 1:
[terraform]: │    1: {"realm": "qhub-thisisatest", "authentication": {"type": "Auth0", "config": {"client_id": "***************************", "client_secret": "***************************", "auth0_subdomain": "***************************"}}}
[terraform]: │ 
[terraform]: │ The given value is not valid for variable "authentication": all map
[terraform]: │ elements must have the same type.
[terraform]: ╵
```

 the above issue is due to the fact that terraform expects a `map` of one type variable for [`authentication`](https://github.com/Quansight/qhub/blob/25b2eaa95741224a24685bae8688f1aa8348d49e/qhub/template/stages/06-kubernetes-keycloak-configuration/variables.tf#L12) which is not the case, as its receiving from input a `string` and `map` as its values, see for reference [this issue/comment](https://github.com/hashicorp/terraform/issues/21384#issuecomment-494859782).

The following change will fix the above issue:
```
variable "authentication" {
  description = "Authentication configuration for keycloak"
  type = object({
    type = string
    config = map(string)
  })
}
```
edit. Due to some restrictions created with the usage of `object` as type, another more general option `type = any` is more advantageous.

## Expected behavior

- `authentication` variable should be correctly loaded into terraform

viniciusdc:
Fixes #1028 

The current `map` type for `authentication` lead to terraform to decide the type of variables it will be receiving from the input. But, when using extra forms of authentication like `Github, Auth0` the extra field `authentication.config` has a dict type structure, and `authentication.type` has an str type structure which lead to the following error when deploying the stage:

```
... map elements must have the same type.
```
edit.: Opted for variable type: `any` as using `object` structure obliges the variable input to have the same structure as specified which conflicts with the authentication modes of qhub (some does not include the `config` key)

## Changes:

- Update variable `authentication` type to include multiple value types.

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

Already tested on minikube local deployment with Auth0 config.

### DIFF
--- a/qhub/template/stages/06-kubernetes-keycloak-configuration/variables.tf
+++ b/qhub/template/stages/06-kubernetes-keycloak-configuration/variables.tf
@@ -15,4 +15,8 @@ variable "authentication" {
     type = string
     config = map(string)
   })
+  default = {
+    config = {}
+    type = ""
+  }
 }
--- a/qhub/template/stages/06-kubernetes-keycloak-configuration/variables.tf
+++ b/qhub/template/stages/06-kubernetes-keycloak-configuration/variables.tf
@@ -11,5 +11,8 @@ variable "keycloak_groups" {
 
 variable "authentication" {
   description = "Authentication configuration for keycloak"
-  type = map
+  type = object({
+    type = string
+    config = map(string)
+  })
 }
--- a/qhub/template/stages/06-kubernetes-keycloak-configuration/variables.tf
+++ b/qhub/template/stages/06-kubernetes-keycloak-configuration/variables.tf
@@ -11,12 +11,5 @@ variable "keycloak_groups" {
 
 variable "authentication" {
   description = "Authentication configuration for keycloak"
-  type = object({
-    type = string
-    config = map(string)
-  })
-  default = {
-    config = {}
-    type = ""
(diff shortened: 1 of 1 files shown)


costrouc (approved):
@viniciusdc makes perfect sense. Thank you for the detailed explanation on the problem.

PR #1039, opened 2022-02-09
Fix Black/Flake8 problems [WIP]

danlester:
Attempt to solve black problems seen on GitHub Actions but not CI.

We used to exclude cookie-cutter template folder from black, but this has obviously been replaced by template/stages which has a few random python files (e.g. jupyterhub_config.py).

These weren't safe to run through black before because they were jinja2 templates. It would be safe now, but I don't think it really makes sense to run black/flake8 on the template folder, so will submit a PR shortly.

## Changes:

- Exclude `qhub/template` from flake8 and black in both pre-commit and pyproject.toml

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -37,12 +37,12 @@ jobs:
       - name: Black Formatting
         run: |
           black --version
-          black qhub --diff
-          black --check qhub
+          black qhub --diff --exclude /qhub/template/
+          black --check qhub --exclude /qhub/template/
       - name: Flake8 Formatting
         run: |
           flake8 --version
-          flake8
+          flake8 qhub --exclude ./qhub/template/
       - name: Test QHub
         run: |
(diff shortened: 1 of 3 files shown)


danlester:
@iameskild Yes, it would be safer than when they were Jinja templates, but at the same time it seems a bit aggressive to set these Python formatting tools loose on large folder trees containing Terraform modules. There happen to be a few random Python files in there, but I would argue they are data files rather than Python files really :)

Let's say Black added a rule to insert `__init__.py` files into all folders on the assumption that we would only run Black on a Python tree in the first place. We wouldn't really want that for our Terraform modules. Arguably, the `templates` folder shouldn't be directly under `qhub` in the first place.

That's my thinking anyway...

PR #1047, opened 2022-02-11
prevent_deploy flag for safeguarding upgrades

### LINKED ISSUE #1040: Infrastructure will not upgrade from older version [bug]
## Describe the bug

As predicted, it is not possible to `qhub deploy` the latest QHub code (post #1003) directly on top of an existing older deployment (e.g. v0.3.14).

## Expected behaviour

Ideally, the QHub would be upgraded in place, retaining NFS data etc and also migrating the file system to use the new uid-less system.

## Actual behaviour

Deployment results in Terraform being completely unaware of the existing infrastructure, with no way to migrate it:
```
[terraform]: module.kubernetes.digitalocean_kubernetes_cluster.main: Creating...
[terraform]: ╷
[terraform]: │ Error: Error creating Kubernetes cluster: POST https://api.digitalocean.com/v2/kubernetes/clusters: 422 (request "7601f428-a99f-4459-be41-892e303f9a15") a cluster with this name already exists
```

## How to reproduce

After deploying 0.3.14 QHub, upgrade the qhub package, run `qhub upgrade` and then `qhub deploy`.

Tried with Digital Ocean, MacOS.

## Possible solutions

It might not be __impossible__ to map the states from the old QHub terraform structure to the new one, but that would be a very big job.

We probably need to guide the user through destroying and redeploying, including backups.

I would suggest `qhub upgrade` introduces as a `qhub-config.yaml` field such as `prevent_deployment: true` so that the user cannot move forward with the redeployment until they explicitly remove that line, indicating they know what they're doing and have backed up...

danlester:
Also fix --skip-remote-state-provision and remove --full-only which is no longer used)

Fixes #1040 

## Changes:

Added a `prevent_default` field to the `qhub-config.yaml` that will be set after `qhub upgrade` to 0.4.0. This stops users blindly redeploying without understanding the full upgrade process.

Also
- Fixed `--skip-remote-state-provision` which was being ignored
- Removed `--full-only` flag which was no longer used

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No

I have added general upgrade documentation but we also need to spell out the full upgrade process in release notes and/or docs.

### DIFF
--- a/docs/source/admin_guide/upgrade.md
+++ b/docs/source/admin_guide/upgrade.md
@@ -42,6 +42,10 @@ If you are deploying QHub from your local machine (not using CI/CD) then you wil
 qhub deploy -m qhub-config.yaml
 ```
 
+At this point you may see an error message saying that deployment is prevented due to the `prevent_deploy` setting in your YAML file. This is a safeguard to ensure that you only proceed if you are aware of possible breaking changes in the current upgrade.
+
+For example, we may be aware that you will lose data due to this upgrade, so need to note a specific upgrade process to keep your data safe. Always check the release notes of the release in this case and get in touch with us if you need assistance. For example, you may find that your existing cluster is intentionally deleted so that a new replacement can be deployed instead, in which case your data must be backed up so it can be restored after the upgrade.
+
 ### CI/CD: render and commit to git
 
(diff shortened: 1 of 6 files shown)


costrouc (approved):
Rebased to resolve merge conflicts. Otherwise PR looks great to me. Thanks @danlester for ensuring a safe upgrade to 0.4 :smile:

PR #1053, opened 2022-02-11
Move external_container_reg to earlier stage

### LINKED ISSUE #1052: external_container_reg needs to be in stage 03 not 08 [bug]
## Describe the bug

external_container_reg Terraform code must be defined earlier on so that we can make use of it earlier on, e.g. if JupyterHub or Keycloak image is coming from the private container registry.

Needs to be in stage 03 not 08.

As it stands, images will fail to pull.

danlester:
Fixes #1052 

## Changes:

Move external_container_reg Terraform sections from stage 08 to stage 03.

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No


costrouc (approved):
@danlester makes sense. Since I remember you will need a custom keycloak image etc. Thanks for catching this.

PR #1056, opened 2022-02-11
Adding security.shared_users_group option for default users group

costrouc:
Closes #957

This adds an option for `security.shared_users_group` which is by
default False. If True all users created after this option is set True
will be added to the `users` group. If any users were created before
this option they must manually be added to this group.


danlester (changes requested):
Code is fine I think, but not sure why the docs still refer to `users` and `groups` under `security` key in the YAML file (and also `primary_group` and `secondary_groups`). These have all been removed in favor of Keycloak managing them all in-cluster.

danlester (approved):

PR #1057, opened 2022-02-11
Depricating terraform_modules option since no longer used

costrouc:
Depricating terraform_modules option since no longer used. This was part of the original PR in #957 .


danlester:
I've added this now.

danlester (approved):

PR #1059, opened 2022-02-12
Adding clearml overrides

costrouc:
Closes #996

Adding this PR so that we can close the old PR that adds clearml
overrides from Vini.


viniciusdc (approved):
LGTM! Thanks @costrouc

PR #995, opened 2022-01-17
make the commit of the terraform rendering optional

leej3:
Combined with https://github.com/Quansight/qhub/pull/720 this enables deployment of QHub with gitops in such a way that secrets are not committed back to the repository.

PR #1068, opened 2022-02-16
Add CICD schema and render workflows

### LINKED ISSUE #1013: Add CI/CD pipelines for github and gitlab back
With the recent changes from #1003 we removed cookiecutter dependency over the programmatic way qhub was deployed and rendered. That broke the way those actions were generated, we will need to add those back and include `ruamel` for generating those "templates".

iameskild:
Fixes | Closes | Resolves #1013 

> Please remove anything marked as optional that you don't need to fill in. Choose one of the keywords preceding to refer to the issue this PR solves, followed by the issue number (e.g Fixes # 666). If there is no issue, remove the line. Remove this note after reading.

## Changes:

Much of this code was inspired by our existing `schema.py` and from conversations and discussion had with @costrouc and @tonyfast. Thank you both for your assistance and feedback. 

- Create GitHub-Action workflow schema 
- Create workflows based on schema

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [x] No

Not yet...


iameskild:
@costrouc workflow files being deprecated:
- image-pr.yaml
- image.yaml
- jupyterhub-pr.yaml
- jupyterhub.yaml

costrouc (approved):

costrouc:
I also see the version issue on `pip install qhub==...` which I think is only showing itself with a development version.

iameskild:
@costrouc keeping it in json is fine with me! 
Thinking through `pip install qhub==...` version issue, this might make it a little tricky to test since the deployment version of qhub won't be on pypi. Perhaps we can add some logic to see if the version of qhub is "dirty" and if so install from github otherwise from pypi.

costrouc:
I'm fine for now just merging this work and leaving it to another PR to figure this out for development. For an actual Pypi release this won't be a problem

costrouc (approved):

viniciusdc (approved):
Amazing, the fact that we are using schema for it makes a good difference when testing in the future. Amazing work @iameskild, have you tested the gitlab actions?

qhub/provider/cicd/github.py
```diff
@@ -82,3 +88,219 @@ def create_repository(owner, repo, description, homepage, private=True):
...
+    return GHA_job_step(name="Install QHub", run=pip_install_qhub(qhub_version))
+
+
+def gen_qhub_ops(config):
+
+    env_vars = gha_env_vars(config)
+    branch = config["ci_cd"]["branch"]
+    qhub_version = config["qhub_version"]
+
+    push = GHA_on_extras(branches=[branch], path=["qhub-config.yaml"])
```
viniciusdc, line 220: are we considering #995 here?

qhub/utils.py
```diff
@@ -349,3 +349,15 @@ def deep_merge(*args):
         return [*d1, *d2]
     else:  # if they don't match use left one
         return d1
+
+
+def pip_install_qhub(qhub_version: str) -> str:
```
viniciusdc, line 354: good catch :+1:

qhub/provider/cicd/github.py
iameskild: I didn't see this but I think won't be too hard. Do you think we should include it here?

qhub/utils.py
iameskild: @costrouc this is unsophisticated way of handling the qhub_version issue for now. Would this be okay to add?

qhub/provider/cicd/github.py
viniciusdc: I think so, as it was linked to the original issue and it would be just adding an `if` for `step5` based on this [new variable](https://github.com/Quansight/qhub/blob/426c610e042c345c1d2f7ba78aa70595bda0d587/qhub/render/__init__.py#L13)

qhub/utils.py
costrouc: Yeup this works with me.

costrouc (approved):
@iameskild looks great to me. After merging this we can test this and impove upon it when we find issues.

PR #1073, opened 2022-02-18
Adds the ability on render to deleted targeted files or directories

### LINKED ISSUE #1049: [enhancment] clear out stages before re-render
As pointed out by @danlester, we should remove some files from the stages directory before re-attempting a qhub deploy (redeploy), as it would be more secure for the user as well as would decrease the chance of terraform complaining or erroring during deployment. e.g. 

> for example if an existing tfstate file exists in stage 01 from before changing the project name, qhub deploy will error because it wants to refresh the older version of the state (if that old bucket was deleted).

As per #1010  currently, we have a way to discern which files exist only in the user space (now, [untracked](https://github.com/Quansight/qhub/blob/main/qhub/render/__init__.py#L99)), so we can add a filter to remove some target files like the state ones.

### Enhcanemtns

- Removal of .tfstate files from stages in each re-render
- Add ability to the user to opt for removal, maybe a target arg? e.g 'qhub render --remove [01-terraform-state/<untracked-file>]'
- Notify user when creating and modifying/removing files (skip it with `--disable-prompt` as well)

costrouc:
Closes #1049

This is to provide a simple mechanism for deprication of files and
allowing for targeted deletion of files. Later this functionality can
be extented to provide qhub extensions the ability to depricate files
within a stage. For how this is just a single list that is populated.

### DIFF
--- a/qhub/render/__init__.py
+++ b/qhub/render/__init__.py
@@ -76,7 +76,7 @@ def render_template(output_directory, config_filename, force=False, dry_run=Fals
 
     source_dirs = [os.path.join(str(template_directory), _) for _ in directories]
     output_dirs = [os.path.join(str(output_directory), _) for _ in directories]
-    new, untrack, updated = inspect_files(
+    new, untracked, updated, deleted = inspect_files(
         source_dirs,
         output_dirs,
         source_base_dir=str(template_directory),
@@ -86,7 +86,14 @@ def render_template(output_directory, config_filename, force=False, dry_run=Fals
             ".terraform.lock.hcl",
             "terraform.tfstate.backup",
         ],
-        ignore_directories=[".terraform"],
+        ignore_directories=[
+            ".terraform",
+            "__pycache__",
+        ],
+        deleted_paths=[
+            'infrastructure',
+            'terraform-state',
+        ],
         contents=contents,
     )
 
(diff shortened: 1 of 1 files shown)


qhub/render/__init__.py
```diff
@@ -86,7 +86,14 @@ def render_template(output_directory, config_filename, force=False, dry_run=Fals
             ".terraform.lock.hcl",
             "terraform.tfstate.backup",
         ],
-        ignore_directories=[".terraform"],
+        ignore_directories=[
+            ".terraform",
+            "__pycache__",
+        ],
+        deleted_paths=[
```
tonyfast, line 93: i think i'm worried that defining these paths deep in the code might wake it hard to manage these deprecations. it seems that this is code style as it is. we might want to tear this out into a `deprecate.py` file or something.

tonyfast (commented):
looks like it will do the trick, but i'm slightly concerned `deleted_files` might be a confusing name. `deleted` raises 🎏  and `deprecated_files` or `deprecated_conventions` has a softer blow.

qhub/render/__init__.py
costrouc: Thanks for the review @tonyfast. What do you think about the recent additions? I've added an explicit DEPREICATED_FILE_PATHS constant within `qhub/depricate.py`. Eventually I'd like to add this logic to our extension mechanism to allow for stages to depricate files.

costrouc: I'd like to keep the `deleted_paths` bit just to communicate that some files will actually be deleted.

viniciusdc (approved):
All right, the deletion logic has no issues. LGTM! Thanks @costrouc

costrouc:
@danlester would you mind reviewing this if it does all you need in #1049

danlester:
This is certainly capable of fixing the problems we've seen, but is probably overkill in terms of expecting developers to use this system mid-release - which is where we've encountered recent problems.

Ultimately, the advice for developers is to clear everything out periodically.

But we will be able to use this if we keep track of files changed between releases, to avoid the same problems out in the wild.

costrouc:
> This is certainly capable of fixing the problems we've seen, but is probably overkill in terms of expecting developers to use this system mid-release - which is where we've encountered recent problems.

> Ultimately, the advice for developers is to clear everything out periodically.

Yeup I agree. It's also go practice for us to delete the directory to make sure that the deployment does not depend on the files and hidden state.

This PR is there to address issues that user's of qhub will face and make as minimal deletions as possible.

Issue #1081, opened 2022-02-19
[bug] Qhub fails to destroy on AWS instance


viniciusdc:
## Describe the bug

Fresh install of qhub (version `0.4.0.dev86+g9fb62c8` based on `main:9fb62c812e`), failed to destroy instance: 
```
[terraform]: module.network.aws_subnet.main[1]: Still destroying... [id=subnet-0e67818d0783e1d0c, 20m1s elapsed]
[terraform]: module.network.aws_subnet.main[0]: Still destroying... [id=subnet-057575eebaf199725, 20m1s elapsed]
[terraform]: 2022-02-18T20:24:16.347-0300 [INFO]  provider.terraform-provider-aws_v3.73.0_x5: 2022/02/18 20:24:16 [WARN] WaitForState timeout after 20m0s: timestamp=2022-02-18T20:24:16.347-0300
[terraform]: 2022-02-18T20:24:16.347-0300 [INFO]  provider.terraform-provider-aws_v3.73.0_x5: 2022/02/18 20:24:16 [WARN] WaitForState starting 30s refresh grace period: timestamp=2022-02-18T20:24:16.347-0300
[terraform]: 2022-02-18T20:24:16.371-0300 [INFO]  provider.terraform-provider-aws_v3.73.0_x5: 2022/02/18 20:24:16 [WARN] WaitForState timeout after 20m0s: timestamp=2022-02-18T20:24:16.371-0300
[terraform]: 2022-02-18T20:24:16.371-0300 [INFO]  provider.terraform-provider-aws_v3.73.0_x5: 2022/02/18 20:24:16 [WARN] WaitForState starting 30s refresh grace period: timestamp=2022-02-18T20:24:16.371-0300
[terraform]: ╷
[terraform]: │ Error: error deleting EC2 Subnet (subnet-0e67818d0783e1d0c): DependencyViolation: The subnet 'subnet-0e67818d0783e1d0c' has dependencies and cannot be deleted.
[terraform]: │ 	status code: 400, request id: a936bad1-4605-4b70-9fc4-25c0dae06131
[terraform]: │ 
[terraform]: │ 
[terraform]: ╵
[terraform]: ╷
[terraform]: │ Error: error detaching EC2 Internet Gateway (igw-00f5928d9219a02fa) from VPC (vpc-027cf7b4d3f134cc1): DependencyViolation: Network vpc-027cf7b4d3f134cc1 has some mapped public address(es). Please unmap those public address(es) before detaching the gateway.
[...]

Issue #865, opened 2021-10-15
[EPIC] Nebari Extension Mechanism


costrouc:
## Description

The are many projects that we would like to integrate into Neabri and additionally other projects that we may not have the time/resources to support in Nebari. We need an extension mechanism to integrate projects that a loosely connected to Nebari. For me prefect and clearml are good examples of this. I would also like to see if this could be adopted within Nebari for some parts.

## Suggestion

Features that the extension mechanism should have.

  - allows us to expose the service at https://<custom-domain>/<custom-prefix>/
  - optionally setup oauth2 client for service
  - provide an "overrides" section so that we can easily control the helm chart
  - helm chart information 

Example

```yaml
extensions:
  - name: Prefect
    chart: <path-to-chart-url>
    version: <chart-version>
    domain: <optional domain name>
    prefix: <prefix-url>
    oauth2_client: true/false
    overrides:
       ...
```

Lets start a discussion about this. We need to use approach as well for a few components.

viniciusdc:
Hi @Adam-D-Lewis, any updates from the Prefect team on this?

costrouc:
From the development of the stages in PR #1003 I would like to propose a general "extension" mechanism that makes all parts of Qhub an extension. Take for example the following from `deploy.py`

```
def provision_01_terraform_state(stage_outputs, config):
     ...
     elif config['provider'] == 'gcp':
        stage_outputs[directory] = terraform.deploy(
            terraform_import=True,
            directory=os.path.join(directory, config['provider']),
            input_vars={
                'name': config['project_name'],
                'namespace': config['namespace'],
                'region': config['google_cloud_platform']['region']
            },
            terraform_objects=[
                QHubGCPProvider(config),
            ],
            state_imports=[(
                "module.terraform-state.module.gcs.google_storage_bucket.static-site",
                f"{config['project_name']}-{config['namespace']}-terraform-state",
            )])
    ...     
```

Later when we look at `stages/02-infrastructure/gcp` we see a terraform module /directory that is called by this function. There is a lot being done in the `terraform.deploy(...)` function but there are important sections seen here.

1. `input_vars` how do values in `qhub_config.yaml` and outputs from prior stages `stage_outputs` map to imput variables to the terraform module.
2. `state_imports` these are ids/attrs that can be used such that terraform can "import" these resources if they already exist. This allows for qhub to be more stateless. We have used this in the past to import terraform-state in the `stages/01-terraform-state`.
[...]

PR #1091, opened 2022-02-22
Reorganizing render, deploy, destroy to unify stages input_vars, tf_objects, checks, and state_imports

costrouc:
This PR is only a refactor of render, deploy, and destroy. The motivation for this is allowing for the closing of #1081. In order to properly delete the aws ingress we need to be able to get the terraform outputs from some early kubernetes stages. This PR will also allow for the easier transition to #865.

There are no modifications to the behavior of the code.

Going into more detail about the heart of the issue:

For each stage we have to supply input variables, dymanic terraform objects, terraform imports. 

We cannot run `terraform output` without supplying the proper input variables. Within deploy.py and destroy.py this was becoming repetative. Thus I've moved the definitions into a `stages/*.py` directory which will need to be done with whatever extension mechanism we chose.

### DIFF
--- a/qhub/deploy.py
+++ b/qhub/deploy.py
@@ -1,72 +1,18 @@
-import socket
-import time
 import logging
 import os
-import sys
-import contextlib
-from subprocess import CalledProcessError
-from typing import Dict
 import tempfile
-import json
 import textwrap
-from urllib.parse import urlencode
+import subprocess
+from typing import Dict
+import contextlib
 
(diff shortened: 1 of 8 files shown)


danlester (approved):
This is a great step forward!

At some point maybe we will refactor using classes to enforce the structure and reduce repetition of strings such as `stages/03-kubernetes-initialize`.

PR #1094, opened 2022-02-22
fix bug in state_01_terraform_state function call

tylerpotts:
Fixes a bug that was introduced where the state_01_terraform_state function call was given too many arguments resulting in this error:

```
TypeError: stage_01_terraform_state() takes 1 positional argument but 2 were given
```

### DIFF
--- a/qhub/deploy.py
+++ b/qhub/deploy.py
@@ -63,7 +63,7 @@ def provision_01_terraform_state(stage_outputs, config):
             terraform_import=True,
             directory=os.path.join(directory, config["provider"]),
             input_vars=input_vars.stage_01_terraform_state(stage_outputs, config),
-            state_imports=state_imports.stage_01_terraform_state(stage_outputs, config),
+            state_imports=state_imports.stage_01_terraform_state(config),
         )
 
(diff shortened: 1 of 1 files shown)


costrouc (changes requested):
Whoops that was a bug the function should take `(stage_outputs, config)` as inputs. Since this is done during the deploy stage.

tylerpotts:
@costrouc Why is that? I'm not seeing any references to `stage_outputs` in the function itself: 

https://github.com/Quansight/qhub/blob/main/qhub/stages/state_imports.py

costrouc:
@tylerpotts this is because I'm trying to enforce an interface to these methods.
 - render stage is a function of `config`
 - deploy stage is a function of `config`, `stage_outputs`
 - destroy stage is a function of `config`, `stage_outputs`

This will make it so when we create an extension mechanism we have enough inputs to be dynamic enough for what we might need. Anything that is done during deploy should have access to those inputs.

tylerpotts:
@costrouc Makes sense, thanks for the explanation. I'll make those changes

PR #1118, opened 2022-02-25
Adding tests to visit additional endpoints

### LINKED ISSUE #1121: [bug] Conda Store auth redirects via http not https
When logging into Conda Store, the redirect back from Keycloak seems to get confused between HTTP and HTTPS.

This can cause blocked content in the browser (in particular, within Cypress tests).

This is the flow:

```
https://qhub.mycompany.com/auth/realms/qhub/login-actions/authenticate?session_code=...
Returns a 302 to https://qhub.mycompany.com/conda-store/oauth_callback?state=...
Which returns a 308 to http://qhub.mycompany.com/conda-store/oauth_callback/?state=... (note http)
Which returns a 308 to http://qhub.mycompany.com/conda-store/oauth_callback/?state=...
For some reason this loads https://qhub.mycompany.com/conda-store/oauth_callback/?state=... (instead of http)
Which returns a 302 to http://qhub.mycompany.com/conda-store/user/
Which returns a 302 to http://qhub.mycompany.com/conda-store/user/
For some reason this loads https://qhub.mycompany.com/conda-store/user/
```

Where I say "for some reason this loads https", I think that's related to [Upgrade-Insecure-Requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Upgrade-Insecure-Requests) provided by the browser's request headers, and [the CSP response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/upgrade-insecure-requests) - however, I suspect that will all be irrelevant if we work to ensure no HTTP is involved in the first place.

I'm not sure if this is a problem inside Conda Store or within QHub's integration of Conda Store.

costrouc:
This adds routes to visit the keycloak user account page, conda-store, and grafana monitoring just to check that pages are up

Closes #1121

### DIFF
--- a/qhub/keycloak.py
+++ b/qhub/keycloak.py
@@ -15,6 +15,7 @@ def do_keycloak(config_filename, *args):
 
     # supress insecure warnings
     import urllib3
+
     urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
     keycloak_admin = get_keycloak_admin_from_config(config)
@@ -27,7 +28,7 @@ def do_keycloak(config_filename, *args):
 
         username = args[1]
         password = args[2] if len(args) >= 3 else None
-        create_user(keycloak_admin, username, password, domain=config['domain'])
+        create_user(keycloak_admin, username, password, domain=config["domain"])
     elif args[0] == "listusers":
         list_users(keycloak_admin)
     else:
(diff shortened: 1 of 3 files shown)


danlester:
I have added waits, and checks that something useful shows up, for the new tests.

Conda-store login currently fails in Cypress. After clicking 'Sign in with OAuth' it hangs because some insecure content is blocked.

This is documented in [Conda Store auth redirects via http not https](https://github.com/Quansight/qhub/issues/1121)

Maybe there's an easy way to fix that, but of course the alternative would be just to not click the sign in button for now - i.e. remove `.click();` and the subsequent `cy.get...` line.

danlester (approved):

PR #1119, opened 2022-02-25
FIX: Remove Conda Store from default images

### LINKED ISSUE #1104: [BUG] - conda-store image does not upgrade correctly to 0.4
### OS system and architecture in which you are running QHub

MacOS, any cloud

### Expected behavior

Run `qhub upgrade -c qhub-config.yaml` on a yaml file from v0.3.14, aiming to upgrade to v0.4.

Before, we had `default_images.conda-store: quansight/qhub-conda-store:v0.3.14`, and we end up with `quansight/qhub-conda-store:v0.4.0` - but that doesn't exist.

For contrast, a new `qhub init` gives `default_images.conda-store: quansight/conda-store-server:v0.3.9` - which is the correct, working image.

### Actual behavior

We are given the wrong image after upgrade, and this won't work with the latest qhub.

Run `qhub upgrade -c qhub-config.yaml` on a yaml file from v0.3.14, aiming to upgrade to v0.4.

Before, we had `default_images.conda-store: quansight/qhub-conda-store:v0.3.14`, and we now get `quansight/qhub-conda-store:v0.4.0` - but that doesn't exist.

### Command output

```bash session
qhub upgrade -c qhub-config.yaml
```

Qhub version in main

### Compute environment

AWS

### Integrations

_No response_

### Anything else?

_No response_

HarshCasper:
Fixes #1104

### DIFF
--- a/qhub/initialize.py
+++ b/qhub/initialize.py
@@ -38,7 +38,6 @@ BASE_CONFIGURATION = {
         "jupyterlab": f"quansight/qhub-jupyterlab:{qhub_image_tag}",
         "dask_worker": f"quansight/qhub-dask-worker:{qhub_image_tag}",
         "dask_gateway": f"quansight/qhub-dask-gateway:{qhub_image_tag}",
-        "conda_store": "quansight/conda-store-server:v0.3.9",
     },
     "storage": {"conda_store": "60Gi", "shared_filesystem": "100Gi"},
     "theme": {
--- a/qhub/schema.py
+++ b/qhub/schema.py
@@ -121,7 +121,6 @@ class DefaultImages(Base):
     jupyterlab: str
     dask_worker: str
     dask_gateway: str
-    conda_store: str
 
 
 # =========== Authentication ==============
--- a/qhub/stages/input_vars.py
+++ b/qhub/stages/input_vars.py
@@ -215,7 +215,7 @@ def stage_07_kubernetes_services(stage_outputs, config):
         "conda-store-environments": config["environments"],
         "conda-store-storage": config["storage"]["conda_store"],
         "conda-store-image": _split_docker_image_name(
-            config["default_images"]["conda_store"]
+            "quansight/conda-store-server:v0.3.9"
         ),
         # jupyterhub
         "cdsdashboards": config["cdsdashboards"],
--- a/qhub/template/stages/07-kubernetes-services/modules/kubernetes/services/conda-store/variables.tf
+++ b/qhub/template/stages/07-kubernetes-services/modules/kubernetes/services/conda-store/variables.tf
@@ -36,7 +36,7 @@ variable "conda-store-image" {
   })
   default = {
     name = "quansight/conda-store-server"
-    tag  = "v0.3.8"
+    tag  = "v0.3.9"
   }
 }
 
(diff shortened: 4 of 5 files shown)


danlester (changes requested):
This is a good start, but the changes to the upgrade process have been misunderstood.

Please completely reinstate the `Upgrade_0_3_12` class. Since 0.3.12 is an older version, there is no need to change that at all. At that point in time, for an upgrade to 0.3.12 we needed to add the `default_images.conda-store` field. We still do, in theory, for any upgrade to 0.3.12. (Although, of course, anyone running `qhub upgrade` with the latest code will actually be upgrading to 0.4.0. However, the way it works is that it can still upgrade from e.g. 0.3.10 to 0.4.0 by going through every intermediate step.)

So, anyway, what I really meant was that in `Upgrade_0_4_0` you need to do the opposite of what we did in `Upgrade_0_3_12`. We need to remove the `default_images.conda-store` key from the config.

If you look at `Upgrade_0_4_0` you can see we do some similar things there. We remove the `security.users` and `security.groups` fields, for example. We also remove a key called `terraform_modules`.

So have a think about how you could just add something similar to remove `default_images.conda-store` if it exists (which it should, but best to check).

And then we also need `qhub deploy` to work of course, so we need to think about how we can not pass in the `default_images.conda-store` field to the Terraform code, and just have it have a non-overridable default for the image within the Terraform code. Where we pass the image variable in now, we just need to ensure the variable is no longer available and we just have a raw string.

danlester (approved):
That will work fine, and I can approve. However, it is just worth noting that you could remove either:
```
        "conda-store-image": _split_docker_image_name(
            "quansight/conda-store-server:v0.3.9"
        ),
```
from `input_vars.py`
OR
```
  default = {
    name = "quansight/conda-store-server"
    tag  = "v0.3.9"
  }
```
from `conda-store/variables.tf`

You don't actually need both! However, it is perfectly acceptable how you have done it because you could argue that you want the separation between the script in input_vars.py and the Terraform module for conda-store. Maybe you think that the Terraform module should be able to be self-contained (and still have a default image) in case whatever calls it doesn't want to pass through the image name, for any reason.

Of course, right now the two sections of code are tightly-coupled and always located close by in the same relatively-small codebase.

So on the other hand, you might not like the way you have essentially duplicated the number of places where you specify the default image name and tag. Imagine if we need to change the conda-store image within the QHub code (e.g. to v.3.10 when it comes out). You might only remember to do it in one place, at which point you might be specifying v0.3.9 in input_vars but v.0.3.10 in variables.tf - in which case you will end up still using v0.3.9 because it overrides the default. Or you might end up with it the other way round: specify v0.3.10 in input_vars but v0.3.9 in variables.tf, in which case you would end up using v0.3.10 after all, for the same reason - that input_vars overrides the default set in variables.tf.

HarshCasper:
Hi @danlester 

```sh
[terraform]: │ Error: Reference to undeclared input variable
[terraform]: │ 
[terraform]: │   on modules/kubernetes/services/conda-store/worker.tf line 101, in resource "kubernetes_deployment" "worker":
[terraform]: │  101:           image = "${var.conda-store-image.name}:${var.conda-store-image.tag}"
[terraform]: │ 
[terraform]: │ An input variable with the name "conda-store-image" has not been declared.
[terraform]: │ This variable can be declared with a variable "conda-store-image" {} block.
[terraform]: ╵

Problem encountered: Terraform error
```

It seems like we would have to anyway have a Terraform variable here and we cannot remove the duplication entirely. We can possibly revisit it at some other time.

PR #1123, opened 2022-02-26
Adding tests for juypterhub-ssh, jhub-client, and vs code 

costrouc:
From my testing jupyterhub-ssh does not fully work as an expected ssh session. Currently it is hanging at times with 

```
/opt/conda/condabin/conda
/opt/conda/condabin/conda
```

And then hanging and not giving a shell.

### DIFF
--- a/.github/workflows/kubernetes_test.yaml
+++ b/.github/workflows/kubernetes_test.yaml
@@ -173,7 +173,15 @@ jobs:
             ./tests_e2e/cypress/screenshots/
             ./tests_e2e/cypress/videos/
 
+      - name: Deployment Pytests
+        run: |
+          export KEYCLOAK_USERNAME=example-user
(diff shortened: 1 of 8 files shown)


costrouc:
From my testing jupyterhub-ssh does not fully work as an expected ssh session. Currently it is hanging at times with 

```
/opt/conda/condabin/conda
/opt/conda/condabin/conda
```

And then hanging and not giving a shell.

danlester (approved):

costrouc:
@danlester I agree. This would make the tests faster. But I don't know how we can get around this since in the UI tests we want to test the `start` and `stop` button. So this will require that at a minimum we start and stop twice.

danlester:
Yes, this was one reason I built the Python tests into Cypress initially - click to Start, then run the tests, then stop!

But as long as multiple notebooks can be done in one session, maybe that's OK.

costrouc:
The `pytest` `test_deployments` will also need to launch servers but this was added with this PR recently

PR #1130, opened 2022-03-01
Upgrade conda-store v0.3.10 and simplify specification of image

### LINKED ISSUE #1126: Upgrade conda-store v0.3.10 and simplify specification of image
In fixing #1104 through PR #1119, we were left with multiple locations where the conda-store image was specified.

@costrouc 's attempt to upgrade conda-store from v0.3.9 to v0.3.10 within PR #1118 was actually ineffective because of this duplication.

@HarshCasper plans to refactor further to avoid future confusion, and of course in the process land on conda-store v0.3.10 as the true default.

HarshCasper:
Fixes #1126

### DIFF
--- a/qhub/stages/input_vars.py
+++ b/qhub/stages/input_vars.py
@@ -214,9 +214,6 @@ def stage_07_kubernetes_services(stage_outputs, config):
         # conda-store
         "conda-store-environments": config["environments"],
         "conda-store-storage": config["storage"]["conda_store"],
-        "conda-store-image": _split_docker_image_name(
-            "quansight/conda-store-server:v0.3.9"
-        ),
         # jupyterhub
         "cdsdashboards": config["cdsdashboards"],
         "jupyterhub-theme": config["theme"]["jupyterhub"],
--- a/qhub/template/stages/07-kubernetes-services/conda-store.tf
+++ b/qhub/template/stages/07-kubernetes-services/conda-store.tf
@@ -9,15 +9,6 @@ variable "conda-store-storage" {
   type        = string
 }
 
-variable "conda-store-image" {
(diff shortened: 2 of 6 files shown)


danlester:
@HarshCasper Oh no, the image `quansight/conda-store-server:v0.3.10` still appears in two places!

Sorry, I didn't realise that both `conda-store-server` and `conda-store-worker` use the same image... So I might have pointed you in the right direction.

Can you find another way to do all of this so that, essentially, the string `quansight/conda-store-server:v0.3.10` (or related objects) appears only once?

There are many approaches you could take.

(1) Maybe reinstate all the Terraform variables but with _no defaults_ in the Terraform variable definitions, and then pass through `quansight/conda-store-server:v0.3.10` from `input_vars.py` as we did before.

(2) Or keep the string removed from `input_vars.py` and reinstate the Terraform variables plus a default only at one level - probably in `conda-store.tf` you would add the default {...} image setting, but make sure you do not also set the default in `conda-store/variables.tf`.

Or there are other possibilities. Personally I think (2) suits us best at this time - we will easily be able to override the image from Python code in the future if we want to.

In (2) this means you can't use the lowest level module (`modules/kubernetes/services/conda-store` or whatever as a folder) without explicitly passing through a `conda-store-image` value as we will do from `conda-store.tf`. But that's fine - the module is unlikely to be used separately from its surrounding codebase so we don't need to worry about it, and in any case whatever uses it can just pass through the value explicitly.

danlester (changes requested):
Changes requested as above.

danlester:
Close #1121

costrouc (approved):

danlester (approved):

PR #1138, opened 2022-03-02
Completely removing mention of jovyan from images, fix nss home dirctory

### LINKED ISSUE #1137: [DOC] - Explain jovyan user directory
Considering the perspective of someone who may not be familiar with JupyterHub, or maybe even Jupyter, the docs should explain what the jovyan user directory is for. 

This can be a problem when setting up ssh keys for GitHub.  `ssh-keygen` puts the new key in the `/home/jovyan/.ssh` directory.  However, someone not familiar with the jovyan directory will look for the key in the `/home/<user>/.ssh` directory (but it's not there).

Related: #1076

costrouc:
Addresses the issue in #1137. Will close #1137 after testing it it verifies that the problem has been solved. Hopefully this will remove jovyan


costrouc:
Closing since I'd like to break this into smaller PRs. The idea is that we should be simplifying the images as much as possible.

Issue #1136, opened 2022-03-01
[BUG] - ssh key missing


ericdatakelly:
### OS system and architecture in which you are running QHub

Ubuntu 20.04.3 LTS (Focal Fossa)

### Expected behavior

- Use `ssh-keygen` to generate a new ssh key
- The ssh key remains in the `.ssh` folder after I logout and back in

### Actual behavior

- After doing the usual steps to generate an ssh key, the key was written to the `/home/jovyan/.ssh` directory and I was able to use git normally to commit and push to GitHub.
- When starting the ssh-agent, this error came up, but the key worked.

```
$ eval "$(ssh-agent -s)"
ERROR: ld.so: object 'libnss_wrapper.so' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
Agent pid 194
```

- I logged off for a while.
- I logged back in and my ssh key was missing and I did not have GitHub access.

### How to Reproduce the problem?

- In a new user account, add an ssh key using the following commands

```
ssh-keygen -t ed25519 -C "your_email@example.com"  # writes to the jovyan ssh directory
eval "$(ssh-agent -s)"
ssh-add /home/jovyan/.ssh/id_ed25519
cat /home/jovyan/.ssh/id_ed25519.pub
```

- Test the key by cloning a github repo, committing, and pushing the changes back to github
- log out
- log in
- Try pushing another commit to the same repo

### Command output

```bash session
$ ssh-keygen -t ed25519 -C "e...@gmail.com"
Generating public/private ed25519 key pair.
Enter file in which to save the key (/home/jovyan/.ssh/id_ed25519): 
Enter passphrase (empty for no passphrase): 
Enter same passphrase again: 
Your identification has been saved in /home/jovyan/.ssh/id_ed25519
Your public key has been saved in /home/jovyan/.ssh/id_ed25519.pub
The key fingerprint is:
SHA256:fYi38...txcI e...@gmail.com
The key's randomart image is:
+--[ED25519 256]--+
|     ..  ..      |
|      .+.+       |
| .. . ..= .      |
|o oE + * o .     |
|+o* + * S = o    |
|.X + . . = + . . |
|=.+ . . . o   o  |
| o.  +.    o  .. |
|   ...o.    .. ..|
+----[SHA256]-----+
$ eval "$(ssh-agent -s)"
ERROR: ld.so: object 'libnss_wrapper.so' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.
[...]

PR #1142, opened 2022-03-03
[BUG] Removing jovyan home directory and issue with nss confiuration

### LINKED ISSUE #1137: [DOC] - Explain jovyan user directory

costrouc:
Closes #1137 #1136

### DIFF
--- a/.github/workflows/kubernetes_test.yaml
+++ b/.github/workflows/kubernetes_test.yaml
@@ -196,10 +196,14 @@ jobs:
       - name: Cleanup qhub deployment
         run: |
           cd local-deployment
-          # This is a known failure at the moment.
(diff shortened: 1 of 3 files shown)


viniciusdc:
I tested on Azure and it seemed to solve the issue

PR #1153, opened 2022-03-08
Readthedocs and documentation updates

tonyfast:
currently readthedocs is failing on builds because qhub isn't installed. this pull request fixes that issue and introduces local documentation build support with nox.

when i make this pull requests, readthedocs should build us some docs or give us errors that hint at fixes.


tonyfast:
alrighty, the build is green (https://readthedocs.org/projects/qhub-cloud/builds/16306372/) and can deploy (https://docs.qhub.dev/en/dox-fixes/).

magsol:
Awesome!! Thanks so much @tonyfast 💪

docs/requirements.txt
```diff
@@ -1,20 +1 @@
...
-pydata-sphinx-theme
-pytest
-sphinx_rtd_theme
-sphinx-autoapi
-sphinx-copybutton
-sphinx-copybutton
-sphinx-material
-sphinxcontrib-bibtex
-stringcase
+jupyter-book
```
viniciusdc, line 1: Hi @tonyfast, is this enough for building the docs? I noticed that it has all the sphinx-related dependencies that we might need, but is there anything else we need to add as well?

noxfile.py
```diff
@@ -0,0 +1,8 @@
+from nox import session
+
+
+@session(reuse_venv=True)
+def docs(session):
+    session.install("-e.")
+    session.install("-rdocs/requirements.txt")
+    session.run("sphinx-build", "docs", "_build/html")
```
viniciusdc, line 8: What's the difference between the previous build workflow and this one?

docs/requirements.txt
tonyfast: jupyter-book has a ton of dependencies. most of them are to incorporate markdown/notebooks as sphinx documentation source. all those prior deps are taken care of by jupyter-book.

noxfile.py
tonyfast: sorry i missed that description. i added instructions to `docs/README.md` for building the docs. you'll need `nox`, `pip install nox`. the run the `nox` `docs` session as:

     nox -s docs

`nox` is a tool to ensure that `sphinx` builds in its own virtual environment, and doesn't effect the developer's primary environments.  its specifically good for something like `jupyter-book` that has a lot of dependencies that are better isolated from other environments. also, it provides a convenient cli that can work locally/ci.

iameskild (approved):
This looks great to me! I ran `nox -s docs` locally and was able to view the docs (`_build/html/index.html`).

It seems like we might want to add `nox` as a dev dependency, what do y'all think?

tonyfast:
we could add it as a dev dependency, but that might make the most sense if we can get linting and testing to use `nox` too.

PR #1171, opened 2022-03-18
Integrate kind into local deployment to no longer require minikube for development

costrouc:
This PR aims to significantly simplify development of qhub for local deployment using kind. Kind will also allow us to test multinode setups locally.

This PR is a work in progress and to me uses the actual meaning of `provider: local` implying that a local kubernetes cluster will be spun up. I suggest we change the prior `provider: local` -> `provider: existing`.

The load balancer is automatically configured in the `qhub deploy` stage `02-infrastructure/local` meaning that the script/metallb ... is no longer required. Also make the only assumption for a local deployment that there is docker.


costrouc:
This PR will not work until we move the images to a separate repo and have them pre built

viniciusdc:
@iameskild how is the work with the docker images? is this unblocked now?

iameskild:
@viniciusdc the [PR to transfer images](https://github.com/nebari-dev/nebari-docker-images/pull/1) has been submitted, reviewed and is awaiting final approval :)

viniciusdc:
As the docker images are now transferred over, we can work on this again. I've prioritized integrating this as part of https://github.com/nebari-dev/nebari-docker-images/issues/2. This will help split up the dependencies for a full integrated Kubernetes tests from a Jupyterlab/Jupyterhub one (anything that only requires a simple ecosystem built)

viniciusdc:
@costrouc could you have a look? I will change add the existing name change in a separate PR later.

.github/workflows/kubernetes_test.yaml
```diff
@@ -51,65 +54,54 @@ jobs:
...
-      - name: Print minikube and kubectl versions
+
+          docker info
+          docker ps
+      - name: Pull docker images
         run: |
-          minikube version
-          kubectl version
-      - name: Use minikube docker daemon
+          images=${{ env.image_names }}
```
costrouc, line 78: I think this is double work. We should remove all the `docker pull` and `kind load` since kubernetes will pull as needed

viniciusdc:
Hi @costrouc, I added a new deployment mode to `qhub deploy` that copies over the previous behavior of `local.` I've tested the new `existing` with Minikube and all succeed (deploy and destroy).

costrouc (commented):
@viniciusdc thanks for pushing this PR through the finish line. This looks good to me. My concern is that we will need to mention the breakage of local -> existing but we will do this in the docs and I do not believe that many groups are depending on this feature.

PR #1176, opened 2022-03-22
Adding examples to QHub jupyterlab 

### LINKED ISSUE #1175: [ENH] - Add examples files for QHub jupyterlab and dashboards
### Feature description

Examples that would be useful:
  - using dask
  - using dashboards
     - panel
     - plotly
     - streamlit
     - voila

### Value and/or benefit

There are several ways this would benifit:
 - demos
 - testing
 - good user getting started

### Anything else?

_No response_

costrouc:
Closes #1175 

Effort was put into making these examples as small as possible. Totally agree that down the road we need to put more effort into these examples.

Also switched to using [fileset](https://www.terraform.io/language/functions/fileset) in the config maps so that we do not have to explicitely include each file. fileset will grab all the files.


costrouc:
I have tested manually that all these dashboards work along with the dask gateway example.

PR #1183, opened 2022-03-23
Upgrade mrparkers to 3.7.0

danlester:
I repeatedly saw problems when trying to deploy locally with minikube.

The qhub keycloak realm seemed to fail to be created, causing the subsequent keycloak config to fail ("Realm not found")

This PR seems to make things better, although I'm not sure of the root cause. I also think related Terraform problems can still occur later on in non-keycloak components, but we will have to keep an eye on that separately.

minikube v1.25.2 on Ubuntu 20.04 (amd64)
Using the docker driver
Kubernetes v1.23.3 on Docker 20.10.12 ...
All in a Parallels VM on Mac

- Upgrade mrparkers/keycloak to 3.7.0
- Set some defaults in keycloak_realm

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

### DIFF
--- a/qhub/template/stages/06-kubernetes-keycloak-configuration/main.tf
+++ b/qhub/template/stages/06-kubernetes-keycloak-configuration/main.tf
@@ -1,14 +1,36 @@
 resource "keycloak_realm" "main" {
-  provider = keycloak
 
   realm        = var.realm
   display_name = var.realm_display_name
+  
+  direct_grant_flow    = "direct grant"
+  enabled              = true
+  browser_flow         = "browser"
+  revoke_refresh_token = false
+  user_managed_access  = false
+  ssl_required         = "external"
+  registration_flow    = "registration"
+    
+  refresh_token_max_reuse    = 0
+  reset_credentials_flow     = "reset credentials"
+  client_authentication_flow = "clients"
+  docker_authentication_flow = "docker auth"
+
+  offline_session_max_lifespan_enabled = false
+  
+  web_authn_policy { 
+  }
+  
+  web_authn_passwordless_policy {
+  }
+
 }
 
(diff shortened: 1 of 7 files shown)


qhub/template/stages/06-kubernetes-keycloak-configuration/main.tf
```diff
@@ -1,14 +1,36 @@
...
+  web_authn_passwordless_policy {
+  }
+
 }
 
 resource "keycloak_group" "groups" {
   for_each = var.keycloak_groups
   realm_id = keycloak_realm.main.id
-  name     = each.value
+  name     = each.key
```
costrouc, line 32: FYI `each.value` and `each.key` are equivalent when working with sets.

This will overwrite the attributes for the group and set it to empty. Is this intentional? I know you said you were using attributes for some of your components.

costrouc (approved):
Looks great to me!

qhub/template/stages/06-kubernetes-keycloak-configuration/main.tf
danlester: To be honest, most of these changes were inspired by TF_LOG=DEBUG output from Terraform. It seemed to be getting confused about the values it was setting on each run - e.g. the difference between missing values and null values made it want to make changes that weren't really going to have any effect, so I'm just trying to set the defaults in a way that causes both terraform and mrparkers to consider them the same. So, yes, it's all intentional but basically guesswork that I believe helped in some way... I'm not too sure whether future deployments will remove all group attributes actually. We can change this in the future if it causes problems.

PR #1203, opened 2022-03-25
Profilegroups

### LINKED ISSUE #1191: [ENH] - Keycloak Group attributes to control JupyterLab profile access
### Feature description

This functionality has been dropped in the recent refactor.

If a Keycloak group had a `profile` attribute containing a list of JupyterLab profiles then members of that group would be able to access the profiles. This check only came into effect for profiles defined with `groups:` (which could be empty) in the `qhub-config.yaml` file. That choice was because profiles defined with no `groups:` (or `users:`) would be available to all users - i.e. no access control - so this ability to override dynamically by setting Keycloak group attributes is not wanted for those groups.

I would like to reinstate this functionality. It is possible that some usage of roles in keycloak would be better, but as a short term solution at least, this functionality can be easily reinstated by adding a mapper to the jupyterhub client as follows:

Screenshot 2022-03-24 at 12 33 13

We would also need to reinstate the code that checks for this in `jupyterhub_config.py` or similar.

I propose we change the attribute name from `profiles` to `jupyterlabprofiles` though.

### Value and/or benefit

This allows profiles to be assigned to groups dynamically using Keycloak without a full redeploy.

### Anything else?

_No response_

danlester:
Fixes #1191 

`profiles.jupyterlab` in `qhub-config.yaml` now has an `access` field - can be `all`, `keycloak`, `yaml`
- If `yaml`, also specify `groups` and/or `users` fields to restrict access to those users
- If `all` everyone will have access
- If `keycloak` access is granted to users/groups with the Keycloak attribute `jupyterlabprofiles` containing the `display_name` in their lists

Added `qhub upgrade` mechanism for this.

Group attributes are no longer overwritten on non-initial deploys (previously the pre-existing groups would have attributes reset).

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

## Documentation

- [ ] Yes, docstrings
- [x] Yes, main documentation

### DIFF
--- a/qhub/schema.py
+++ b/qhub/schema.py
@@ -41,6 +41,12 @@ class AuthenticationEnum(str, enum.Enum):
     custom = "custom"
 
 
+class AccessEnum(str, enum.Enum):
+    all = "all"
+    yaml = "yaml"
+    keycloak = "keycloak"
+    
+
 class Base(pydantic.BaseModel):
     ...
 
@@ -290,12 +296,22 @@ class KubeSpawner(Base):
 
 
 class JupyterLabProfile(Base):
+    access: AccessEnum = AccessEnum.all
     display_name: str
     description: str
     default: typing.Optional[bool]
     users: typing.Optional[typing.List[str]]
     groups: typing.Optional[typing.List[str]]
     kubespawner_override: typing.Optional[KubeSpawner]
+    
+    @root_validator
+    def only_yaml_can_have_groups_and_users(cls, values):
+        if values["access"] != AccessEnum.yaml:
+            if values.get("users",None) is not None or values.get("groups",None) is not None:
+                raise ValueError(
+                    f"Profile must not contain groups or users fields unless access = yaml"
+                )
+        return values
 
 
 class DaskWorkerProfile(Base):
@@ -313,7 +329,7 @@ class Profiles(Base):
     jupyterlab: typing.List[JupyterLabProfile]
     dask_worker: typing.Dict[str, DaskWorkerProfile]
 
-    @validator("jupyterlab", pre=True)
+    @validator("jupyterlab")
     def check_default(cls, v, values):
         """Check if only one default value is present"""
         default = [attrs["default"] for attrs in v if "default" in attrs]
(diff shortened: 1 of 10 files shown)


qhub/schema.py
```diff
@@ -290,13 +296,26 @@ class Config:
 
 
 class JupyterLabProfile(Base):
+    access: AccessEnum = AccessEnum.all
```
costrouc, line 299: I agree with the default value to keep this backwards compatible.

costrouc (approved):
I don't have any issues with how this PR was written. Makes perfect sense and I appreciate the default value of `all` for the jupyterlab profiles. Also like how you used the oauth client and added the variable for controlling if jupyterlab profiles are exposed to the application. We can add other mappers in the future with this pattern.

My only issue is the name `jupyterlabprofiles` could we name this something with a delimiter e.g. `jupyterlab-profiles` or `jupyterlab_profiles`. The reason is I envision that we will want to use this for other profiles in the future. Like dask. Dask-Gateway can follow this exact same pattern of what you have implemented.

Longer term I would like to think of a way that we could manage profiles outside of the configuration `qhub-config.yaml` it feels like a bad fit and an area where developers will often want to change and modify.

costrouc (approved):
I don't have any issues with how this PR was written. Makes perfect sense and I appreciate the default value of `all` for the jupyterlab profiles. Also like how you used the oauth client and added the variable for controlling if jupyterlab profiles are exposed to the application. We can add other mappers in the future with this pattern.

My only issue is the name `jupyterlabprofiles` could we name this something with a delimiter e.g. `jupyterlab-profiles` or `jupyterlab_profiles`. The reason is I envision that we will want to use this for other profiles in the future. Like dask for `dask_profiles` or `dask-profiles`. Dask-Gateway can follow this exact same pattern of what you have implemented.

Longer term I would like to think of a way that we could manage profiles outside of the configuration `qhub-config.yaml` it feels like a bad fit and an area where developers will often want to change and modify.

costrouc:
Flaky internet posted twice :upside_down_face:

danlester:
Good idea - I have added an underscore.

PR #1222, opened 2022-04-01
add `retrolab` to the base jupyter image

tonyfast:
this pull request adds `retrolab` to the base jupyter environment. `notebook>=7` will deploy a classic version of the notebook that is based on `jupyterlab` components. til then `retrolab` needs to be explicitly installed. likely, we'll remove this change in the future. in the meantime, it allows us to understand the feel or `retrolab` in common workflows.


iameskild (approved):
Added this as a profile on quansight-beta and it seems to work as expected, very nice :)

PR #1224, opened 2022-04-05
Update qhub-dask pinned version

### LINKED ISSUE #1212: [BUG] - Dask environment not working due to recent package release
### OS system and architecture in which you are running QHub

Linux

### Expected behavior

When executing the simple dask-gateway [pipeline ](https://docs.qhub.dev/en/stable/source/user_guide/dask_gateway.html?#starting-a-cluster) we should be able to susceffuly launch a dask scheduler pod and a cluster instance.Succesful

### Actual behavior

Executing the `cluster = gateway.new_cluster()` block from the above docs, leads to an unexpected error during spawning of the Dask scheduler. Upon further inspection of the scheduler logs, we can see the same error message as shown in [this issue](https://github.com/dask/distributed/issues/6013).

Even though we currently pin the necessary version so dask within `qhub-dask` meta-package, we still need to add a new exception to avoid the new version of `click` which was the root cause of the problem.

image

### Possible Solution
Add the following to the dask environment upon initialization 
```yaml
 - click ~= 8.1.0 # Or downgrade and pin to 8.0.2
```

In the description

### Command output

_No response_

Latest qhub on main, but the environment build is independent of cli version

### Compute environment

_No response_

### Integrations

_No response_

### Anything else?

_No response_

iameskild:
Fixes | Closes | Resolves #1212

- Update qhub-dask from `0.3.13` to `0.4.0`

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [x] Yes
- [ ] No

CI should be sufficient.

- [ ] Yes
- [x] No

## Documentation

- [ ] Yes, docstrings
- [ ] Yes, main documentation

### DIFF
--- a/qhub/initialize.py
+++ b/qhub/initialize.py
@@ -15,6 +15,7 @@ from qhub.provider import git
 from qhub.utils import (
     namestr_regex,
     qhub_image_tag,
+    qhub_dask_version,
     check_cloud_credentials,
     set_kubernetes_version,
 )
@@ -211,7 +212,7 @@ DEFAULT_ENVIRONMENTS = {
             "python",
             "ipykernel",
             "ipywidgets",
-            "qhub-dask ==0.3.13",
+            f"qhub-dask =={qhub_dask_version}",
             "python-graphviz",
             "numpy",
             "numba",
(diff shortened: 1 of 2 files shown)


iameskild:
This is a solution that will work without much overhead however there are other ways we can get this to work without having to manually update the version of qhub-dask each time we cut a new release. 
Something like `qhub_dask_version = __version__` would work well for those installing qhub from pypi or conda-forge but will not work well for developers since there will not be a version of `qhub-dask` that corresponds to `__version__`.

I created #1226 to circle back to some of these idaes.

costrouc (approved):

PR #1235, opened 2022-04-07
Support for pinning the IP address of the load balancer via terraform overrides

### LINKED ISSUE #1210: [ENH] - Support for pinning the IP address of the load balancer and ability to use internal VPC
### Feature description

For Traefik we currently ask Kubernetes to create a load balancer for us without having any control over the ip that is given or whether it uses an ip address within the current VPC or external (external is default). We have a PR that does this for AWS https://github.com/Quansight/qhub/pull/650 but never merged it in due to not being totally sure on how we are going to support a feature like this. 

For me the conflicting problem is how are we going to support all these options in the `qhub-config.yaml` and ensure that they are all documented. I'd like to use this issue as a motivation to having a key `terraform_overrides` which get passed to the terraform module.

The idea would be to have the following in `qhub-config.yaml` where `ingress.terraform_overrides` is an arbitrary nested dictionary. This feature is similar to a prior feature we have of passing overrides to helm charts.

```yaml
ingress:
  terraform_overrides:
      load_balancer_annotations:
          "kubernetes.io/role/internal-elb": 1
      load_balancer_ip: "1.2.3.4"
```

This will need to be added to the https://github.com/Quansight/qhub/blob/main/qhub/schema.py along with the documentation that this new key is available.

Next we need to make sure that we are passing these overrides to the https://github.com/Quansight/qhub/tree/main/qhub/template/stages/04-kubernetes-ingress stage where ingress is being deployed. This is done via adding to https://github.com/Quansight/qhub/blob/main/qhub/stages/input_vars.py#L144 and passing anything in `terraform_overrides` as input.

Finally need to add these new variables to https://github.com/Quansight/qhub/blob/main/qhub/template/stages/04-kubernetes-ingress/variables.tf and properly pass in the values to https://github.com/Quansight/qhub/blob/main/qhub/template/stages/04-kubernetes-ingress/modules/kubernetes/ingress/main.tf#L58.

### Value and/or benefit

Enterprise support often requires deploying within an existing VPC and pinnning the ip address.

### Anything else?

_No response_

aktech:
Fixes | Closes | Resolves #1210 

- Adds support for terraform overrides in ingress's load balancer.

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

## Documentation

- [ ] Yes, docstrings
- [ ] Yes, main documentation

### DIFF
--- a/qhub/schema.py
+++ b/qhub/schema.py
@@ -382,6 +382,15 @@ class QHubExtension(Base):
     envs: typing.Optional[typing.List[QHubExtensionEnv]]
 
 
+class IngressTerraformOverrides(Base):
+    load_balancer_annotations: typing.Optional[typing.Dict]
+    load_balancer_ip: str
+
+
+class Ingress(Base):
+    terraform_overrides: IngressTerraformOverrides
+
+
 # ======== External Container Registry ========
 
 # This allows the user to set a private AWS ECR as a replacement for
@@ -455,6 +464,7 @@ class Main(Base):
     prevent_deploy: bool = (
         False  # Optional, but will be given default value if not present
     )
+    ingress: Ingress
 
     # If the qhub_version in the schema is old
     # we must tell the user to first run qhub upgrade
--- a/qhub/schema.py
+++ b/qhub/schema.py
@@ -382,13 +382,8 @@ class QHubExtension(Base):
     envs: typing.Optional[typing.List[QHubExtensionEnv]]
 
 
-class IngressTerraformOverrides(Base):
-    load_balancer_annotations: typing.Optional[typing.Dict]
-    load_balancer_ip: str
-
-
 class Ingress(Base):
-    terraform_overrides: IngressTerraformOverrides
+    terraform_overrides: typing.Any
 
 
 # ======== External Container Registry ========
--- a/qhub/schema.py
+++ b/qhub/schema.py
@@ -464,7 +464,7 @@ class Main(Base):
     prevent_deploy: bool = (
         False  # Optional, but will be given default value if not present
     )
-    ingress: Ingress
+    ingress: typing.Optional[Ingress]
 
     # If the qhub_version in the schema is old
     # we must tell the user to first run qhub upgrade
--- a/qhub/stages/input_vars.py
+++ b/qhub/stages/input_vars.py
@@ -142,6 +142,9 @@ def _calculate_note_groups(config):
 
 
 def stage_04_kubernetes_ingress(stage_outputs, config):
+    ingress_terraform_overrides = config.get("ingress", {}).get(
+        "terraform_overrides", {}
+    )
     return {
         "name": config["project_name"],
         "environment": config["namespace"],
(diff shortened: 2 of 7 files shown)


qhub/schema.py
```diff
@@ -382,6 +382,15 @@ class QHubExtension(Base):
     envs: typing.Optional[typing.List[QHubExtensionEnv]]
 
 
+class IngressTerraformOverrides(Base):
+    load_balancer_annotations: typing.Optional[typing.Dict]
+    load_balancer_ip: str
+
+
+class Ingress(Base):
+    terraform_overrides: IngressTerraformOverrides
```
costrouc, line 391: I'd prefer if this is `terraform_overrides: Any` and we defer to terraform for the checks.

qhub/stages/input_vars.py
```diff
@@ -152,6 +155,10 @@ def stage_04_kubernetes_ingress(stage_outputs, config):
         "certificate-secret-name": config["certificate"]["secret_name"]
         if config["certificate"]["type"] == "existing"
         else None,
+        "load-balancer-annotations": ingress_terraform_overrides.get(
```
costrouc, line 158: Using what I showed above we just supply `terraform_overrides` e.g. `**config.get("ingress", {}).get("terraform_overrides": {})`. Terraform variables will do all the checking for us.

costrouc (changes requested):
I agree with all the terraform modifications. Would like the python side of things to be more flexible. This allows us keep the schema for qhub simpler.

qhub/schema.py
costrouc: So no need for the `IngressTerraformOverrides`

aktech: The problem with this is, "Any" won't reach terraform, as we parse these in the `input_vars.py`, so anything other than this won't be passed, unless we pass them explicitly.

aktech: Ah, I see your comment below now.

qhub/stages/input_vars.py
aktech: The problem is, we need `load-balancer-ip` and `load-balancer-annotations` separately, to be passed to different variables, so that they can be used at their designated places:

In `qhub/template/stages/04-kubernetes-ingress/modules/kubernetes/ingress/main.tf`

spec:

```hcl
  spec {
    selector = merge({
      "app.kubernetes.io/component" = "traefik-ingress"
    }, var.load-balancer-annotations)

```

and 

```hcl
    type = "LoadBalancer"
    load_balancer_ip = var.load-balancer-ip
```

costrouc: I think this should work. Suppose

```yaml
ingress:
  terraform_overrides:
      load-balancer-annotations:
           k1: v1
           k2: v2
      load-balancer-ip: 1.2.3.4
```

Then in `input-vars.py` we supply:

```
    return {
        "name": config["project_name"],
        "environment": config["namespace"],
        "node_groups": _calculate_note_groups(config),
        "enable-certificates": (config["certificate"]["type"] == "lets-encrypt"),
        "acme-email": config["certificate"].get("acme_email"),
        "acme-server": config["certificate"].get("acme_server"),
        "certificate-secret-name": config["certificate"]["secret_name"]
        if config["certificate"]["type"] == "existing"
        else None,
        **config.get("ingress", {}).get("terraform_overrides": {})
    }

```

This way `terraform_overrides` key can override any setting. Does this make sense?

aktech: > This way `terraform_overrides` key can override any setting.

Yes, any variable present in the terraform. This makes more sense. I'll add this.

costrouc (approved):

PR #1259, opened 2022-04-22
[WIP] Add support for VPC deployment for GCP via terraform overrides

### LINKED ISSUE #1211: [ENH] - Allow for deploying managed kubernetes within VPC for each cloud provider
### Feature description

Related to #1210. In #1210 we are adding support for deploying the load balancer within an existing VPC. However this does not completely solve the problem in that the kubernetes infrastructure itself must also be deployed within the vpc. This is a common issue that we have seen on multiple client deployments. Currently our workaround is to require the client deploys a managed kubernetes cluster and we use the `local` deployment option to deploy onto the existing kubernetes cluster. 

We however do not want to require this. In order to support deploying within an existing vpc we need to add some conditional bits to the `02-infrastructure` stage. This will require custom terraform for each provider so I recommend starting with a cloud provider that you are most comfortable with. I'd suggest adding a `terraform_overrides` key to each providers key. E.g.:
  - `google_cloud_platform.terraform_overrides`, `amazon_web_services.terraform_overrides`, etc. following the same pattern as #1210. The goal here is to limit the number of options exposed in the configuration while allowing us to have all these knobs in the terraform modules.

My guess for gcp would be that the qhub-config.yaml would look like:

```
google_cloud_platform:
    terraform_overrides:
       vpc_id: ....
```

These overrides would be passed to https://github.com/Quansight/qhub/blob/main/qhub/template/stages/02-infrastructure/gcp/variables.tf.

### Value and/or benefit

Ability to deploy qhub within existing VPCs on all cloud providers.

### Anything else?

_No response_

aktech:
Fixes | Closes | Resolves #1211 

- Defines terraform override variables to able to deploy in VPC for GCP.

## Types of changes

- [x] Documentation Update
- [ ] Build related changes

### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Documentation

- [ ] Yes, docstrings
- [ ] Yes, main documentation

### DIFF
--- a/docs/source/installation/configuration.md
+++ b/docs/source/installation/configuration.md
@@ -722,6 +722,8 @@ This is quite useful for pinning the IP Address of the load balancer.
 
 ### Deployment inside Virtual Private Network
 
+#### Azure
+
 Using terraform overrides you can also deploy inside a virtual private network.
 
 An example configuration for Azure is given below:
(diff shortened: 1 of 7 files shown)


costrouc (approved):

costrouc:
This looks great to me @aktech and I was able to test this out locally

PR #1265, opened 2022-05-03
Github actions broken for new repos: rename github secrets so github will be happy again

tylerpotts:
Github no longer allows repository secrets that begin with `GITHUB_` : 
image

This fixes that


viniciusdc:
Hi, @tylerpotts thanks for the contribution. We are currently migrating Qhub to Nebari org, and some of the configurations here would be affected, we will be closing this PR for now and I will open an issue to track these changes. I would love to see your PR after the migration.

PR #1007, opened 2022-01-25
Create new lint.yaml

iameskild:
Fixes | Closes | Resolves #
#1005 

> Please remove anything marked as optional that you don't need to fill in. Choose one of the keywords preceding to refer to the issue this PR solves, followed by the issue number (e.g Fixes # 666). If there is no issue, remove the line. Remove this note after reading.

## Changes:

- Create a new `lint.yaml`

## Types of changes

What types of changes does your code introduce?

_Put an `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds a feature)
- [ ] Breaking change (fix or feature that would cause existing features to not work as expected)
- [ ] Documentation Update
- [x] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Other (please describe):

### Requires testing

- [ ] Yes
- [ ] No

### In case you checked yes, did you write tests?

- [ ] Yes
- [ ] No

## Further comments (optional)

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered and more.

PR #972, opened 2021-12-16
[WIP] CI/Chore: Linting YAML on GHA

HarshCasper:
## Description

This PR:

- Adds a GHA step to lint YAML files
- Adds a `.yamllint` to define rules and configurations

PR #1282, opened 2022-05-14
Update pre-commit, remove vale

iameskild:
*Think of this PR as more of a discussion.*

## Theme (or goals)
- move linting and formatting as close to the developer as possible (`pre-commit`) for quicker feedback that is NOT reliant on CI.
- on the road to a more consistent and manageable CI 

## Description and motivations for these changes:
- Completely remove Vale and replace it with `codespell` pre-commit hook.
  - Although Vale is a far more sophisticated of spell checker than `codespell`, the benefits of using `codespell` over Vale include:
    - `codespell` has a very low false positive rate and only makes suggestions/fixes for common words, skipping technical terms. This means we don't have to continuously maintain a dictionary like we do for Vale. 
    - `codespell` can make the fixes automatically (using `pre-commit`, with `codespell --write` enabled). 
    - `codespell` can run in files besides just `.md`, see `setup.cfc::codespell` for files currently excluded.
    - And perhaps the biggest reason for this change is to move spell checking closer to the user (not CI reliant). 
      - If Vale had a pre-commit hook, this might not be necessary but Vale doesn't. And forcing developers/writers to download, config and run `vale` from the command line seems like it adds a lot friction for new contributors. 
- Clean up and update `pre-commit`.
  - Add `codespell` pre-commit hook, as mentioned above.
  - Add `terraform_fmt` pre-commit hook to format Terraform files according the [Terraform style guidelines](https://www.terraform.io/language/syntax/style). 
  - Add `isort` pre-commit hook to cleanly sort python imports.
  - Add `check-json` pre-commit hook to check json files ;)
  - There are also a few minor updates to the existing pre-commit hooks.
    - The most noticeable change is moving as many of the configuration settings into either `pyproject.toml`, `setup.cfg` or within the `.pre-commit-config.yaml` itself. Files as such `.hadolint.yaml` and `.mdformat.toml` can subsequently be removed. 
- Completely remove `docs.yaml` (Documentation Linter) Github Action workflow.
  - If we remove Vale, and if are fine removing the `link-checker` for the time being, then we can completely remove this workflow. 
    - I personally find the current `link-checker` to be exceptionally flaky. It feels like it reports false positives on every other one of my doc related PRs. As a result, I've spent more time than I'd like running down upstream bugs, making minor changes to the `link-checker` GHA or simply ignoring the failed CI.
- Create `run-pre-commit.yaml` which hands all linting and formatting. 
  - This workflows simply runs `pre-commit run --all-files` to ensure all of our files are always properly linted / formatted. Triggered for every PR and for every push to `main`. 

These proposed changes are ultimately the result of the frustration I've felt while writing or updating docs, consistently getting errors from Vale or the link-checker or both. Like I mentioned above, I think Vale is definitely a more comprehensive spell checker (and style checker) but the added benefits don't outweigh the friction it causes the development workflow especially compared to a baseline using `codespell`.

If accepted, we can close #1007 #972.

I would love to hear others have to say about this propose change :)

cc. @costrouc @tonyfast @magsol @viniciusdc @HarshCasper @aktech @kcpevey @trallard


iameskild:
I intentionally committed with `--no-verify` so that we can focus on my proposed changes and not get lost in all the noise created from the pre-commit hooks themselves. 

Have a look at the [GHA logs for `run-pre-commit.yaml`](https://github.com/Quansight/qhub/runs/6431237731?check_suite_focus=true) to see which files will be affected by the pre-commit hooks:

viniciusdc:
Amazing work @iameskild, just finshed looking trought the changes, still need to know the end result of the action. But I liked this organization, one thing that I suggest is to add the following `if` condition to the `pre-commit` job: 
```if: github.event.pull_request.merged == false```
this will skip the workflow to run when the PR is merged (to avoid running twice)

**Comment by viniciusdc** [comment]
Also, not sure about what the others think of this but we could add a message to the PR automatically in case the pre-commit action fails, explaining to the user (if it's a new user) about pre-commit and how to install it, something like this:
```yaml
...
    steps:
      - uses: mshick/add-pr-comment@v1
        with:
          message: |
            First of all, thank you for taking the time to contribute! :tada: 
            It seems that some tests are falling, please refer to the pre-commit logs for more details or check if pre-commit is 
            correctly installed. If you're a first-time contributor, you can follow this steps to automatically set up pre-commit and 
            avoid these errors.
          repo-token: ${{ secrets.GITHUB_TOKEN }}
          repo-token-user-login: 'github-actions[bot]' # The user.login for temporary GitHub tokens
          allow-repeats: false # This will avoid duplicating comments
```

We can add a link to some step-by-step installation from pre-commit itself, or add extra information in the comment.

.github/workflows/docs.yml
```diff
@@ -1,39 +0,0 @@
...
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Broken link checker
-        uses: gaurav-nelson/github-action-markdown-link-check@v1
-        with:
-          config-file: '.github/workflows/markdown.links.config.json'
-          folder-path: 'docs/'
-          check-modified-files-only: 'yes'
-          file-extension: '.md'
-          base-branch: 'main'
```
HarshCasper, line 39: Can we possibly keep it around or replace it as a pre-commit hook?

.github/workflows/run-pre-commit.yaml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Install pre-commit
+        run: |
+          pip install pre-commit
+  
+      - name: Run pre-commit
+        run: pre-commit run --all
```
HarshCasper, line 23: Please add a newline at the EOF

HarshCasper, line 23: Is it possible to run individual `pre-commit` hooks at each step to make it easy to debug? Right now, I don't know where does a `pre-commit` fail, hence making the process of finding issues difficult.

HarshCasper (changes requested):

.github/workflows/run-pre-commit.yaml
viniciusdc: I don't mind the way it is now, but I see what you mean. So I don't mind adding extra steps as well, for a better layout.

.github/workflows/docs.yml
viniciusdc: I would suggest having this as a Cron job then, as they did in [here](https://github.com/GoogleChrome/lighthouse/blob/master/.github/workflows/cron-weekly.yml)

iameskild:
> One thing that I suggest is to add the following `if` condition to the `pre-commit` job: `if: github.event.pull_request.merged == false` this will skip the workflow to run everytime when a PR is merged (to avoid running twice)

I like this!

.github/workflows/docs.yml
iameskild: I can see a benefit of it being run as a cronjob. I'm not sure about the rest of the team, but I miss to run into issues with this particular link checker on a regular basis. Perhaps finding another one altogether would also be useful...

.github/workflows/run-pre-commit.yaml
iameskild: Thanks for the suggestion @HarshCasper. Having it all in one command will still output all of the failures.

.github/workflows/docs.yml
iameskild: Perhaps we will save the link checker discussion for another time.

iameskild:
@viniciusdc @HarshCasper @costrouc this PR is ready to be merged :)

HarshCasper (approved):

iameskild:
I need to understand why the Kubernetes tests are failing before merging...

iameskild:
I need to understand why the Kubernetes tests are failing before merging...

iameskild:
@viniciusdc some of the recent PRs had "unformatted" files which caused the pre-commit CI to fail. At this point, we will need to pull down the changes, make pre-commit happy and push those changes up. Is this workflow too complex? 

I will need to investigate why the jupyterlab image failed to build though.

qhub/template/stages/07-kubernetes-services/modules/kubernetes/services/jupyterhub/files/jupyterhub/01-theme.py
```diff
@@ -1,10 +1,9 @@
+import z2jh
```
viniciusdc, line 1: These files should NOT be modified -- the import order is important here, this might fix the Kubernetes tests, @iameskild add an exception to the python linting to avoid this dir.

iameskild: Awesome!! Thanks a lot for this find @viniciusdc! I'll go ahead and make these changes. Do you know why this ordering matters?

viniciusdc: I do know exactly the problem, as I only encountered it when importing the KubeSpwaner. But it seems that by changing the order some extra configurations are not imported correctly and because of this, some settings ends up with missing imports... something on these lines. I will try to find a more elaborate answer.

iameskild: Thanks @viniciusdc! I was just curious since I haven't come across an issue like this before.

PR #1300, opened 2022-05-24
ci: prevent image scans on main image builds

HarshCasper:

- Prevent image scans on main image builds


viniciusdc (approved):
LGTM, thanks Harsh

PR #1344, opened 2022-06-23
Remove the need for AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY to be set with Digital Ocean deployment

### LINKED ISSUE #1343: [ENH] - Remove the need for AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY to be set with Digital Ocean deployment
### Feature description

Since we already have the environment variables needed with SPACES_ACCESS_KEY_ID and SPACES_SECRET_ACCESS_KEY. This is really just a terraform-ism that we have been exposing to the users see https://www.terraform.io/language/settings/backends/s3 and the environment variables that are expected.

### Value and/or benefit

For the user experience it would be nice to have less environment variables required from the user.

### Anything else?

_No response_

costrouc:
Closes #1343

AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY will be automatically set based on SPACES_ACCESS_KEY_ID and SPACES_SECRET_ACCESS_KEY.

## Types of changes

- [ ] Documentation Update
- [X] Refactoring (no functional changes, no API changes)
- [ ] Build related changes

### Requires testing

- [X] Yes
- [ ] No


qhub/utils.py
```diff
@@ -345,6 +345,22 @@ def keycloak_provider_context(keycloak_credentials: Dict[str, str]):
...
 
 
+@contextlib.contextmanager
+def terraform_state_context(provider: str, terraform_state: str):
+    credentials = {}
+
+    if provider == "do" and terraform_state == "remote":
+        credentials.update(
+            {
+                "AWS_ACCESS_KEY_ID": os.eviron["SPACES_ACCESS_KEY_ID"],
```
viniciusdc, line 355: So, basically, we use `AWS` when we need to provide a remote backend for DO, is that right? This means that if the user wants to use `remote` for a DO deployment it also would need to have an AWS account and those variables in hand, right?

costrouc: Digital Ocean spaces is a compliant S3 object storage. The reason we have to do this is due to a limitation/assumption in the terraform remote state. We are not actually using anything related to AWS here but making it look to terraform remote state

costrouc:
@viniciusdc this will allow you to remove the need for setting those environment variables for the Digital Ocean deploymet.

qhub/utils.py
viniciusdc: > due to a limitation/assumption in the terraform remote state

viniciusdc:
@costrouc may I ask if you can do a DO `render` over this branch to validate if everything works? -- as the CI envs here possess all the necessary keys already, we might be unable to identify some exceptions

qhub/utils.py
costrouc: https://www.terraform.io/language/settings/backends/s3

viniciusdc:
I will be testing this on DO, today

qhub/utils.py
```diff
@@ -345,6 +328,22 @@ def keycloak_provider_context(keycloak_credentials: Dict[str, str]):
...
```
viniciusdc, line 338:
```suggestion
                "AWS_ACCESS_KEY_ID": os.environ["SPACES_ACCESS_KEY_ID"],
```

viniciusdc:
HI @costrouc sorry for the messy revert there, It seems that the destroy command fails with the missing environments -- I tried adding a quick workaround, but the error persisted so I reverted the PR.

viniciusdc:
to fix the destroy part we just need to add the new terraform state context manager in `gather_stage_outputs` and `destroy_stages` of `destroy.py`

PR #1352, opened 2022-06-29
Add Python-3.10

### LINKED ISSUE #1346: [MAINT] - Add Python `3.10` support
Python 3.10 is stable. This ticket is to collect the requirements around adding support:

- [ ] Add `3.10` to the test matrix
- [ ] Push to qhub-beta
- [ ] ...

HarshCasper:
Fixes #1346


peytondmurray:
Out of curiosity, why does `PYTHON_VERSION` need to be updated in `qhub/provider/cicd/{gitlab,github}.py`? I know @iameskild said it needed to be updated in the issue...

I think there are also a few places in the docs where 3.10 support needs to be mentioned:

* `setup.cfg`: PYPI package tags
* `docs/index.md`: Python version badge

Otherwise looks good to me!

iameskild (changes requested):
Thanks for working on this @HarshCasper! 

I probably should've done this to begin with but could we move the `PYTHON_VERSION` used in the `cicd` folder into [`constants.py`](https://github.com/Quansight/qhub/blob/main/qhub/constants.py)? Perhaps another name like `LATEST_SUPPORTED_PYTHON_VERSION` or something would be more informative as well.

Issue #1349, opened 2022-06-28
[Docs] Add docs regarding `filesystem` and `global` environments


viniciusdc:
The environment names for `filesystem` and `default` leads to some misconception about what these environments are used for and how they can be managed. We need to create some documentation around what they are, how they’re used, and what they mean.

trallard:
TODO:

- [ ] renaming of `filesystem` `default` 
- [ ] update docs as needed

viniciusdc:
We found a bug within conda-store that was blocking the proposed changes, this was fixed in https://github.com/Quansight/conda-store/pull/346. This issue is blocked until the latest version of conda-store is integrated on Nebari.

pavithraes:
Closing as a duplicate of https://github.com/nebari-dev/nebari-docs/issues/80

PR #1357, opened 2022-07-01
[ENH] Switch default and filesystem name envs

viniciusdc:

- Update `filesystem` and `default` environment names to `nebari-git` and `global` respectfully 
 
related to #1349 

## Types of changes

- [x] Breaking change (fix or feature that would cause existing features to not work as expected)
- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

## Documentation

- [ ] Yes, docstrings
- [ ] Yes, main documentation

This will require a full deployment of Qhub to validate all services run as excepted.


iameskild (dismissed):
This looks good to me! I know we discussed this as a team but I still have some reservations about `nebari-git` as a namespace name. However given that this was a team decision, I'm fine with it.

As I was thinking this through a few questions came to mind. 
- What happens when we upgrade from `0.4.2` to `0.4.3`, does the `filesystem` and `default` namespaces just linger around? 
  - My assumption is that they will stick around but might not be useable.
- And if so, does this cause issues for new environments?
- And do the users need to recreate their old envs that were previously under the `filesystem` namespace? 
  - I assume when they redeploy, the envs in the config file will be automatically recreated under the `nebari-git` namespace but if there were other envs created using the conda-store UI, I don't think those will be.

viniciusdc:
I think I am missing something 
image

viniciusdc:
I am out of ideas now... @costrouc may you have a look if I am missing something? Even with a fresh install, the old namespaces were created. 

I've also tried using 
```
c.CondaStore.default_namespace = "global"
c.CondaStore.filesystem_namespace = "nebari-git"
```

trallard:
@costrouc  this is waiting for your input ☝🏽

trallard:
Blocked by [#conda](https://github.com/Quansight/qhub/pull/1385)

iameskild:
@viniciusdc I opened this [PR in nebari-docker-images repo](https://github.com/nebari-dev/nebari-docker-images/pull/21) given that we recently decoupled the images. I will try to test this on quansight-beta 👍

viniciusdc:
> @viniciusdc I opened this [PR in nebari-docker-images repo](https://github.com/nebari-dev/nebari-docker-images/pull/21) given that we recently decoupled the images. I will try to test this as well, given the concerns I raised above +1

Thanks, @iameskild. I will then focus on updating this one. Also, I forgot to answer your points above, apologies, I do believe the old environments will still be in there, but some of the services might not show them. The users will definitely need to recreate the environments in the new namespaces as I don't think conda-store will treat those changes as a proper "renaming."

viniciusdc:
Ok... so this will need to be updated:
- `default` -> `global` -> `nebari-system`

@costrouc Is there a way to hide this environment from the user?

costrouc:
@viniciusdc yes we just drop permissions for users with this given namespace and they will no longer see it.

tests_deployment/test_jupyterhub_ssh.py
```diff
@@ -94,7 +94,10 @@ def test_exact_jupyterhub_ssh(paramiko_object):
         ("whoami", constants.KEYCLOAK_USERNAME),
         ("pwd", f"/home/{constants.KEYCLOAK_USERNAME}"),
         ("echo $HOME", f"/home/{constants.KEYCLOAK_USERNAME}"),
-        ("conda activate default && echo $CONDA_PREFIX", "/opt/conda/envs/default"),
```
costrouc, line 97: I don't think this should be changed.

qhub/template/stages/07-kubernetes-services/modules/kubernetes/services/jupyterhub/main.tf
```diff
@@ -50,7 +50,7 @@ resource "helm_release" "jupyterhub" {
               kind      = "configmap"
             }
 
-            "/opt/conda/envs/default/share/jupyter/lab/settings" = {
```
costrouc, line 53: I don't think this should be changed. This is the name of the conda env from the docker image.

costrouc (dismissed):

viniciusdc:
Thanks @HarshCasper I will have a look and test this as well!!! :smile:

viniciusdc:
@HarshCasper based on what we discussed today, we will need to make the `nebari-git` a customizable name in the config -- we can use the input variables and the ztjb custom config to pass through the values.

Regarding the name for `global,` we need to make sure this is just used in the installation part of jupyterhub/jupyterlab -- I am inclined to say that's the only case. Still, we can verify by checking wich env is used in the user pod terminal...

To hide an environment, we can change the permission part of global to admin only -- this should hide for the others...

iameskild (commented):
I tested this on an existing deployment and it was successful. I was able to create an env in the `global` namespace.

As a team, we should also decide how we want to handle this for existing users as the existing namespaces will be sticking around. Some possible options include:
- an `initContainer` that will move / clean the old namespaces - moderately challenging
- provide instructions on how to delete the old namespaces - simply but a bit hostile
  - this might also include rebuilding the envs that were once in the `filesystem` namespace

cc @viniciusdc

trallard:
TODO: add troubleshooting section for folks upgrading to `0.4.5` and reference section on envs

iameskild (approved):
Thanks a lot for seeing this through to the end @viniciusdc :)

Issue #1292, opened 2022-05-19
Nebari rename steps


costrouc:
# Step-by-step-migration plan

### Accounts and tooling

- [x] create quay.io and docker organization for nebari @aktech
- [x] Get access to PyPI nebari
- [x] create `documentation` repository and move docs to this directory
- [x] create `images` repository and move images to this directory and kickoff CI building images
    - [x] #1174

### Tasks to be done

- [x] #1445
- [x] #1446
- [ ] #1447
- [x] Create rename milestone 
  - [x] assign issues and tasks
- [x] #1395 
- [x] Open PR for CI/CD to work on both, `<release-branch>` and `main`
- [x] #1448
- [ ] #1449
- [x] #1451
- [x] #1452
- [x] #1453
- [x] #1454
- [x] Delete all stale and merged branches
- [x] #1444

### Migration day end-game

These tasks are numbered to denote order in which they should be performed:

1. Wrap up `qhub` repository
   - [x] #1455
   - [x] #1456

2. Transfer repo over to `nebari-dev`

- [ ] Transfer repository to `nebari-dev` GH organisation (should also rename to nebari)
  - [x] #1457
  - [x] #1458
  - [x] #1459
  - [ ] #1460
  - [ ] `quansight/qhub-deployment-*` do not move
- [x] #1461
- [x] #1462
- [x] #1463
- [x] #1464
- [x] #1465

After the first release:

- [x] #1466

trallard:
> * written as `Nebari` never lowercase?
Our organisation is `nebari-dev` so I assume the correct one is `nebari` lowercase - up for discussion

> * we don't use a short name always use `nebari`
correct 

> Need to be ready to pull the trigger:
Missing:
- Need to work with design to create the branding and related assets 

> All within 1 day:
> 
> * [ ]  move to `nebari-dev` github organization move qhub repo (non-destructive)
>   
>   * [ ]  `nebrari-dev/qhub`
Should be [`nebari-dev/nebari`](https://github.com/nebari-dev/nebari)

>   * [ ]  `quansight/qhub-jupyterhub-theme` -> `nebari-dev/nebari-jupyterhub-theme`
>   * [ ]  For qhub change the cirun to have a different repo quansight/qhub -> nebari-dev/nebari

>   * [ ]  `quansight/qhub-deployment-*` do not move
Is there a reason why we are not moving? Could we do a blue-green deployment if the concern is availability related?

iameskild:
[...]

PR #1371, opened 2022-07-15
Decouple docker images

iameskild:
Fixes | Closes | Resolves - related to #1292 

- Now that the docker images have been transferred to its [own repo](https://github.com/nebari-dev/nebari-docker-images), we can remove the references here. As a result, a lot of code is being removed. 
- With the images transferred, there are few minor changes that were needed in order to properly set the tags for the jupyterlab, jupyterhub and dask-worker images.
    - for testing (`kubernetes.yaml`), we will pull the latest image tagged from the `main` branch of `nebari-docker-images` to ensure that everything from both repos works together. I believe this preserves the spirit of how the tags were set prior to this change.
    - for `qhub init`, a small function (`set_qhub_image_tag`) will download a list of the GitHub tags for the `nebari-docker-images` repo and use the latest (no dev or pre-releases) tag.
        -  A similar function has been created to set the `qhub_dask_version` as well.
    - in both cases, these image tags can be overridden by setting the `QHUB_IMAGE_TAG` environment variable.
        - A similar env var (`QHUB_DASK_VERSION`) has been introduced to override the `qhub_dask_version`.  
  
 
## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Documentation

- [ ] Yes, docstrings
- [ ] Yes, main documentation


.github/workflows/kubernetes_test.yaml
```diff
@@ -32,9 +32,9 @@ jobs:
       run:
         shell: bash -l {0}
     steps:
-      - name: 'QHUB_GH_BRANCH set for PR'
```
trallard, line 35: Since this is tied - we might need better tags for our Docker images.
- Up to now they have been coupled with the code so they have been following  semver
- Using the branch as part of a tag is good practice
- I would suggest having a combination of `branch-shortsha-date` to ensure reproducibility and easier tracking plus we will eventually move to CalVer for nebari so we can remove the need for a SemVer tag

qhub/cli/__init__.py
```diff
@@ -51,6 +51,9 @@ def cli(args):
     if QHUB_GH_BRANCH:
         logging.info(f"Modifying for development branch {QHUB_GH_BRANCH}")
 
+    if QHUB_IMAGE_TAG:
+        logging.info(f"Modifying image tag, using `{QHUB_IMAGE_TAG}` instead.")
```
trallard, line 55: Modifying what to image, tag? Unless we are explicitly setting some config then overwriting this logging message and the above one seem particularly cryptic.

qhub/provider/cicd/common.py
```diff
@@ -0,0 +1,13 @@
+import os
+
+
+def pip_install_qhub(qhub_version: str) -> str:
+    qhub_gh_branch = os.environ.get("QHUB_GH_BRANCH")
```
trallard, line 5: what happens if there is no ENV for whatever reason? need to provide a fallback or at least raise an error here

qhub/provider/cicd/github.py
```diff
@@ -1,22 +1,28 @@
...
-            raise ValueError(
-                f"environment variable={name} is required for github automation"
-            )
+def github_request(url, method="GET", json=None, authenticate=True):
+    auth = None
+    if authenticate:
+        for name in ("GITHUB_USERNAME", "GITHUB_TOKEN"):
+            if os.environ.get(name) is None:
+                raise ValueError(
+                    f"environment variable={name} is required for github automation"
```
trallard, line 21:
```suggestion
                    f"Environment variable={name} is required for GitHub automation"
```

qhub/utils.py
```diff
@@ -388,13 +387,33 @@ def deep_merge(*args):
...
-    # dev branches
-    if len(qhub_version.split(".")) > 3 and qhub_gh_branch:
-        pip_install = (
-            f"pip install git+https://github.com/Quansight/qhub.git@{qhub_gh_branch}"
+def set_docker_image_tag() -> str:
+    """Set docker image tag for `jupyterlab`, `jupyterhub`, and `dask-worker`."""
+    try:
+        qhub_image_tag = get_latest_repo_tag(DOCKER_IMAGE_OWNER, DOCKER_IMAGE_REPO)
+    except ConnectionError:
+        print("Unable to connect to the GitHub API, falling back to the default value.")
```
trallard, line 395:
```suggestion
        print(f"Unable to connect to the GitHub API, falling back to the default value: {DDEFAULT_QHUB_IMAGE_TAG}.")
```

For those that do not know what the default is

qhub/utils.py
```diff
@@ -388,13 +387,33 @@ def deep_merge(*args):
...
+
+def set_qhub_dask_version() -> str:
+    """Set version of `qhub-dask` meta package."""
+    try:
+        resp = requests.get(CONDA_FORGE_CHANNEL_DATA_URL)
+        qhub_dask_version = resp.json()["packages"]["qhub-dask"]["version"]
+        resp.raise_for_status()
+    except ConnectionError:
+        print(
+            "Unable to connect to the Conda-Forge channel data, falling back to the default value."
```
trallard, line 412: Same as above - what is default value?

trallard (changes requested):

trallard:
Added a few comments, nothing super major @iameskild, thanks for sorting this out

qhub/provider/cicd/common.py
iameskild: If `os.environ.get` can't find the specified env var, it will default to None. That said, I will make this explicit :)

.github/workflows/kubernetes_test.yaml
iameskild: Sounds good to me! As discussed in yesterday's meeting, we will use `date-branch-shortSha`. This will need to be updated in the nebari-docker-images repo.

qhub/cli/__init__.py
iameskild: I updated these messages to be a bit more informative and moved them to more appropriate places 👍

iameskild:
@trallard I believe this ready for a final review. Once this PR is merged, all future updates to the images will need to be made in the nebari-docker-images repo 👍

costrouc (approved):
Looks like a good solution moving forward for updating the nebari image tags. Happy to see that we are decoupling the docker images from Nebari!

PR #1383, opened 2022-07-28
add code server version to fix build

HarshCasper:
https://github.com/Quansight/qhub/issues/1380

### DIFF
--- a/qhub/template/image/scripts/install-code-server.sh
+++ b/qhub/template/image/scripts/install-code-server.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 set -xe
 DEFAULT_PREFIX="${1}"; shift # path to environment yaml or lock file
+CODE_SERVER_VERSION=4.5.0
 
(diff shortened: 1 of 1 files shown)


qhub/template/image/scripts/install-code-server.sh
```diff
@@ -15,7 +15,7 @@ if [[ ! $(sha256sum install.sh) == "${expected_sum}  install.sh" ]];then
 fi
 
 mkdir /opt/tmpdir
-sh ./install.sh --method standalone --prefix /opt/tmpdir
+sh ./install.sh --method standalone --prefix /opt/tmpdir --version 4.5.0
```
aktech, line 18: Thanks for working on this, this is written at two places, here and in line 20, please define a variable and use that at both places.

HarshCasper: Fixed it now.

aktech (approved):

PR #1386, opened 2022-07-29
CI: Merge test and release to PyPi workflows into one

HarshCasper:
Fixes https://github.com/Quansight/qhub/issues/1331


.github/workflows/release.yaml
```diff
@@ -1,13 +1,58 @@
...
+name: Test & Publish PyPi release
 
 on:
   release:
     types: [created]
+  workflow_dispatch:
+    inputs:
+      qhub_tag:
+        description: 'Tag for test PyPI release'
+        required: True
```
iameskild, line 10:
```suggestion
```
We were using `workflow_dispatch` in the past so that we could manually trigger the test-release workflow, since we have them both in the same workflow, I don't think we need this any more.

.github/workflows/release.yaml
```diff
@@ -1,13 +1,58 @@
...
-  pypi:
+  test-pypi:
+    name: Test PyPi release
     runs-on: ubuntu-latest
 
+    steps:
+      - name: Set up python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
```
iameskild, line 21:
```suggestion
          python-version: 3.10
```
Let's bump this :)

iameskild: Could you also please bump the `python-version` for the second job on line 60?

iameskild (changes requested):
Thanks @HarshCasper for working on this! I requested a few minor changes but otherwise this is looking good 💯

iameskild (approved):

iameskild (approved):

PR #1387, opened 2022-08-02
Set LATEST_SUPPORTED_PYTHON_VERSION as str

iameskild:
When `LATEST_SUPPORTED_PYTHON_VERSION` is converted from float to str, the trailing 0 is dropped, which results in `"3.1"`. This PR just converts the `3.10` into a string to begin with avoiding this issue.


viniciusdc (approved):
Nice catch!!

PR #1393, opened 2022-08-12
[ENH] Add `pyarrow` and `s3fs`

viniciusdc:
Related to https://github.com/Quansight/qhub/issues/1359

- Add two new packages to Dask env, `pyarrow` and `s3fs`

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Documentation

- [ ] Yes, docstrings
- [ ] Yes, main documentation

### DIFF
--- a/qhub/initialize.py
+++ b/qhub/initialize.py
@@ -243,7 +243,7 @@ DEFAULT_ENVIRONMENTS = {
             "python-graphviz",
             "pyarrow",
             "s3fs",
-            "gcsf",
+            "gcsfs",
             "numpy",
             "numba",
             "pandas",
--- a/qhub/initialize.py
+++ b/qhub/initialize.py
@@ -243,6 +243,7 @@ DEFAULT_ENVIRONMENTS = {
             "python-graphviz",
             "pyarrow",
             "s3fs",
+            "gcsf",
             "numpy",
             "numba",
             "pandas",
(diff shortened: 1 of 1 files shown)


iameskild (approved):
Looks good to me :)

I wonder if it's also worth adding `gcsfs`, what do you think?

viniciusdc:
> Looks good to me :)
> 
> I wonder if it's also worth adding `gcsfs`, what do you think?

If we add `s3sf` I don't see a problem, though to be honest I would prefer to keep `s3fs` and `gcsfs` as something the user should add to a personal env based on `filesystem/dask`. The reason is, that dask should be the basic env for running dask operations, anything outside this should go to a user/group-based env.

qhub/initialize.py
```diff
@@ -241,6 +241,8 @@
             "ipywidgets",
             f"qhub-dask =={qhub_dask_version}",
             "python-graphviz",
+            "pyarrow",
+            "s3fs",
```
viniciusdc, line 245:
```suggestion
            "s3fs",
            "gcsf",
```

qhub/initialize.py
```diff
@@ -241,6 +241,9 @@
...
+            "gcsf",
```
viniciusdc, line 246:
```suggestion
            "gcsfs",
```

iameskild (approved):

PR #1410, opened 2022-08-19
ci: add support for develop branch

HarshCasper:


.github/workflows/kubernetes_test.yaml
```diff
@@ -32,9 +33,9 @@ jobs:
       run:
         shell: bash -l {0}
     steps:
-      - name: "Set QHUB_IMAGE_TAG=main"
+      - name: "Set QHUB_IMAGE_TAG=develop"
```
trallard, line 36: This hardcodes for `develop` and drops `main` altogether - the idea is that it works for both not one or the other. 
Could leverage GH actions context to get the branch/ref name and set here

.github/workflows/kubernetes_test.yaml
```diff
@@ -32,9 +33,9 @@ jobs:
...
         run: |
-          echo "QHUB_IMAGE_TAG=main" >> "$GITHUB_ENV"
+          echo "QHUB_IMAGE_TAG=develop" >> "$GITHUB_ENV"
```
trallard, line 38: same as above

trallard (changes requested):

trallard:
After going over the release workflow - and migration tasks - it does seem we need the workflow to run against `main` AND release branches, so we'd need something like:


`branches: [ main, 'release/v[0-9].[0-9]+' ]` cc @costrouc  ofc need to agree on the `release` prefix - but I think that should be a reasonable branching strategy WDYT?

.github/workflows/run-pre-commit.yaml
```diff
@@ -2,7 +2,7 @@ name: Run pre-commit
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, develop ]
```
trallard, line 5: we no longer have a develop branch - there is only a release (with the corresponding CalVer patter) and a main branch

HarshCasper:
Hi @trallard — Can you please have a look at it once again?

PR #1425, opened 2022-09-02
Fixing provider=existing for local/existing

costrouc:
This fixes and issue when using provider=existing in 0.4.4rc4. This is broken before the release. 

Can reproduce when using `qhub init existing ...`.

### DIFF
--- a/qhub/deploy.py
+++ b/qhub/deploy.py
@@ -195,7 +195,10 @@ def guided_install(
     check_cloud_credentials(config)
 
     stage_outputs = {}
-    if config["provider"] != "local" and config["terraform_state"]["type"] == "remote":
+    if (
+        config["provider"] not in {"existing", "local"}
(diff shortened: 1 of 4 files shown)


qhub/deploy.py
```diff
@@ -195,7 +195,10 @@ def guided_install(
     check_cloud_credentials(config)
 
     stage_outputs = {}
-    if config["provider"] != "local" and config["terraform_state"]["type"] == "remote":
+    if (
+        config["provider"] not in {"existing", "local"}
```
iameskild, line 199: @costrouc I added this check on the `provider` as well.

iameskild (approved):
Nice catch @costrouc! I made one small commit but otherwise, this looks good to me :)

costrouc:
Thanks for catching that @iameskild !

PR #1428, opened 2022-09-08
Add option for AWS node-groups to run in a single subnet/AZ

### LINKED ISSUE #1388: [ENH] - Add ability to spinup dask workers in a single AZ (AWS)
### Feature description

Ability to spin up dask workers in a single availability Zone in AWS.

### Value and/or benefit

While running data intensive tasks via dask workers, it happens quite often that dask workers are spun up in various AZs (Availability zones), which can cause lot of data transfer across AZs, which is not very cheap.

Having this ability will make spinning up large number of dask workers very cost efficient.

### Anything else?

_No response_

iameskild:
Fixes | Closes | Resolves #1388 

- This adds a node-group option for it to be created using a single subnet (ie single AZ). At the moment, this will be the default behavior for the `worker` node-group but there might be interest in having this be the default for all node-groups.
  - Note: this only affects AWS. 

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [x] Yes
- [ ] No

I have manually tested this but more testing might be need to ensure that we are not breaking other features.

- [ ] Yes
- [ ] No

## Documentation

- [ ] Yes, docstrings
- [ ] Yes, main documentation


brl0:
@iameskild, just a quick heads-up that this change does have the effect of limiting available IP addresses, which limits the number of concurrent workers.

iameskild:
Thanks for the feedback @brl0! This was a concern that I had and I'm glad we were able to confirm one way or the other.

iameskild:
@brl0 we recently added an AWS EKS addon ([`aws-ebs-csi-driver`](https://github.com/Quansight/qhub/blob/c08ee7164693ddd05f7956a932f7628bfee98ebc/qhub/template/stages/02-infrastructure/aws/modules/kubernetes/main.tf#L53-L63)) and I believe there is an addon that will allow us to reduce the number of IP addresses assigned to each of the pods (if I remember correctly its currently ~30 per pod). 

I recall researching this before and indeed I did haha - https://github.com/Quansight/qhub/issues/828#issuecomment-935116226 
That said, I will likely read through this a bit more but this is promising :)

Have a look here for more info: https://github.com/aws/amazon-vpc-cni-k8s

brl0:
@iameskild, I think this same configuration needs to also be applied to the Node Group for user instances as well, since there is significant data transfer between user instances and workers when computing large dataframes etc, because if this change is not applied to user instances additional fees could be incurred for transfers between those them and the workers.  

Also, just curious, but do you know if the subnet mask size can be increased to avoid unnecessarily low worker count constraint this change otherwise implies.

iameskild:
@costrouc @viniciusdc this is the last PR that needs to be included in the January release 🎉

costrouc (approved):
No issues with this PR. Will be great to finally run under one subnet as an option. This will remove a lot of the ways in which aws kubernetes behaves weird.

PR #1429, opened 2022-09-08
Add `--disable-checks` flag to deploy

iameskild:
Fixes | Closes | Resolves #

-Under special circumstances, users might need to disable the checks that take place at the end of each stage. This new flags allows for that possibility.  

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Documentation

- [ ] Yes, docstrings
- [ ] Yes, main documentation

### DIFF
--- a/qhub/deploy.py
+++ b/qhub/deploy.py
@@ -97,7 +97,7 @@ def provision_ingress_dns(
     dns_provider: str,
     dns_auto_provision: bool,
     disable_prompt: bool = True,
-    check: bool = True,
+    check: bool = False,
 ):
     directory = "stages/04-kubernetes-ingress"
 
@@ -189,6 +189,7 @@ def guided_install(
     dns_provider,
     dns_auto_provision,
     disable_prompt=False,
+    disable_checks=False,
     skip_remote_state_provision=False,
 ):
     # 01 Check Environment Variables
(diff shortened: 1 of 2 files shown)


qhub/deploy.py
```diff
@@ -156,31 +158,31 @@ def provision_06_kubernetes_keycloak_configuration(stage_outputs, config, check=
...
-def provision_08_qhub_tf_extensions(stage_outputs, config, check=True):
+def provision_08_qhub_tf_extensions(stage_outputs, config, disable_checks=False):
     directory = "stages/08-qhub-tf-extensions"
 
     stage_outputs[directory] = terraform.deploy(
         directory=directory,
         input_vars=input_vars.stage_08_qhub_tf_extensions(stage_outputs, config),
     )
 
-    if check:
```
costrouc, line 183: I often find double negatives harder to reason about but don't have a strong opinion that it must be this way. What do you think?

```
provision_02_infrastructure(stage_outputs, config, disable_checks)

def provision_02_infrastructure(stage_outputs, config, disable_checks=False):
      if not disable_checks:
          ...
```

vs. 

```python
provision_02_infrastructure(stage_outputs, config, (not disable_checks))

def provision_02_infrastructure(stage_outputs, config, check=True):
      if check:
          ...
```

iameskild: Yeah I understand where you're coming from. In this case, I feel like it reads in *fairly* comprehensible way but I'm open to other options as well. I do find this pattern a little pattern a little confusing myself `provision_02_infrastructure(stage_outputs, config, (not disable_checks))`. 

Perhaps we could do something like this instead:

```
    if disable_checks:
        return 
    checks.stage_02_infrastructure(stage_outputs, config
```

costrouc (approved):

PR #1431, opened 2022-09-09
Adding option to supply additional arguments to ingress via `ingress.terraform_overrides.additional-arguments`

costrouc:
ingress.terraform_overrides.additional-arguments: [...]

See https://doc.traefik.io/traefik/reference/static-configuration/cli/

### DIFF
--- a/qhub/template/stages/04-kubernetes-ingress/main.tf
+++ b/qhub/template/stages/04-kubernetes-ingress/main.tf
@@ -11,4 +11,5 @@ module "kubernetes-ingress" {
   certificate-secret-name   = var.certificate-secret-name
   load-balancer-annotations = var.load-balancer-annotations
   load-balancer-ip          = var.load-balancer-ip
+  additional-arguments      = var.additional-arguments
 }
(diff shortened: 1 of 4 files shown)


costrouc:
For tomorrow I think the cli option we are looking to use is `--entryPoints.websecure.forwardedHeaders.insecure`

iameskild (approved):
To confirm, this [section here](https://github.com/Quansight/qhub/blob/c5dc6ddc85e97b4f02ce07e716043575dfbc04d7/qhub/stages/input_vars.py#L195) will unpack the `ingress.terraform_overrides` and so long as the values for `additional-arguments` (<-- typed with a `-` and not `_`) are passed as a list, this should just work as is. 

The above was mostly for me but if I've misunderstood, let me know. Otherwise, this looks great :)

costrouc:
@iameskild that is right. Longer term I think it makes sense to centralize these options so that terraform_overrides can be supplied to any stage.

Yes an example would be 

```yaml
ingress:
  terraform_overrides:
    additional-arguments:
      - "--entryPoints.websecure.forwardedHeaders.insecure"
```

PR #1519, opened 2022-10-27
Remove old CLI + cleanup

iameskild:
Fixes | Closes | Resolves #

> Please remove anything marked as optional that you don't need to fill in.
> Choose one of the keywords preceding to refer to the issue this PR solves, followed by the issue number (e.g Fixes # 666).
> If there is no issue, remove the line. Remove this note after reading.

## Changes introduced in this PR:

- Removes the old CLI from the codebase
- Update reference to logo in theme

## Types of changes

What types of changes does your PR introduce?

_Put an `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds a feature)
- [ ] Breaking change (fix or feature that would cause existing features to not work as expected)
- [ ] Documentation Update
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Other (please describe):

### Requires testing

- [ ] Yes
- [ ] No

### In case you checked yes, did you write tests?

- [ ] Yes
- [ ] No

## Documentation

Does your contribution include breaking changes or deprecations?
If so have you updated the documentation?

- [ ] Yes, docstrings
- [ ] Yes, main documentation
- [ ] Yes, deprecation notices

## Further comments (optional)

If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered and more.

PR #1521, opened 2022-10-28
Update `skip_remote_state_provision` default value

viniciusdc:
Fixes | Closes | Resolves #

- Update  `skip_remote_state_provision` default value from `True` to `False`

By keeping the `skip_remote_state_provision` default value as `True,` we were inadvertently revoking the creation of the terraform state bucket during the first deployments. Which is used during the stage provision phase:
https://github.com/nebari-dev/nebari/blob/77d15e7a16e87bfef234bdf4802bf622f6030618/nebari/deploy.py#L201-L208

I think the reason we didn't spot this sooner as due to until #1519 we still used a mix of the new CLI with the old `qhub` entry points.

Image of the error caused during deployment on GCP with `terraform_state: remote` (default)  set in the config
```
nebari deploy -c nebari-config.yaml --dns-provider cloudflare --dns-auto-provision --disable-prompt
```

image

Even with no mention of `--skip-remote-state-provision` you can see the log message right at the beginning finalizing the skip.

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

### Requires testing

- [ ] Yes
- [ ] No

- [ ] Yes
- [ ] No

## Documentation

- [ ] Yes, docstrings
- [ ] Yes, main documentation

### DIFF
--- a/nebari/cli/main.py
+++ b/nebari/cli/main.py
@@ -289,7 +289,7 @@ def deploy(
         help="Disable the checks performed after each stage",
     ),
     skip_remote_state_provision: bool = typer.Option(
-        True,
+        False,
         "--skip-remote-state-provision",
         help="Skip terraform state deployment which is often required in CI once the terraform remote state bootstrapping phase is complete",
     ),


viniciusdc:
There seems to be a problem with pytest running the jupyterhub ssh test. I will handle that in another PR. I ran IT with this branch all working.

PR #1580, opened 2022-12-01
MAINT - Miscellaneous maintenance tasks

trallard:
- :fire: Remove RFD template
- :wrench: Update pre-commit config
- :arrow_up: Upgrade actions versions

Multiple maintenance updates needed 

Note that since I updated the pre-commit config I ran it through all the files so it would be best to review the changes at [`932cd18` (#1580)](https://github.com/nebari-dev/nebari/pull/1580/commits/932cd18e887c0c6f3ade4ae74d68b97f0a4eeb9c) to prevent having too many changed files noise


iameskild (approved):
Thanks for the clean up @trallard! I'm not entirely sure why all of the doc files were touched but either way it doesn't matter since they're being deprecated soon anyway 🪴

trallard:
I am not sure 🤷🏽 I only updated the pre-commit hooks so it makes me believe the previous versions were not catching some issues or maybe since the pre-commits are run on CI the last time those md files were  worked on they were not formatted? who knows?

PR #1595, opened 2022-12-08
Make conda-store file system read-only by default

alimanfoo:
This a small suggestion, but I'd suggest making the conda-store file system read-only by default. This will prevent users accidentally modifying environments created via conda-store by running pip install or conda install from their server.


pavithraes:
@alimanfoo Thanks for opening this PR!

iameskild (approved):
I've tested this on an existing deployment and I didn't run into any issues. I created a few new conda-store envs and had no trouble using them.

@costrouc do you have any objection to merging this?

costrouc (approved):
@iameskild look great to me and ready for merge!

PR #1628, opened 2023-01-28
Unpin packages in default dashboard env

### LINKED ISSUE #1617: [BUG] - Shipped dashboard env doesn't work with streamlit
### Describe the bug

We pinned back the dashboard env that we ship with because originally the conda solve was holding back nebari deployment completion. This is no longer the case with the current conda store integration. 

The streamlit pin (and lack of numpy pin) creates an env which works for the other dashboards, but when you try to run streamlit, it attempts to use a deprecated `numpy.float` function. 

We just need to update the default dashboard env to the latest streamlit (by removing all the pins) and everything should happily work again. 

### Expected behavior

I expect the default dashboard environment that nebari ships with to work with the streamlit example. 

Side note - I'm updating/removing the examples, but this issue will still be a case when that happens. 

NA

Attempt to deploy the streamlit example

### Command output

_No response_

_No response_

### Compute environment

None

### Integrations

_No response_

### Anything else?

_No response_

iameskild:
## Reference Issues or PRs

Fixes #1617 

This simply unpins the packages used by the default `nebari-git/dashboard` environment. It was originally pinned because the solve was taking a long time but with the switch to using mamba to build these environments, that hasn't been an issue. Our other default environment `nebari-git/dask` already has it's packages unpinned.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?


kcpevey (approved):
I dont think these should be left unpinned. That said, Pam and I have been working on updating the dashboard examples which is also requiring an update to this environment. So, I'm approving this to get this bug fixed, but I'll follow up with a PR which will make improvements on it (not to be included in this release).

Issue #1495, opened 2022-10-14
[ENH] - Submit workflows to Argo-Workflow from JupyterLab server


iameskild:
### Feature description

At the moment, there is no easy way to submit workflows to Argo-Workflow other than using their UI. It would be a better user experience if we could submit workflows from within the JupyterLab pod. 

### Value and/or benefit

This allows users to submit their workflows where they work. 

### Anything else?

See https://github.com/Quansight/qhub/issues/1230 for more details.

kcpevey:
Hera can be used to submit workflows from the jupyterlab server (via python instead of the yaml UI). I've got some examples and I'm working on adding them to nebari documentation because the process is NOT straightforward.

dcmcand:
We can use the scheduler functionality to run notebooks using argo-workflows currently. If there are other use cases needed, please open more specific tickets.

PR #1639, opened 2023-02-01
Update Argo Workflows to latest version

Adam-D-Lewis:
## Reference Issues or PRs

Related to https://github.com/nebari-dev/nebari/issues/1495, but I wouldn't say this closes it necessarily. 

Enables basic functionality of previously broken argo integration.

The Argo Workflows integration did not allow end users to submit workflows.  I looked into this a bit, and found at the moment, you can attempt to submit workflows in 2 ways:  
    1) From the UI 
    2) using the argo CLI

Neither way was working for the example argo workflow.  This was because the workflow needed some permissions which were not granted by default.  This comment (https://github.com/argoproj/argo-workflows/issues/2522#issuecomment-604590774) gives a decent review of the 3 relevant service accounts at play.  We haven't adequately configured the 3rd one listed, the container service account used by the actual pods spun up as part of the workflow.  By default, the pods use the default service account for the namespace where the workflow is deployed.  That default service account has no permissions by default.  

I solved this issue by setting the helm chart configuration to create a service account that workflows can use. I then set the workflow controller config to use that service account by default for all workflows (the service account used can still be overridden).  I also upgraded the helm chart, and added some additional user guide documentation on how to submit a workflow from the Argo CLI.

- [x] Documentation Update
- [ ] Build related changes

## Testing

- [x] Did you test the pull request locally?  I tested by running a local kind cluster and submitting workflows as explained in the docs I added.
- [ ] Did you add new tests?


docs/source/user_guide/argo_workflows.md
```diff
@@ -0,0 +1,32 @@
...
+
+You can submit a workflow by clicking "SUBMIT NEW WORKFLOW" on the landing page assuming you have the appropriate
+permissions.
+
+![Argo Server Landing Page](../images/argo-server-landing-page.png)
+
+## Submitting a workflow via Argo CLI
+
+You can submit or manage workflows via the Argo CLI. The Argo CLI can be downloaded from the
+[Argo Releases](https://github.com/argoproj/argo-workflows/releases) page. After downloading the CLI, you can get your
```
Adam-D-Lewis, line 21: This can be done outside of Nebari.  I'm not sure if we want to provide some better solution to install the CLI tool inside Nebari.  It's not available via conda that I could find.

nebari/template/stages/07-kubernetes-services/modules/kubernetes/services/argo-workflows/main.tf
```diff
@@ -62,9 +72,6 @@ resource "helm_release" "argo-workflows" {
           "${var.node-group.key}" = var.node-group.value
         }
       }
-      controller = {
```
costrouc, line 65: This was recently added by @iameskild. Do you know why this is no longer necessary?

costrouc (commented):
I haven't tested out this PR but I'll like @iameskild to review this PR so that we can review this more. I think that Argo is critical to Nebari and we need to ensure the user experience is great with this component. kbatch looks like it is not actively being managed and this will likely be our long term workflow solution.

costrouc:
@iameskild can you review/work with @Adam-D-Lewis on this PR?

nebari/template/stages/07-kubernetes-services/modules/kubernetes/services/argo-workflows/main.tf
Adam-D-Lewis: The emissary Workflow Executor is the default since version 3.3 of Argo Workflows.  The upgraded helm chart deploys v3.4.4 of Argo Workflows.

docs/source/admin_guide/argo-workflows.md
```diff
@@ -1,26 +1,18 @@
 # Argo Workflows
```
kcpevey, line 1: These docs are deprecated. The new documentation repo is here. 

The Argo docs haven't been filled out yet - [this issue](https://github.com/nebari-dev/nebari-docs/issues/49) and [this issue ](https://github.com/nebari-dev/nebari-docs/issues/48). I can move the changes you made over to the new documentation if you'd like.

Adam-D-Lewis: That would be great, thanks

docs/source/admin_guide/argo-workflows.md
```diff
@@ -35,10 +27,12 @@ argo_workflows:
 
 ## Disabling Argo Workflows
 
-To turn off the cluster monitoring on QHub deployments, simply turn off the feature flag within your `qhub-config.yaml`
-file. For example:
+To turn off the cluster monitoring on Nebari deployments, simply turn off the feature flag within your
+`nebari-config.yaml` file. For example:
```
Adam-D-Lewis, line 31: This should say "To turn off the argo workflows"

kcpevey:
I moved the documentation updates to [here](https://github.com/nebari-dev/nebari-docs/pull/268)

viniciusdc:
@iameskild do you have an idea why the Kubernetes tests are not running for this?

viniciusdc (approved):
Looks excellent, tested on Local/AWS instance. We need to move the docs to the new one and also we need a bigger task example to run.

kcpevey:
@ericdatakelly may have a better example to run

docs/source/user_guide/argo_workflows.md
Adam-D-Lewis: Correction: Users can download the binary from the releases page and use it inside of Nebari as well

ericdatakelly:
I think we need several examples to cover different use cases.  This looks like a good start, so how about if we make some new issues for other use cases?  I'm working on the case in which a user wants to run an ML pipeline with multiple tasks that depend on output/artifacts from previous steps.  I was running it from the ARGO UI.  Now that Hera Workflows are working, I can move to that implementation (in python code), which is a better fit for the use case.

iameskild (commented):
Thanks for making these update @Adam-D-Lewis! 

Although it looks like the Kubernetes test didn't run, I can assume folks have this deployed and working on a fresh install. I instead tried deploying these changes to an existing cluster and unfortunately ran into some problems when trying to update the Argo-Workflow CRDs. 

```
[terraform]: │ Error: rendered manifests contain a resource that already exists. Unable to continue with update: CustomResourceDefinition "clusterworkflowtemplates.argoproj.io" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "argo-workflows"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "dev"
[terraform]: │ 
[terraform]: │   with module.argo-workflows[0].helm_release.argo-workflows,
[terraform]: │   on modules/kubernetes/services/argo-workflows/main.tf line 6, in resource "helm_release" "argo-workflows":
[terraform]: │    6: resource "helm_release" "argo-workflows" {
[terraform]: │ 
[terraform]: ╵
```


At first I thought adding `force_update = true` would do the trick but upon further investigation, this looks like it is an upstream issue with Helm and updating CRDs in general.
- Relevant issue in helm terraform-provider: https://github.com/hashicorp/terraform-provider-helm/issues/944
- Relevant issue in helm: https://github.com/helm/helm/issues/11321

At this point, if we need to get around this we might want to "render the CRDs via the helm_template data-source and apply them using kubernetes_manifest..." ([quote source](https://github.com/hashicorp/terraform-provider-helm/issues/944#issuecomment-1242301169)).

Adam-D-Lewis:
@iameskild So what would you say are next steps needed on this PR?

iameskild:
Hey @Adam-D-Lewis I see two potential workarounds:
1. Add the Argo-Workflow CRDs as `kubernetes_manfifest` ([like we do for traefik](https://github.com/nebari-dev/nebari/blob/develop/nebari/template/stages/03-kubernetes-initialize/modules/traefik_crds/main.tf))
  - this is likely quite a tedious process
2. Try adding another TF module that handles creating/managing the CRDs
  - [this blog post](https://rpadovani.com/terraform-helm-crds-manager) gives me hope that this is possible
  
Perhaps @costrouc has some other ideas for how to get around this.

Adam-D-Lewis:
I see that helm uses an annotation `keep: true` to persist resources in the event of an uninstall.  I wonder if setting that to false by default would fix the problem since I see it's set to true in the crds in the argo workflows helm chart by default.  



Is there a good way to test that quickly, Eskild?  I saw you tried updating an existing deployment to test this PR.  Is the best way to do that to deploy from develop branch then deploy them update from this PR branch or is there a more automated way to test updates?

Adam-D-Lewis:
I see that helm uses an annotation `keep: true` to persist resources in the event of an uninstall.  I wonder if setting that to false by default would fix the problem since I see it's set to true in the crds in the argo workflows helm chart by default.  

Is there a good way to test that quickly, Eskild?  I saw you tried updating an existing deployment to test this PR.  Is the best way to do that to deploy from develop branch then deploy them update from this PR branch or is there a more automated way to test updates?

Adam-D-Lewis:
@iameskild I'll work on adding the Argo-Workflow CRDs as kubernetes_manifest since it seems like the most established solution.

Adam-D-Lewis:
Okay, I added the CRDs, but haven't tested deploying it yet.  So existing users of nebari will need to run `kubectl delete -f <existing CRD>` for each of the existing CRDs, I believe.  Additionally, if we need to update the CRDs in the future (e.g. we are updating Argo Workflows version), we would modify them here manually and then it should update without any manual steps once users deploy next, is that right?

iameskild:
@Adam-D-Lewis this is looking good to me! Ultimately, the best solution is one that doesn't require the user to run any other commands besides `nebari deploy ...`, I'm just not entirely sure how we get there. If we don't include Argo-Workflow CRDs from their Helm chart (if that's possible), I wonder if they will simply get deleted. Maybe something worth testing. 

As for updates to Argo-Workflows in the future, this approach would indeed require us to modify these CRDs. As far as I can tell, CRD updates are fairly irregular; I've only had to make one minor update to the Traefik CRD in the past year +.

Adam-D-Lewis:
Yeah, I disabled the CRDs being deployed via helm so this might work as we hope.  I'll give it a try.

Adam-D-Lewis:
Okay, I tried redeploying these changes to an existing nebari cluster and am getting the following error 
```
[terraform]: │ Error: Cannot create resource that already exists
[terraform]: │ 
[terraform]: │   with module.argo-workflows[0].kubernetes_manifest.customresourcedefinition_clusterworkflowtemplates_argoproj_io,
[terraform]: │   on modules/kubernetes/services/argo-workflows/main.tf line 254, in resource "kubernetes_manifest" "customresourcedefinition_clusterworkflowtemplates_argoproj_io":
[terraform]: │  254: resource "kubernetes_manifest" "customresourcedefinition_clusterworkflowtemplates_argoproj_io" {
[terraform]: │ 
[terraform]: │ resource "/clusterworkflowtemplates.argoproj.io" already exists
[terraform]: ╵
[terraform]: ╷
[terraform]: │ Error: Cannot create resource that already exists
[terraform]: │ 
[terraform]: │   with module.argo-workflows[0].kubernetes_manifest.customresourcedefinition_cronworkflows_argoproj_io,
[terraform]: │   on modules/kubernetes/services/argo-workflows/main.tf line 313, in resource "kubernetes_manifest" "customresourcedefinition_cronworkflows_argoproj_io":
[terraform]: │  313: resource "kubernetes_manifest" "customresourcedefinition_cronworkflows_argoproj_io" {
[terraform]: │ 
[terraform]: │ resource "/cronworkflows.argoproj.io" already exists
[terraform]: ╵
[terraform]: ╷
[terraform]: │ Error: Cannot create resource that already exists
[terraform]: │ 
[terraform]: │   with module.argo-workflows[0].kubernetes_manifest.customresourcedefinition_workfloweventbindings_argoproj_io,
[terraform]: │   on modules/kubernetes/services/argo-workflows/main.tf line 443, in resource "kubernetes_manifest" "customresourcedefinition_workfloweventbindings_argoproj_io":
[terraform]: │  443: resource "kubernetes_manifest" "customresourcedefinition_workfloweventbindings_argoproj_io" {
[terraform]: │ 
[terraform]: │ resource "/workfloweventbindings.argoproj.io" already exists
[terraform]: ╵
[terraform]: ╷
[terraform]: │ Error: Cannot create resource that already exists
[terraform]: │ 
[terraform]: │   with module.argo-workflows[0].kubernetes_manifest.customresourcedefinition_workflows_argoproj_io,
[terraform]: │   on modules/kubernetes/services/argo-workflows/main.tf line 501, in resource "kubernetes_manifest" "customresourcedefinition_workflows_argoproj_io":
[terraform]: │  501: resource "kubernetes_manifest" "customresourcedefinition_workflows_argoproj_io" {
[terraform]: │ 
[terraform]: │ resource "/workflows.argoproj.io" already exists
[terraform]: ╵
[terraform]: ╷
[terraform]: │ Error: Cannot create resource that already exists
[terraform]: │ 
[terraform]: │   with module.argo-workflows[0].kubernetes_manifest.customresourcedefinition_workflowtasksets_argoproj_io,
[terraform]: │   on modules/kubernetes/services/argo-workflows/main.tf line 1488, in resource "kubernetes_manifest" "customresourcedefinition_workflowtasksets_argoproj_io":
[terraform]: │ 1488: resource "kubernetes_manifest" "customresourcedefinition_workflowtasksets_argoproj_io" {
[terraform]: │ 
[terraform]: │ resource "/workflowtasksets.argoproj.io" already exists
[terraform]: ╵
[terraform]: ╷
[terraform]: │ Error: Cannot create resource that already exists
[terraform]: │ 
[terraform]: │   with module.argo-workflows[0].kubernetes_manifest.customresourcedefinition_workflowtemplates_argoproj_io,
[terraform]: │   on modules/kubernetes/services/argo-workflows/main.tf line 1554, in resource "kubernetes_manifest" "customresourcedefinition_workflowtemplates_argoproj_io":
[terraform]: │ 1554: resource "kubernetes_manifest" "customresourcedefinition_workflowtemplates_argoproj_io" {
[terraform]: │ 
[terraform]: │ resource "/workflowtemplates.argoproj.io" already exists
[terraform]: ╵
```
I then manually deleted all the argo crds and redeployed successfully.

I then edited one of the CRDs and redeployed again and it deployed the modified CRD fine.

Adam-D-Lewis:
So this PR would require the users to manually delete the 8 argo CRDs (`kubectl delete crd clusterworkflowtemplates.argoproj.io`, etc.) before updating, but future changes to argo CRDs wouldn't be necessary since they are going to be managed by Terraform from here on out.

Adam-D-Lewis:
Will the PR be able to be merged given that it requires some one time manual update steps by existing users?  Is there some way to prompt users about manual steps during an update?

kcpevey:
Seems like we have an issue to add more documentation around updating versions but I can't seem to find it atm. That bit of info should at the very least be included there.

iameskild:
Hey @Adam-D-Lewis, thanks for the update! I was really hoping to avoid having users run any `kubectl` commands when they wanted to upgrade; this might require users to download and install tools they don't regularly use. We could get around this by having a more robust `nebari upgrade` command/workflow, but since we're not there, I'm not sure what other options we have. Do you have any other ideas?

@costrouc do you know of any way to force delete outdated CRDs?

Adam-D-Lewis:
It appears that this problem was a result of having the CRDs in the root of the helm chart so the maintainers of the argo workflow helm chart moved the CRDs to correct the issue moving forwards (https://github.com/argoproj/argo-helm/issues/1785).

I don't think we need to define the CRDs ourselves either since they say that one of the benefits of moving the CRDs was to fix the error we're seeing with updates.

iameskild:
That looks promising, great find @Adam-D-Lewis! 

So if I understand correctly, simply removing any reference to the Arg-Workflow CRDs (and just relying on the base helm chart) will still create the CRDs automatically? 
I guess my second question is, whether this would resolve the need to manually delete the existing CRDs for Nebari deployments as we upgrade from the old version of Argo-Workflows to the latest version?

Adam-D-Lewis:
By the way, the upgrade step would look like this for users.
image

Adam-D-Lewis:
I tested the latest version locally.  I think users will need to disable argo_workflows, deploy, then enable, and redeploy before this will work correctly.

Adam-D-Lewis:
@iameskild Yeah, I think this is ready for review now.

Adam-D-Lewis:
From Dharhas: "You need a link to docs that explain the whole process including how to set up kubectl." 
@viniciusdc is working on PR to explain how to set up kubectl and afterwards I can a link in the text that is displayed during the update step before me merge this PR.

Adam-D-Lewis:
I made Dharhas's suggested change.  Here's the prompt now: 

image

nebari/upgrade.py
```diff
@@ -385,6 +386,32 @@ def _version_specific_upgrade(
...
+
+        rich.print(
+            f"\n\n[bold cyan]Note:[/] Upgrading requires a one-time manual deletion of the Argo Workflows Custom Resource Definitions (CRDs). \n\n[red bold]Warning:  [link=https://{config['domain']}/argo/workflows]Workflows[/link] and [link=https://{config['domain']}/argo/workflows]CronWorkflows[/link] created before deleting the CRDs will be erased when the CRDs are deleted and will not be restored.[/red bold] \n\nThe updated CRDs will be installed during the next [cyan bold]nebari deploy[/cyan bold] step. Argo Workflows will not function after deleting the CRDs until the updated CRDs are installed in the next nebari deploy. You must delete the Argo CRDs before upgrading to {self.version} or deploy step will fail.  Please delete them before proceeding by generating a kubeconfig (see [link=https://www.nebari.dev/docs/how-tos/debug-nebari/#generating-the-kubeconfig]docs[/link]), installing kubectl (see [link=https://www.nebari.dev/docs/how-tos/debug-nebari#installing-kubectl]docs[/link]), and running the following command:\n\n\t[cyan bold]{kubectl_delete_argo_crds_cmd} [/cyan bold].\n"
```
iameskild, line 402:
```suggestion
            f"\n\n[bold cyan]Note:[/] Upgrading requires a one-time manual deletion of the Argo Workflows Custom Resource Definitions (CRDs). \n\n[red bold]Warning:  [link=https://{config['domain']}/argo/workflows]Workflows[/link] and [link=https://{config['domain']}/argo/workflows]CronWorkflows[/link] created before deleting the CRDs will be erased when the CRDs are deleted and will not be restored.[/red bold] \n\nThe updated CRDs will be installed during the next [cyan bold]nebari deploy[/cyan bold] step. Argo Workflows will not function after deleting the CRDs until the updated CRDs are installed in the next nebari deploy. You must delete the Argo CRDs before upgrading to {self.version} or deploy step will fail.  Please delete them before proceeding by generating a kubeconfig (see [link=https://www.nebari.dev/docs/how-tos/debug-nebari/#generating-the-kubeconfig]docs[/link]), installing kubectl (see [link=https://www.nebari.dev/docs/how-tos/debug-nebari#installing-kubectl]docs[/link]), and running the following command:\n\n\t[cyan bold]{kubectl_delete_argo_crds_cmd} [/cyan bold]\n"
```
The final full-stop appears as part of the command, I think it'd be better if we removed it. Thoughts?

nebari/upgrade.py
```diff
@@ -385,6 +386,32 @@ def _version_specific_upgrade(
...
+            ""
+        )
+
+        continue_ = Prompt.ask("Have you deleted the Argo CRDs? \[y/N]", default="N")
+        if not continue_ == "y":
+            raise ValueError(
```
iameskild, line 408: Can we just `exit()` instead?

```python
if not continue_ == "y":
    print(f"You must delete the Argo CRDs before upgrading to {self.version}")
    exit()
```

This avoids the large traceback being printed to the screen and allows the user to see the command they need to run.

iameskild (changes requested):
This is looking great @Adam-D-Lewis! 

I just tested this myself and things seem to be working as expected. I just left two minor comments but otherwise, I feel like this is ready to be merged.

nebari/upgrade.py
Adam-D-Lewis: good idea, I pushed a commit

iameskild (approved):
This looks great to me! As I mentioned yesterday, I tested this on an existing test cluster and everything seemed to work as expected. Thanks @Adam-D-Lewis 🎉

PR #1630, opened 2023-01-31
Set GKE release_channel to unspecified to prevent auto k8s updates

iameskild:
## What does this implement/fix?

_Put a `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds a feature)
- [ ] Breaking change (fix or feature that would cause existing features not to work as expected)
- [ ] Documentation Update
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [x] Build related changes
- [ ] Other (please describe):

## Testing

- [x] Did you test the pull request locally?
- [ ] Did you add new tests?

I created a new Nebari cluster to test that this would work as expected. I also upgraded an existing cluster to validate that it would work for those clusters as well. No issues detected. Node pools may be destroyed and recreated if there are version differences. More information about `release_channel` can be found [here](https://cloud.google.com/kubernetes-engine/docs/concepts/release-channels) and [here](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#release_channel).

PR #1648, opened 2023-02-09
Add option to specify GKE release channel

### LINKED ISSUE #1608: [ENH] - Specify k8s version and release channel for GCP
### Feature description

When deploying a new instance of nebari on GCP, I'd like to be able to specify both the k8s version and the [release channel](https://cloud.google.com/kubernetes-engine/docs/concepts/release-channels). 

Currently the k8s version can be specified in the nebari config, but not the release channel IIRC. This should be doable in terraform via the [release_channel](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#release_channel) attribute.

Also I'm fairly sure the k8s version given in the nebari config is ignored during deployment. E.g., from a recent deployment I have different values in nebari config (1.23.13-gke.900) from what I ended up getting in the running cluster (1.23.12-gke.1600).

### Value and/or benefit

If the release channel is not specified, the default release channel is used, which is the "regular" channel. However, this can result in k8s upgrades which then remove some k8s API functions that a nebari deployment relies on. This can lead to a state where you cannot run redeployment of an existing nebari service, because the deploy command fails due to k8s API changes. It also generates errors for a running deployment where components like traefik are trying to use API functions that have been removed.

To ensure long-term stability of a nebari deployment, it would be preferable to be able to specify the "stable" channel, or even to specify "no channel" which will give you a static GKE version. 

### Anything else?

This happened to us with a service initially deployed using qhub 0.3. When k8s 1.22 was released into the regular channel, our cluster was upgraded, but several API functions were removed in this upgrade and we could no longer redeploy the service. This forced us to build a new service from scratch and migrate users over.

iameskild:
This small PR allows users to overwrite the default [`release_channel`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#release_channel) set for their GKE cluster. This is an enhancement to this recent PR: #1630 

## Reference Issues or PRs

Fixes #1608 

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?


iameskild:
@costrouc @viniciusdc @kcpevey this is ready for a final review. Thanks :)

viniciusdc (approved):
LGTM, though we will need to update https://www.nebari.dev/docs/explanations/custom-overrides-configuration with this new extra configuration

Adam-D-Lewis:
I agree with Vini, could you make a docs PR that would be merged along with this one? @iameskild

iameskild:
Thanks @Adam-D-Lewis @viniciusdc!! I'll add docs for this ASAP and ping you when it's ready 👍

iameskild:
@Adam-D-Lewis @viniciusdc here is the PR for the docs: https://github.com/nebari-dev/nebari-docs/pull/326

Adam-D-Lewis (approved):

Issue #1653, opened 2023-02-15
META - Add playwright tests to replace (& expand) Cypress


kcpevey:
### Feature description

Playwright is more user friendly than our current cypress tests. We'd like to move towards testing with playwright. The ultimate goal is to test all of our integrations through playwright on CI. 

### Value and/or benefit

Because its more user-friendly, adding more tests will require less effort and we'll be able to fill out the test suite. 

### Tasks

```[tasklist]
### Preliminary
- [ ] https://github.com/nebari-dev/nebari/pull/1665
```

```[tasklist]
### Basic tests
- [x] Login to nebari and launch a server
```

```[tasklist]
### Integration tests
- [x] Run a Jupyter notebook (high prio)
- [ ] Run a Python script via VSCode
- [ ] Run a Dask gateway job
- [ ] Run a kbatch job
- [ ] Create a new env in conda store (high prio)
- [ ] Jupyter Scheduler
```

kcpevey:
Basic
* [ ] Login to nebari and launch a server

Integration tests
* [ ] Run a jupyter notebook
* [ ] launch a dashboard via cds dashboards
* [ ] Run a python script via VS Code
* [ ] Run a dask gateway job
* [ ] Run a kbatch job
* [ ] Create a new environment in conda-store

trallard:
@kcpevey do we have a PR for this (even if WIP)?

kcpevey:
This is the initial - https://github.com/nebari-dev/nebari/pull/1665. It won't close this issue so its not directly linked to this issue (only mentioned)

kcpevey:
@aktech can you update the check boxes with the current state?

iameskild:
I think we can consider this closed. All of the remaining items are being tracked here: https://github.com/nebari-dev/nebari/issues/1958

PR #1665, opened 2023-02-28
Initial playwright setup

kcpevey:
## Reference Issues or PRs

First setup of #1653 

## Testing

- [ ] Did you add new tests?

### DIFF
--- a/tests_e2e/playwright/basic.py
+++ b/tests_e2e/playwright/basic.py
@@ -1,0 +1,264 @@
+
+
+from playwright.sync_api import Playwright, sync_playwright, expect
+import os
+import dotenv
+import time
+import logging
+import re
+
+dotenv.load_dotenv()
+
+logger = logging.getLogger()
+logger.setLevel(logging.DEBUG)
+
+console = logging.StreamHandler()
+console.setLevel(level=logging.DEBUG)
+formatter =  logging.Formatter('%(levelname)s : %(message)s')
+console.setFormatter(formatter)
+logger.addHandler(console)
+
+
+nebari_url = 'https://nebari.quansight.dev'
+google_email = os.environ['GOOGLE_EMAIL']
+google_password = os.environ['GOOGLE_PASSWORD']
+
+
+class Navigator():
+    """Base class for Nebari Playwright testing. This provides setup and 
+    teardown methods that all tests will need and some other generally useful
+    methods such as clearing the workspace. Specific tests such has "Run a 
+    notebook" will be a separate class which inherits from this one. It will
+    leverage the base class and add other specific tests. 
+
+    Navigator classes are able to run either standalone, or inside of pytest. 
+    This makes it easy to develop new tests, but also fully prepared to be
+    included as part of the test suite. 
+    """
+
+    def __init__(
+        self, 
+        nebari_url = 'https://nebari.quansight.dev',
+        google_email= os.environ['GOOGLE_EMAIL'],
+        username = os.environ['GOOGLE_EMAIL'],
+        google_password = os.environ['GOOGLE_PASSWORD'],
+        headless=False, 
+        slow_mo=100,
+        playwright=None,
+        browser=None,
+        context=None,
+        page=None,
+        ):
+        self.nebari_url = nebari_url
+        self.google_email = google_email
+        self.google_password = google_password
+        self.username = username
+        self.playwright = playwright
+        self.browser = browser
+        self.context = context
+        self.page = page
+
+        if self.playwright and self.browser and self.context and self.page:
+            logger.info('Existing playwright instance supplied.')
+        else:
+            logger.info('Creating new playwright instance.')
+            self.setup(browser=browser, headless=headless, slow_mo=slow_mo)
+        self.wait_for_server_spinup = 5 * 60 * 1000 # 5 minutes in ms
+
+    
+    def setup(self, browser='chromium', headless=False, slow_mo=100):
+        logger.debug('>>> Setting up browser for Playwright')
+
+        self.playwright = sync_playwright().start()
+        if browser == 'chromium':
+            self.browser = self.playwright.chromium.launch(headless=headless, slow_mo=slow_mo)
+        elif browser == 'webkit':
+            self.browser = self.playwright.webkit.launch(headless=headless, slow_mo=slow_mo)
+        elif browser == 'firefox':
+            self.browser = self.playwright.firefox.launch(headless=headless, slow_mo=slow_mo)
+        else:
+            raise RuntimeError(f'{browser} browser is not recognized.')
+        self.context = self.browser.new_context()
+        self.page = self.context.new_page()
+        
+
+    def teardown(self):
+        self.context.close()
+        self.browser.close()
+        self.playwright.stop()
+        logger.debug('>>> Teardown complete.')
+
+    def google_login_start_server(self) -> None:
+        """Go to a nebari deployment, login via Google, and start a new server."""
+        logger.debug('>>> Sign in via Google and start the server')
+        self.page.goto(self.nebari_url)
+        expect(self.page).to_have_url(re.compile(f"{self.nebari_url}*"))
+
+        self.page.get_by_role("button", name="Sign in with Keycloak").click()
+        self.page.get_by_role("link", name="Google").click()
+        self.page.get_by_role("textbox", name="Email or phone").fill(self.google_email)
+        self.page.get_by_role("button", name="Next").click()
+        self.page.get_by_role("textbox", name="Enter your password").fill(self.google_password)
+        # locator = self.page.get_by_role('link', name="lkjjfjjjjlkjsdf")
+        # locator.wait_for(timeout=300, state='attached')
+        time.sleep(0.5)
+        self.page.get_by_role("button", name="Next").click()
+        # TODO add expect title JuppyterrHubbb
+
+        time.sleep(0.5) # let the page load (use this to avoid a long wait in the try statement)
+
+        # if server is not yet running
+        start_locator = self.page.get_by_role("button", name="Start My Server")
+        try: 
+            start_locator.wait_for(timeout=3000, state='attached') 
+            start_locator.click()
+
+            # your server is not yet running, would like to start it?
+            self.page.get_by_role('button', name='Launch server').click()
+            # select instance type
+            self.page.locator('#profile-item-small-instance-with-conda-store-ui').click()
+            self.page.get_by_role("button", name="Start").click()
+
+        except:
+            # if the server is already running
+            start_locator = self.page.get_by_role("button", name="My Server", exact=True)
+            start_locator.wait_for(timeout=3000, state='attached')
+
+            start_locator.click()
+
+        # wait for redirect
+        self.page.wait_for_url(f'{self.nebari_url}/user/{self.username}/*')
+        # let page load after redirect
+        time.sleep(5)
+        
+
+    def _check_for_kernel_popup(self):
+        """Is the kernel popup currently open? 
+        
+        Returns
+        -------
+        True if the kernel popup is open. 
+        """
+
+        visible = self.page.get_by_text('Select Kernel', exact=True).is_visible()
+        return visible
+
+
+    def reset_workspace(self):
+        logger.debug('>>> Reset JupyterLab workspace')
+
+        # server is already running and there is no popup
+        popup = self._check_for_kernel_popup()
+
+        # server is on running and there is a popup
+        if popup:
+            self._set_environment_via_popup(kernel=None)
+            
+        # go to Kernel menu
+        kernel_menuitem = self.page.get_by_text("Kernel", exact=True)
+        kernel_menuitem.click()
+        # shut down multiple running kernels
+        try:
+            shut_down_all = self.page.get_by_text("Shut Down All Kernels...", exact=True)
+            shut_down_all.wait_for(timeout=300, state='attached')
+            shut_down_all.click()
+        except:
+            logger.debug('Did not find "Shut Down All" menu option')
+        
+        # shut down kernel if only one notebook is running
+        kernel_menuitem.click()
+        try:
+            shut_down_current = self.page.get_by_text("Shut Down Kernel", exact=True)
+            shut_down_current.wait_for(timeout=300, state='attached')
+            shut_down_current.click()
+        except:
+            logger.debug('Did not find "Shut Down Kernel" menu option')
+
+        # go back to root folder
+        self.page.get_by_title(f"/home/{google_email}", exact=True).locator("path").click()
+        
+        # go to File menu
+        self.page.get_by_text("File", exact=True).click()
+        # close all tabs
+        self.page.get_by_role('menuitem', name='Close All Tabs', exact=True).click()
+
+        # there may be a popup to save your work, don't save
+        if self.page.get_by_text('Save your work', exact=True).is_visible():
+            self.page.get_by_role('button', name='Discard', exact=True).click()
+
+    def _set_environment_via_popup(self, kernel=None):
+        """Set the environment kernel via the popup dialog box. 
+        If kernel is `None`, `No Kernel` is selected and the popup is dismissed. 
+
+        Attributes
+        ----------
+        kernel: str or None
+            (Optional) name of conda environment to set. Defaults to None. 
+            
+        """
+        if kernel is None:
+            # close dialog (deal with the two formats of this dialog)
+            try:
+                self.page.get_by_text("Cancel", exact=True).click()
+            except:
+                self.page.locator("div").filter(has_text="No KernelSelect").get_by_role("button", name="No Kernel").wait_for(timeout=300, state='attached')
+        else:
+            # set the environment
+            self.page.get_by_role("combobox").nth(1).select_option(f"{{\"name\":\"{kernel}\"}}")
+            # click Select to close popup (deal with the two formats of this dialog)
+            try:
+                self.page.get_by_role("button", name="Select", exact=True).click()
+            except:
+                self.page.locator("div").filter(has_text="No KernelSelect").get_by_role("button", name="Select").click()
+
+
+    def set_environment(self, kernel):
+        
+        popup=self._check_for_kernel_popup()
+        # if there is not a kernel popup, make it appear
+        if not popup:
+            self.page.get_by_text("Kernel", exact=True).click()
+            self.page.get_by_role("menuitem", name="Change Kernel…").get_by_text("Change Kernel…").click()
+
+        self._set_environment_via_popup(kernel)
+
+
+class RunNotebook(Navigator):
+
+
+    def run_notebook(self, path, expected_output_text):
+        logger.debug('>>> Run test notebook')
+        
+        # TODO: add nbgitpuller here?
+
+        # navigate to specific notebook
+        file_locator = self.page.get_by_text("File", exact=True)
+
+        file_locator.wait_for(timeout=300000, state='attached') # 5 minutes max for server to spin up
+        file_locator.click()
+        self.page.get_by_role("menuitem", name="Open from Path…").get_by_text("Open from Path…").click()
+        self.page.get_by_placeholder("/path/relative/to/jlab/root").fill(path)
+        self.page.get_by_role("button", name="Open").click()
+        # give the page a second to open, otherwise the options in the kernel menu will be disabled. 
+        time.sleep(1)
+        # select kernel
+        # self._handle_kernel_popup(kernel='conda-env-global-global-dashboard-v3-py')
+        self.set_environment(kernel='conda-env-global-global-dashboard-v3-py')
+
+        # make sure that this notebook is one currently selected
+        self.page.get_by_role("tab", name=path).get_by_text(path).click()
+
+        # restart run all cells
+        self.page.get_by_text("Kernel", exact=True).click()
+        self.page.get_by_role("menuitem", name="Restart Kernel and Run All Cells…").get_by_text("Restart Kernel and Run All Cells…").click()
+        self.page.get_by_role("button", name="Restart", exact=True).click()
+
+        # expect(self.page.get_by_text(expected_output_text, exact=True)).to_have_text(f"{expected_output_text}") # not sure why this doesn't work
+        assert self.page.get_by_text(expected_output_text, exact=True).is_visible()
+
+
+nav = RunNotebook()
+nav.google_login_start_server()
+nav.reset_workspace()
+nav.run_notebook(path='dashboard_panel.ipynb', expected_output_text='success 3333')
+nav.teardown()
--- a/.github/workflows/test-playwright.yaml
+++ b/.github/workflows/test-playwright.yaml
@@ -1,0 +1,56 @@
+name: "Playwright"
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/test.yaml"
+      - "tests/**"
+      - "tests_deployment/**"
+      - "tests_e2e/**"
+      - "scripts/**"
+      - "nebari/**"
+      - "setup.cfg"
+      - "pyproject.yoml"
+  push:
+    branches:
+      - main
+      - develop
+      - release/\d{4}.\d{1,2}.\d{1,2}
+    paths:
+      - ".github/workflows/test.yaml"
+      - "tests/**"
+      - "tests_deployment/**"
+      - "tests_e2e/**"
+      - "scripts/**"
+      - "nebari/**"
+      - "setup.cfg"
+      - "pyproject.yoml"
+  workflow_call:
+
+jobs:
+  test-e2e:
+    name: "Playwright"
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version:
+          - "3.10"
+        browser:
+          - "chromium"
+    steps:
+      - name: "Checkout Infrastructure"
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install Nebari and Playwright
+        run: |
+          pip install .[dev]
+          playwright install
+      - name: Test Nebari using Playwright
+        run: |
+          pytest --version
+          pytest --browser ${{ matrix.browser }} --headed tests_e2e/playwright
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -80,6 +80,7 @@ dev = [
     "pre-commit",
     "pytest",
     "pytest-timeout",
+    "pytest-playwright",
 ]
 
 [project.urls]


tests_e2e/playwright/basic.py
```diff
@@ -0,0 +1,288 @@
...
+            timeout=300000, state="attached"
+        )  # 5 minutes max for server to spin up
+        file_locator.click()
+        self.page.get_by_role("menuitem", name="Open from Path…").get_by_text(
+            "Open from Path…"
+        ).click()
+        self.page.get_by_placeholder("/path/relative/to/jlab/root").fill(path)
+        self.page.get_by_role("button", name="Open").click()
+        # give the page a second to open, otherwise the options in the kernel menu will be disabled.
+        time.sleep(1)
```
costrouc, line 265: This is fragile. I would rely on the the locator api to check that a locator is e.g. visible. https://playwright.dev/docs/api/class-locator#locator-is-visible

kcpevey: I agree its fragile. I struggled getting `is_visible` and `wait_for` to work with pages loading, particularly after redirect. I will go back and give it more effort to see if I can get rid of the sleeps.

kcpevey: I resolved this using `self.page.wait_for_load_state('networkidle')`.

kcpevey:
A big challenge that I don't have a great solution for is sometimes jlab opens with a popup already open (requesting a kernel for a notebook) and sometimes it doesn't. Since its a popup, the only thing we have access to are the popup dialog items. The only solution I can find is to wrap a try/except around getting a popup dialog item. If it doesn't exist, it will fail and that's the only way I have to know that the popup doesn't exist. Not great, but I tried many other options and nothing worked.

costrouc:
@kcpevey I'd like to split this PR into smaller PRs after we merge this initial PR. We should limit the scope to replacing only the cypress tests as an MVP. If possible too I'd like to minimize the number of abstrations we create. Then we work on adding each smaller feature. This PR as is just isn't well scoped and hard to finish in a certain amount of time.

kcpevey:
@costrouc I completely agree. That's why I was hoping to get this first one merged before adding more tests. This just shows what I was hoping would be one "easy" integration test. I will add more details in the README on my vision of how the rest of the integration test suite would be filled out. I don't intend on removing cypress until we get a bunch of other PRs in and we're sure this is stable.

kcpevey:
@costrouc @iameskild 

* When should the playwright tests run? I was thinking *after* a merge to `develop` or `main`?
* Where should they run? I'm guessing after a test deployment of the current branch? Where can I find that? and the URL for it?

tests_e2e/playwright/basic.py
```diff
@@ -0,0 +1,302 @@
...
+from playwright.sync_api import expect, sync_playwright
+
+logger = logging.getLogger()
+logger.setLevel(logging.DEBUG)
+
+console = logging.StreamHandler()
+console.setLevel(level=logging.DEBUG)
+formatter = logging.Formatter("%(levelname)s : %(message)s")
+console.setFormatter(formatter)
+logger.addHandler(console)
```
kcpevey, line 15: need to clean this up

costrouc:
@kcpevey I'd like the playwright tests to run in https://github.com/nebari-dev/nebari/blob/main/.github/workflows/kubernetes_test.yaml#L165-L169

kcpevey:
> @kcpevey I'd like the playwright tests to run in https://github.com/nebari-dev/nebari/blob/main/.github/workflows/kubernetes_test.yaml#L165-L169 and remove https://github.com/nebari-dev/nebari/blob/main/.github/workflows/kubernetes_test.yaml#L138-L163.

Thanks! I think we should keep cypress for now and run both test suites until we have full coverage of the cypress tests with the playwright tests. Would you agree?

.github/workflows/kubernetes_test.yaml
```diff
@@ -92,12 +96,12 @@ jobs:
...
       - name: Add DNS entry to hosts
         run: |
-          sudo echo "172.18.1.100  github-actions.nebari.dev" | sudo tee -a /etc/hosts
+          sudo echo "172.18.1.100  $NEBARI_BASE_URL" | sudo tee -a /etc/hosts
       - name: Initialize Nebari Cloud
         run: |
           mkdir -p local-deployment
           cd local-deployment
-          nebari init local --project=thisisatest --domain github-actions.nebari.dev --auth-provider=password
+          nebari init local --project=thisisatest --domain $NEBARI_BASE_URL --auth-provider=password
```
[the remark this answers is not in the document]
viniciusdc: @kcpevey this should fix the problem with the certificates in the tests, but every new deployment will re-request a new cert which will eventually reach the monthly rate limit. So as an extended solution, I would recommend checking if playwright has some settings to enable accessing untrusted sites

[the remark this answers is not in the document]
kcpevey: Would you expect a setting for `ignore_https_errors` to work?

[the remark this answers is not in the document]
kcpevey: The `ignore_https_errors` errors appears to be working.

kcpevey:
Need to add asserts after each major step (or more often) to make errors easier to track. I just had a failure on the `My Server` try/except but it was because of an incorrect username/password on the previous page.

tests_e2e/playwright/test_playwright.py
```diff
@@ -0,0 +1,16 @@
+from run_notebook import RunNotebook
+
+
+def test_notebook(navigator):
+    test_app = RunNotebook(navigator=navigator)
+    test_app.nav.clone_repo(
+        "https://github.com/nebari-dev/nebari.git",
+        branch="add_playwright",
```
kcpevey, line 8: This needs to be removed just before merging.

tests_e2e/playwright/run_notebook.py
```diff
@@ -0,0 +1,101 @@
...
+        instance_name="small-instance",
+        headless=False,
+        slow_mo=100,
+    )
+    nav.login()
+    nav.start_server()
+    nav.reset_workspace()
+    test_app = RunNotebook(navigator=nav)
+    test_app.nav.clone_repo(
+        "https://github.com/nebari-dev/nebari.git", branch="add_playwright"
```
kcpevey, line 94: The branch needs to be removed just before merging.

kcpevey:
I see that this is failing CI. It was passing ((see here)[https://github.com/nebari-dev/nebari/actions/runs/4755557238/jobs/8449860728]) before I made an unassociated change to azure tests, which means the test is still flaky 🤯. I will sort that out.

kcpevey:
Ok, all the test pass except the k8s test failure which is also on `develop` (https://github.com/nebari-dev/nebari/actions/runs/4781817203/jobs/8511173379). Ready for review.

.github/workflows/test-provider.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
       matrix:
         provider:
           - aws
-          - azure
+          # - azure
```
pmeier, line 47: We do we exclude Azure here?

.gitignore
```diff
@@ -45,5 +46,7 @@ nebari-config.yaml
 
 .vscode/
 .pytest_cache
-
+.ipynb_checkpoints
+.DS_Store
 /.ruff_cache
+videos/
```
pmeier, line 52: Nit: can we move this up? Stuff that is very much project related should be at the top an not buried down below with the general ignores for all kinds of Python tools.

tests_e2e/cypress/integration/main.js
```diff
@@ -56,7 +56,7 @@ describe('First Test', () => {
         .should('have.attr', 'value', 'Start').click();
 
       // Minimal check that JupyterLab has opened
-      cy.get('div#jp-MainLogo', { timeout: 60000 }).should('exist').wait(2000);
+      cy.get('div#jp-MainLogo', { timeout: 60000 }).should('exist').wait(10000);
```
pmeier, line 59: Do we need this? Seems unrelated to the PR.

tests_e2e/playwright/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+logger = logging.getLogger()
+
+
+@pytest.fixture(scope="session")
+def _navigator_session(browser_name):
+    """Set up a navigator instance, login with username/password, start
+    a server. Teardown when session is complete.
+    Do not use this for individual tests, use `navigator` fixture
+    for tests."""
+    dotenv.load_dotenv()
```
pmeier, line 17: Can you explain why we need the whole `dotenv` package? Is this just for debugging locally? If so, can't we manage without an extra dependency? It seems just loading a Python file should suffice?

tests_e2e/playwright/navigator.py
```diff
@@ -0,0 +1,448 @@
...
+        elif browser == "webkit":
+            self.browser = self.playwright.webkit.launch(
+                headless=headless, slow_mo=slow_mo
+            )
+        elif browser == "firefox":
+            self.browser = self.playwright.firefox.launch(
+                headless=headless, slow_mo=slow_mo
+            )
+        else:
+            raise RuntimeError(f"{browser} browser is not recognized.")
```
pmeier, line 127: Nit

```suggestion
        try:
            self.browser = getattr(self.playwright, browser).launch(
                headless=headless, slow_mo=slow_mo
            )
        except AttributeError:
            raise RuntimeError(f"{browser} browser is not recognized.") from None
```

tests_e2e/playwright/navigator.py
```diff
@@ -0,0 +1,448 @@
...
+        logger.debug(">>> Teardown complete.")
+
+    def login(self) -> None:
+        """Login to nebari deployment using the auth method on the class."""
+        if self.auth == "google":
+            self.login_google()
+        elif self.auth == "password":
+            self.login_password()
+        else:
+            raise ValueError(f"Auth type of {self.auth} is invalid.")
```
pmeier, line 150: Nit: since we expect to add more login methods here, let's make this more extensible from the start:

```suggestion
        try:
            return {
                "google": self.login_google,
                "password": self.login_password,
            }()
        except KeyError:
            raise ValueError(f"Auth type of {self.auth} is invalid.") from None
```

tests_e2e/playwright/navigator.py
```diff
@@ -0,0 +1,448 @@
...
+        # if server is not yet running
+        start_locator = self.page.get_by_role("button", name="Start My Server")
+        try:
+            start_locator.wait_for(timeout=3000, state="attached")
+            start_locator.click()
+
+            # I'm not sure why this is no longer needed. Leaving it here for
+            # future reference.
+            # # your server is not yet running, would like to start it?
+            # self.page.get_by_role("button", name="Launch server").click()
```
pmeier, line 203: As discussed offline, let's remove this unless we need it.

tests_e2e/playwright/navigator.py
```diff
@@ -0,0 +1,448 @@
...
+
+            # # wait for server to spinup
+            # server_message = self.page.get_by_text(
+            #     f"Server ready at /user/{self.username}/",
+            #     exact=True,
+            # )
+            # server_message.wait_for(
+            #     timeout=self.wait_for_server_spinup,
+            #     state="attached",
+            # )
```
pmeier, line 218: Same as above.

tests_e2e/playwright/navigator.py
```diff
@@ -0,0 +1,448 @@
...
+        """Start a nebari server. There are several different web interfaces
+        possible in this process depending on if you already have a server
+        running or not. In order for this to work, we look for html elements,
+        wait because the page may need to load, and if they don't appear, we
+        check for another server start option.
+        """
+        # if server is not yet running
+        start_locator = self.page.get_by_role("button", name="Start My Server")
+        try:
+            start_locator.wait_for(timeout=3000, state="attached")
```
pmeier, line 197: Do we need both conditions here? If we have a precise way to query if something is loaded, i.e. `state="attached"`, do we also need a blanket timeout? What happens if the state is not "attached" after the timeout?

I saw this pattern multiple times. Same question applies.

tests_e2e/playwright/navigator.py
```diff
@@ -0,0 +1,448 @@
...
+        kernel_menuitem.click()
+        # shut down multiple running kernels
+        try:
+            shut_down_all = self.page.get_by_text(
+                "Shut Down All Kernels...", exact=True
+            )
+            shut_down_all.wait_for(timeout=300, state="attached")
+            shut_down_all.click()
+        except Exception:
+            pass
```
pmeier, line 287: Nit:

```suggestion
        with contextlib.suppress(Exception):
            shut_down_all = self.page.get_by_text(
                "Shut Down All Kernels...", exact=True
            )
            shut_down_all.wait_for(timeout=300, state="attached")
            shut_down_all.click()
```

Saw this pattern multiple times. Suggestion applies everywhere.

tests_e2e/playwright/navigator.py
```diff
@@ -0,0 +1,448 @@
...
+            format, e.g. https://github.com/nebari-dev/nebari.git)
+        branch: bool or str
+            (Optional) branch to checkout from the repo. Defaults to None to use
+            the default branch.
+
+        Returns
+        -------
+        None
+        """
+        import datetime as dt
```
pmeier, line 402: Why is this a lazy import? Any concerns putting it at the top with all other imports?

tests_e2e/playwright/navigator.py
```diff
@@ -0,0 +1,448 @@
...
+        # ensure that playwright doesn't move on/change context until all the
+        # above commands are complete.
+        time.sleep(wait_for_completion)
+        self.reset_workspace()
+
+
+if __name__ == "__main__":
+    import os
+
+    import dotenv
```
pmeier, line 433: Same as above. Why are those lazy imports?

tests_e2e/playwright/run_notebook.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+        if restart_dialog_button.is_visible():
+            restart_dialog_button.click()
+
+        output_locator = self.nav.page.get_by_text(expected_output_text, exact=True)
+        output_locator.wait_for(timeout=runtime)  # wait for notebook to run
+        assert output_locator.is_visible()
+
+
+if __name__ == "__main__":
```
pmeier, line 79: I don't really understand the purpose of this. The `Navigator` main is a smoke test for `nebari` in general. But this main here runs a fixed notebook similar to what the test is doing right? Since we are also cloning a repo here, we can't really use a local notebook that I want to test. Meaning, this main seems somewhat obsolete. Can we remove or did I miss something?

tests_e2e/playwright/test_playwright.py
```diff
@@ -0,0 +1,16 @@
+from run_notebook import RunNotebook
+
+
+def test_notebook(navigator):
+    test_app = RunNotebook(navigator=navigator)
+    test_app.nav.clone_repo(
```
pmeier, line 6: Why do we need to clone a repo here and not just use the repo we are already in?

pmeier (commented):
@kcpevey Thanks a ton for this PR! `nebari` desperately needs more testing and this is an awesome first step!

I did an initial pass on just the stuff I noticed while looking at the diff. I will run all of this next week and do another pass afterwards.

As discussed offline, I prefixed some comments with "Nit:". They are optional, and we can fix them later. I prefer to fix them here though, and happy to do so myself. Please react with :+1: if you want handle it, and with :-1: if you want me to handle it.

tests_e2e/playwright/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+            password=os.environ["PASSWORD"],
+            headless=False,
+            browser=browser_name,
+            auth="password",
+            instance_name=os.environ["INSTANCE_NAME"],
+            video_dir="videos/",
+        )
+        nav.login_password()
+        nav.start_server()
+        yield nav
```
pmeier, line 34: What if we fail to create the `Navigator`? We should oust this outside the `try / except` or otherwise the `nav.teardown()` fails as well.

tests_e2e/playwright/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+            nebari_url=os.environ["NEBARI_FULL_URL"],
+            username=os.environ["USERNAME"],
+            password=os.environ["PASSWORD"],
+            headless=False,
+            browser=browser_name,
+            auth="password",
+            instance_name=os.environ["INSTANCE_NAME"],
+            video_dir="videos/",
+        )
+        nav.login_password()
```
pmeier, line 32: Developing locally, I don't have a domain. Meaning, I'm just using an IP, e.g. `172.18.1.100` as domain. In this case, the login fails, because playwright can't connect. It needs to be `https://172.18.1.100`. This seems to be handled automatically for proper domains, but we can't guarantee this.

tests_e2e/playwright/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+        nav.start_server()
+        yield nav
+    except Exception as e:
```
pmeier, line 35: I was hit with a timeout error here quite often while stepping through this with a debugger. We need an option to increase the overall timeout or disable it completely to enable local development.

tests_e2e/playwright/navigator.py
```diff
@@ -0,0 +1,448 @@
...
+        self.browser = browser
+        self.instance_name = instance_name
+        self.video_dir = video_dir
+
+        self.setup(
+            browser=self.browser,
+            headless=self.headless,
+            slow_mo=self.slow_mo,
+        )
+        self.wait_for_server_spinup = 30000  # 5 * 60 * 1000  # 5 minutes in ms
```
pmeier, line 83: This is missing a `0`. As is, we have a timeout for 30 seconds or 0.5 minutes.

```suggestion
        self.wait_for_server_spinup = 300_000  # 5 * 60 * 1_000  # 5 minutes in ms
```

tests_e2e/playwright/navigator.py
```diff
@@ -0,0 +1,448 @@
...
+                "button",
+                name="My Server",
+                exact=True,
+            )
+            start_locator.wait_for(timeout=3000, state="attached")
+            start_locator.click()
+
+        # wait for server spinup
+        self.page.wait_for_url(
+            f"{self.nebari_url}user/{self.username}/*",
```
pmeier, line 232: This is missing a slash?

```suggestion
            f"{self.nebari_url}/user/{self.username}/*",
```

tests_e2e/playwright/navigator.py
```diff
@@ -0,0 +1,448 @@
...
+    def start_server(self) -> None:
+        """Start a nebari server. There are several different web interfaces
+        possible in this process depending on if you already have a server
+        running or not. In order for this to work, we look for html elements,
+        wait because the page may need to load, and if they don't appear, we
+        check for another server start option.
+        """
+        # if server is not yet running
+        start_locator = self.page.get_by_role("button", name="Start My Server")
+        try:
```
pmeier, line 196: Instead of trying to click a button that might not be there, can't we just check that directly? We could just check if `start_locator` is visible, no? If not, we can then try the other button.

.github/workflows/kubernetes_test.yaml
```diff
@@ -149,10 +153,21 @@ jobs:
...
+      - name: Playwright Tests
+        env:
+          USERNAME: ${{ env.CYPRESS_EXAMPLE_USER_NAME }}
+          PASSWORD: ${{ env.CYPRESS_EXAMPLE_USER_PASSWORD }}
+          NEBARI_FULL_URL: "https://${{ env.NEBARI_BASE_URL }}/"
+          INSTANCE_NAME: "small-instance"
+        working-directory: tests_e2e/playwright
+        run: |
+          envsubst < .env.tpl > .env
+          xvfb-run pytest --browser chromium
```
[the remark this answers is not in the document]
pmeier: I never needed the `xvfb-run` command on my local system. Is this required just for CI?

pmeier (commented):
I've run the test locally now and got it to pass after some back and forth. There are quite a few rough edges that we need to address. The biggest for me is the lack of documentation. I'm comfortable to drop into our CI configs to see what we are actually running, but we can't expect that from everyone. Anything that goes beyond `pytest $TEST_FOLDER`, has to be documented.

I haven ot yet gone through all the branches in the navigator / notebook runner to see if I can reproduce. I'll do that as soon as all the other things are addressed.

.github/workflows/kubernetes_test.yaml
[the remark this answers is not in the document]
kcpevey: `xvfb-run` is needed for running headed mode, which is what the tests currently run. I dont remember the details here if headed is required for videos. I'll have to test it.

.github/workflows/kubernetes_test.yaml
```diff
@@ -161,6 +176,8 @@ jobs:
           path: |
             ./tests_e2e/cypress/screenshots/
             ./tests_e2e/cypress/videos/
+            ./tests_e2e/playwright/videos/
+            ${{ env.NEBARI_CONFIG_PATH }}
```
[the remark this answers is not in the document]
kcpevey: No, this needs to be cleaned up.

tests_e2e/cypress/integration/main.js
kcpevey: This has been changed in main now

tests_e2e/playwright/.env.tpl
```diff
@@ -0,0 +1,4 @@
+USERNAME=""
+PASSWORD=""
+NEBARI_FULL_URL=""
+INSTANCE_NAME=""
```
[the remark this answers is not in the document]
kcpevey: I will add some info on what goes in here. I would say that the tests should be small enough to use the (default) small instance type. The Navigator class can handle alternative instance types... but in order to leverage that we would have to shift to creating Navigator instances (setup and login) for every test. This is because we can't have two playwright objects running at the same time.

tests_e2e/playwright/conftest.py
kcpevey: Are you suggesting I do `from dotenv import load_dotenv` ?

tests_e2e/playwright/conftest.py
kcpevey: I dont understand where the error is here. Is the error because this code is trying to access an https that is unavailable? Or is this code injecting https and it shouldnt? Or does the nebari login require https but you don't have it?

tests_e2e/playwright/navigator.py
kcpevey: The cypress url in GH Action has a test, I was unifying both tests so this slash is not necessary. A better solution would be to handle either case in the init.

tests_e2e/playwright/navigator.py
kcpevey: This one is an oversight.

tests_e2e/playwright/navigator.py
kcpevey: This one is intentional as its not needed for the remainder of the file, only when this is run as main.

.github/workflows/kubernetes_test.yaml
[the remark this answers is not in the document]
pmeier: Happy to keep it if we need it for CI. In that case, we should add a comment there explaining as much.

.github/workflows/kubernetes_test.yaml
```diff
@@ -161,6 +176,8 @@ jobs:
           path: |
             ./tests_e2e/cypress/screenshots/
             ./tests_e2e/cypress/videos/
+            ./tests_e2e/playwright/videos/
```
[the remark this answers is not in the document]
pmeier: Thanks for the heads-up, I didn't know that. Here is the relevant part of the gitignore documentation:

> - If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular `.gitignore` file itself. Otherwise the pattern may also match at any level below the `.gitignore` level.
> - If there is a separator at the end of the pattern then the pattern will only match directories, otherwise the pattern can match both files and directories.
> - For example, a pattern `doc/frotz/` matches `doc/frotz` directory, but not `a/doc/frotz` directory; however `frotz/` matches `frotz` and `a/frotz` that is a directory (all paths are relative from the `.gitignore` file).

Since we currently have

```gitignore
# tests
videos/
```

we are indeed ignoring `videos` folders everywhere.

tests_e2e/playwright/navigator.py
pmeier: So is this runtime optimization? This was added as dev requirement and thus should always be available.

tests_e2e/playwright/run_notebook.py
pmeier: Could you walk me to the process on how to run a local notebook then? IIUC, I would need to

1. Fix the `nebari_url` parameter
2. Set the `USERNAME` / `PASSWORD` environment variables either manually or through `dotenv`
3. Comment out the whole `clone_repo` step
4. Fix the path in `run_notebook` to the local notebook that I want to test.

If I would do the same through a regular test, the steps would be

1. Copy a test like `test_notebook`
2. Comment out the whole `clone_repo` step
3. Fix the path in `run_notebook` to the local notebook that I want to test.
4. Set the `USERNAME` / `PASSWORD` / ... environment variables either manually or through `dotenv`
5. Run `pytest`

Unless I'm missing something here, this is somewhat equivalent. The upside of going the `pytest` route is that you already have written your test, which is what you likely want to do when running a notebook like this.

tests_e2e/playwright/.env.tpl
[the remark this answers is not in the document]
pmeier: Fair enough. Should we actually hard-code the instance type then? Not sure if we need to collect it from an environment variable in that case.

tests_e2e/playwright/conftest.py
pmeier: For me `self.page.goto(self.nebari_url)` failed if `self.nebari_url` was just an IP. I needed to prefix that manually with `f"https://{self.nebari_url}"` to get it to work.

tests_e2e/playwright/conftest.py
pmeier: With `pytest-playwright` installed, I see this in the help message:

```
Interrupt test run and dump stacks of all threads after a test times out:
  --timeout=TIMEOUT     Timeout in seconds before dumping the stacks.  Default is 0 which
                        means no timeout.
  --timeout_method={signal,thread}
                        Deprecated, use --timeout-method
  --timeout-method={signal,thread}
                        Timeout mechanism to use.  'signal' uses SIGALRM, 'thread' uses a timer
                        thread.  If unspecified 'signal' is used on platforms which support
                        SIGALRM, otherwise 'thread' is used.
```

Could we maybe use that?

tests_e2e/playwright/conftest.py
pmeier: No, the question is rather "Do we need dotenv at all"? Unless I'm missing something, we only use it to set some environment variables that we later use inside the test. Can't we just set them manually? We do we need an extra dependency for that?

.github/workflows/kubernetes_test.yaml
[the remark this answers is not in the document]
kcpevey: comment added

tests_e2e/playwright/conftest.py
kcpevey: I like it for easy-of-use so that I don't have to set env vars to run a test. I can keep a `.env` in my local repo, have it gitignored, and everything just works. Otherwise, I'd have to manually set them in the py file before I ran locally. Which would then cause a git diff. I added it as `dev` requirement, which I didn't think would be that big a deal. If you take serious issue with it, I can remove.

tests_e2e/playwright/navigator.py
kcpevey: You dont have to have both, but this will wait for the attached state, and use the timeout of 3 seconds instead of the default 30 seconds. That helps speed things up.

kcpevey: I did rework this a bit though

tests_e2e/playwright/navigator.py
kcpevey: yes, runtime optimization. Its added for ease-of-use when developing or running locally. On CI, the env vars are set in the GH Action yaml.

tests_e2e/playwright/run_notebook.py
kcpevey: >Comment out the whole clone_repo step

This runs on the nebari instance, which doesn't have the local copy of the repo so this is still needed. 

Other than that, yes, they are equivalent. `pytest` will replicate CI, obviously, but I like to run as main to allow me to use the VS Code debugger. Developer convenience.

tests_e2e/playwright/conftest.py
kcpevey: The `https` is always required. I'll add a note to the docs

tests_e2e/playwright/run_notebook.py
pmeier: As discussed offline, I was mistakenly under the impression that both scenarios I mentioned above give the same DevX. I did no know that the VS Code debugger is a mess when debugging tests. Thus, this main has a good reason to exist.

.github/workflows/kubernetes_test.yaml
```diff
@@ -153,6 +154,19 @@ jobs:
         with:
           working-directory: tests_e2e
 
+      - name: Playwright Tests
+        env:
+          USERNAME: ${{ env.CYPRESS_EXAMPLE_USER_NAME }}
+          PASSWORD: ${{ env.CYPRESS_EXAMPLE_USER_PASSWORD }}
+          NEBARI_FULL_URL: https://github-actions.nebari.dev/
+          INSTANCE_NAME: "small-instance"
```
pmeier, line 162: I think this can go now?

tests_e2e/playwright/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+`pip install -e ".[dev]"`
+
+Then install playwright itself (rerquired).
+
+`playwright install`
+
+:::note
+If you see the warning `BEWARE: your OS is not officially supported by Playwright; downloading fallback build.`, it is not critical. Playwright will likely still
+work https://github.com/microsoft/playwright/issues/15124
```
pmeier, line 31: This doesn't render properly for me

Screenshot from 2023-05-11 11-02-13

tests_e2e/playwright/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Create environment file
+
+Create a copy of the `.env` template file
+
+```bash
+cp .env.tpl .env
+```
+
```
pmeier, line 40: We should add a sentence here that one has to put these values in the newly created `.env` file.

tests_e2e/playwright/README.md
```diff
@@ -0,0 +1,84 @@
...
+* USERNAME: Nebari username for username/password login OR Google email address or Google sign in
+* PASSWORD: Password associated with USERNAME
+* NEBARI_FULL_URL: full url path to Nebari instance, e.g. "https://nebari.quansight.dev/"
+
+## Writing Playwright tests
```
pmeier, line 45: Could we also add a section about how to run the tests? IIUC, it should be just the regular `pytest` invocation, right? Maybe also add a sentence about the stored videos?

tests_e2e/playwright/conftest.py
```diff
@@ -0,0 +1,46 @@
...
+            instance_name="small-instance",  # small-instance included by default
+            video_dir="videos/",
+        )
+        nav.login_password()
+        nav.start_server()
+        yield nav
+    except Exception as e:
+        logger.debug(e)
+        with contextlib.suppress(NameError):
+            nav.teardown()
```
pmeier, line 39: IIUC, creating a `Navigator` instance just spawns a browser. Meaning, the only thing that can go wrong there is a wrong browser name (error on our side) or the browser cannot be spawned (error on `playwright` side). If that is correct, I would prefer two `try / except` blocks.

Functionally that won't make a difference, but I think it makes the implementation easier to understand. While correct,

```python
with contextlib.suppress(NameError):
    nav.teardown()
```
is a little too much for my taste.

```suggestion
        try:
        nav = Navigator(
            nebari_url=os.environ["NEBARI_FULL_URL"],
            username=os.environ["USERNAME"],
            password=os.environ["PASSWORD"],
            headless=False,
            browser=browser_name,
            auth="password",
            instance_name="small-instance",  # small-instance included by default
            video_dir="videos/",
        )
    except Exception as e:
        logger.debug(e)
        raise

    try:
        nav.login_password()
        nav.start_server()
        yield nav
    except Exception as e:
        logger.debug(e)
        raise
    finally:
        nav.teardown()
```

tests_e2e/playwright/navigator.py
pmeier: Maybe just different opinions on the style here. We usually don't scope imports by usage inside a file, i.e. if an import is only used in one function, it still goes at the top of the file. The only reason I use lazy imports is if the package in question is an optional dependency that might not be available, but is not needed for most parts of the module. That is not the case here. `os` is from the standard library and `dotenv` is a dev requirement.

I would prefer to put them at the top, but I won't block over this. Feel free to resolve without fix if you disagree with me.

tests_e2e/playwright/test_playwright.py
pmeier: So the general dev workflow would be

1. Create the notebook
2. Push it to a branch
3. Repeat 2. every time you make a change to the notebook

? That is quite inconvenient. Is there no better way to get this file into the cluster whether it is deployed locally or remotely?

In any case, we definitely need to document this workflow in the README.

tests_e2e/playwright/navigator.py
```diff
@@ -0,0 +1,433 @@
...
+        )
+
+        # the jupyter page loads independent of network activity so here
+        # we wait for the File menu to be available on the page, a proxy for
+        # the jupyterlab page being loaded.
+        file_locator = self.page.get_by_text("File", exact=True)
+        file_locator.wait_for(
+            timeout=self.wait_for_server_spinup,
+            state="attached",
+        )
```
pmeier, line 226: I asked about this before. Could you explain why we need both conditions? My understanding is the following:

We wait for state to become `"attached"` **or** until the timeout. But what happens if it was actually the timeout that ended the waiting? IIUC, the code afterwards just assumes that the state is reached. But what if that never happens? Can't we get errors down the line that will be hard to debug because they are based on false assumptions?

Since we have the global timeout we are in no danger to run into an infinite wait here. Thus, unless I'm mistaken above, we should probably remove the timeout from `wait_for` if we have a way to determine if the locator reached the state we want.

tests_e2e/playwright/navigator.py
pmeier: Continuation in https://github.com/nebari-dev/nebari/pull/1665#discussion_r1190905879.

tests_e2e/playwright/navigator.py
pmeier: Ah, ok. That makes more sense. Meaning, this state has to be reached within the timeout frame and otherwise we error out, right? I'm ok with that then, but I don't think it is really necessary as we have the global timeout. Up to you if we want to keep it.

pmeier:
We need to at least resolve https://github.com/nebari-dev/nebari/pull/1665#discussion_r1190893467 before merge. Some documentation is missing and possibly some workflow improvements for better DevX. There are a few more open comments, but nothing blocking.

kcpevey:
I'm going to try to carve out some time to get this wrapped up this week.

aktech:
@kcpevey @pmeier @iameskild  @costrouc  I have changed the git clone to write file on the nebari instance via bash using `cat` and `EOF` to remove the dependency of having tests in a branch on git. I think this pull request is in good shape to be merged. I wanted to wrap this up as I might need this for creating some testing setup. Let me know if anyone have any objections/reviews/comments.

kcpevey:
@aktech Thank you for taking this over! As for the reliance on a branch, that was just to show tests passing and not intended to be merged. Thank you improving things and pushing things through!

tests_e2e/playwright/.env.tpl
```diff
@@ -0,0 +1,3 @@
+USERNAME="USERNAME_OR_GOOGLE_EMAIL"
+PASSWORD="PASSWORD"
```
pmeier, line 2: These should be 

```suggestion
KEYCLOAK_USERNAME="USERNAME_OR_GOOGLE_EMAIL"
KEYCLOAK_PASSWORD="PASSWORD"
```

to avoid name clashes.

tests_e2e/playwright/README.md
```diff
@@ -0,0 +1,189 @@
...
+`playwright install`
+
+> If you see the warning `BEWARE: your OS is not officially supported by Playwright; downloading fallback build., it is not critical.` Playwright will likely still work microsoft/playwright#15124
+
+### Create environment file
+
+Create a copy of the `.env` template file
+
+```bash
+cp .env.tpl .env
```
pmeier, line 36:
```suggestion
cd tests_e2e/playwright
cp .env.tpl .env
```

tests_e2e/playwright/README.md
```diff
@@ -0,0 +1,189 @@
...
+```
+
+Fill in the newly created `.env` file with the following values:
+
+* USERNAME: Nebari username for username/password login OR Google email address or Google sign in
+* PASSWORD: Password associated with USERNAME
+* NEBARI_FULL_URL: full url path to Nebari instance, e.g. "https://nebari.quansight.dev/"
```
pmeier, line 43:
```suggestion
* NEBARI_FULL_URL: full url path including scheme to Nebari instance, e.g. "https://nebari.quansight.dev/"
```

tests_e2e/playwright/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Running the Playwright tests
+
+The playwright tests are run inside of pytest using
+
+```python
+pytest tests_e2e/playwright/test_playwright.py
+```
```
pmeier, line 51: This currently doesn't work, since the test expects `test_data` to be in the current directory.

pmeier (commented):
A few minor things inline that I'll fix in a second. Apart from them we also need to document that this needs a user to be present on the cluster and we can't do with `root`. @aktech will add this. Afterwards, LGTM.

pmeier (approved):
Thanks @kcpevey for the bulk of the work and @aktech for the polish. Happy with the result now. LGTM if CI is green!

aktech (approved):
thanks @kcpevey awesome work 🎉  and thanks @pmeier for persistence with this

PR #1699, opened 2023-04-07
Use --quiet flag for conda install in CI

pmeier:
Per title. Although we definitely want verbose output in CI, `--quiet` is actually harmless:

```
$ conda install --help | grep -e '--quiet'
  -q, --quiet           Do not display progress bar.
```

So this just disables just the progress bars, which create quite a few unnecessary lines. For example: https://github.com/nebari-dev/nebari/actions/runs/4621921970/jobs/8173947067#step:4:646


viniciusdc (approved):
Good point! Thanks @pmeier

PR #1700, opened 2023-04-07
improve pytest configuration

pmeier:
`pytest`'s default configuration is horrible with respect to a few aspects:

- traceback (`--tb`): default is `"long"` which not only shows a few lines above and below the one the error happened for each frame, but also the docstring for every stack. This easily buries the information I need in a lot of noise. `"short"` is better, but still pretty noisy. `"native"` is what Python gives you by default, which already includes a lot of information.
- xpass (`-rX` / `xfail_strict = True`): By default, the `@pytest.mark.xfail` is not strict. This means whether the test fails or not, `pytest` will never return a non-zero exit code. Since most people just look for a boolean signal from CI this can easily hide errors. With the `xfail_strict = True` setting, unexpectedly passing tests cause the run to fail.
- warnings (`-Wd`): `pytest` follows Pythons convention and adopts the same warning filters. This means for example that `DeprecationWarning`'s are not shown. However that is very useful information for developers. Setting `-Wd` disables the filter and thus all warnings are shown.

This PR adapts the `pytest.ini` configuration file to have more sensible "defaults".

cc @kcpevey since we touched on this offline before.


pytest.ini
```diff
@@ -1,3 +1,13 @@
...
+    # show tests that (f)ailed, (E)rror, or (X)passed in the summary
+    -rfEX
+    # Make tracebacks shorter
+    --tb=native
+    # enable all warnings
+    -Wd
 markers =
     conda: conda required to run this test (deselect with '-m \"not conda\"')
+testpaths =
+    tests
```
pmeier, line 12: This is more of a QoL improvement. with this one can type `pytest` without specifying anything else to run these tests. LMK if you want me to revert this.

pytest.ini
```diff
@@ -1,3 +1,13 @@
 [pytest]
+addopts =
+    # show tests that (f)ailed, (E)rror, or (X)passed in the summary
+    -rfEX
+    # Make tracebacks shorter
+    --tb=native
+    # enable all warnings
+    -Wd
```
pmeier, line 8: I personally would go a step further

```suggestion
    # turn warnings into error
    -We
```

Right now there aren't any warnings so this is a good time to adopt such a change. With this option set, we get an early heads-up if something is about the change in the future.

pmeier:
I also added the `pytest.ini` file the to CI workflow triggers

kcpevey (approved):
These look good to me. Not sure if we want one more approval on this. Maybe from @iameskild or @viniciusdc ?

costrouc:
Team agrees this is a useful commit to have warnings as errors.

PR #1706, opened 2023-04-10
Adding conda-store as a docker registry within nebari

costrouc:
A blocker for allowing conda-store to be a registry for nebari was allowing the kubernetes cluster in an authenticated fashion be able to pull images. This adds the needed kubernetes secret and exposes the registry via port `5000`. 

This feature will only work with trusted https certificates at the moment. So as of now this can not be too tightly coupled to how things are done in nebari since we still use self-signed certificates occasionally. I'm unsure how to solve this in an automated way.

### DIFF
--- a/nebari/template/stages/04-kubernetes-ingress/modules/kubernetes/ingress/main.tf
+++ b/nebari/template/stages/04-kubernetes-ingress/modules/kubernetes/ingress/main.tf
@@ -2,11 +2,13 @@ locals {
   default_cert = [
     "--entrypoints.websecure.http.tls.certResolver=default",
     "--entrypoints.minio.http.tls.certResolver=default",
+    "--entrypoints.docker-registry.http.tls.certResolver=default",
   ]
   certificate-settings = {
     lets-encrypt = [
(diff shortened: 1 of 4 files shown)


Adam-D-Lewis:
This feature will only work with trusted https certificates at the moment. So as of now this can not be too tightly coupled to how things are done in nebari since we still use self-signed certificates occasionally. I'm unsure how to solve this in an automated way. - Isn't using certmanager at least an option for handling this?

costrouc:
> Isn't using certmanager at least an option for handling this? I'm seeing a similar issue in that I have to make the nebari-workflows admission-controller public (go through traefik) b/c kubernetes needs to communicate with the admission controllers over TLS.

Sadly not an option since it is nearly impossible to make the docker/containerd image puller not verify tls across clouds and on-prem. At least from the research that I did. https://stackoverflow.com/questions/53545732/how-do-i-access-a-private-docker-registry-with-a-self-signed-certificate-using-k

iameskild:
Hi @costrouc, I think closing this PR makes sense at this time. That said, if you disagree, feel free to reopen it and I can review it. Thanks :)

PR #1710, opened 2023-04-12
improve CLI tests

pmeier:

Our current CLI tests are not sufficient, since our checking logic does not do what it is supposed to be doing. We only catch errors in the subprocess:

https://github.com/nebari-dev/nebari/blob/daecbcf37497b1b8b4e3f0f52667ababf3c023ad/tests/test_cli.py#L15-L23

However, `subprocess.call` does not fail on exit codes > 0:

```bash
$ ls foo
"foo": No such file or directory (os error 2)
$ echo $?
2
$ python -c 'import subprocess; subprocess.call(["ls", "foo"])'
ls: cannot access 'foo': No such file or directory
$ echo $?
0
```

This PR uses [`subprocess.run`](https://docs.python.org/3/library/subprocess.html#subprocess.run) instead, which is the [recommended way](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module) and has the `check=True` flag to enable error code checking.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,7 +1,4 @@
-import os
 import subprocess
-import typing
-from pathlib import Path
 
 import pytest
 
@@ -10,17 +7,6 @@ from nebari.utils import load_yaml
 
 PROJECT_NAME = "clitest"
 DOMAIN_NAME = "clitest.dev"
-
-
-def run_cli_cmd(command: str, working_dir: typing.Union[str, Path]):
-    """Run the provided CLI command using subprocess."""
-    try:
-        os.chdir(working_dir)
-        subprocess.call(command.split())
-    except subprocess.CalledProcessError:
-        return False
-
-    return True
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


tests/test_cli.py
```diff
@@ -32,28 +18,35 @@ def run_cli_cmd(command: str, working_dir: typing.Union[str, Path]):
 )
 def test_nebari_init(tmp_path, namespace, auth_provider, ci_provider, ssl_cert_email):
     """Test `nebari init` CLI command."""
-    command = f"nebari init local --project {PROJECT_NAME} --domain {DOMAIN_NAME} --disable-prompt"
+    command = [
```
pmeier, line 21: Minor refactoring. Instead of building a string here and splitting it again afterwards, let's just use a list from the start.

tests/test_cli.py
```diff
@@ -67,7 +60,6 @@ def test_nebari_init(tmp_path, namespace, auth_provider, ci_provider, ssl_cert_e
         assert ci_cd.get("type", {}) == ci_provider
     else:
         assert ci_cd == ci_provider
-        ci_cd = config.get("ci_cd", None)
```
pmeier, line 70: Driveby: this wasn't used anywhere.

iameskild (approved):
Nice catch @pmeier! This looks good to me!

PR #1720, opened 2023-04-12
use -V (upper case) for --version short form

### LINKED ISSUE #1719: [ENH] - use -V instead of -v as shortform for --version
### Feature description

Per title. In the Python ecosystem, `-V` (upper case) is usually used to display version information:

```
❯ python -V
Python 3.8.16
❯ pytest -V
pytest 7.2.2
```

`-v` (lower case) is usually reserved `--verbose`, which we currently don't have, but maybe will have in the future.

### Value and/or benefit

Consistency lowers entrance burden.

### Anything else?

_No response_

pmeier:
## Reference Issues or PRs

Fixes #1719.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

## Any other comments?

This could be considered BC breaking, but I feel that would be a tad too much.

### DIFF
--- a/nebari/cli/main.py
+++ b/nebari/cli/main.py
@@ -82,7 +82,7 @@ app.add_typer(
 def version(
     version: Optional[bool] = typer.Option(
         None,
-        "-v",
+        "-V",
         "--version",
         help="Nebari version number",
         is_eager=True,


pavithraes:
@iameskild @pmeier -- Would this be a breaking/surprising change for existing users? If so, maybe can have a deprecation warning or an error message for the lowercse `-v` as we change this behavior?

pmeier:
> Would this be a breaking

Technically, yes. 

> surprising change for existing users

IMHO, no. As explained in #1719, with this change, our CLI is less surprising than before.

> If so, maybe can have a deprecation warning or an error message for the lowercse `-v` as we change this behavior?

We could, but I wouldn't recommend it. For the reasons above and since we don't even have a BC policy as of now, I don't see the point. LMK if folks feel otherwise. Deprecation is not hard to implement here.

Issue #1707, opened 2023-04-11
[BUG] - local deploy cannot guaranteed be done without having access to a domain


pmeier:
### Describe the bug

Related to #1703. TL;DR: our guide tells users to fix `/etc/hosts`, but the deployed cluster doesn't see this change. There are three ways to achieve that:

1. Have access to a domain and create a DNS entry for `--domain=...` to point to `172.18.1.100`. This is what we do for `github-actions.nebari.dev` to get our CI going:
   https://github.com/nebari-dev/nebari/blob/daecbcf37497b1b8b4e3f0f52667ababf3c023ad/.github/workflows/kubernetes_test.yaml#L95
2. Use a service like [nip.io](https://nip.io/) and set something like `--domain=172.18.1.100.nip.io` during `nebari init`. With this users no longer need to have a domain available, but still need to have access to the internet. Plus, we add an external dependency, since our guide then depends on nip.io being available.
3. Instead of using an URL, users could also use the IP directly, i.e. `--domain=172.18.1.100`. It is not as pretty as having an URL, but probably good enough for local deploy. The major advantage is that this can be done even without internet access making it the most reliable and easiest to use option of the three. Doing this, the user also doesn't need to fix `/etc/hosts`.

Regardless of what we decide we want to use, 2. and 3. (as well as our current docs) make an assumption that doesn't hold in 100% of the cases: the cluster IP is fixed at `172.18.1.100`. 

However, we don't know that. It should work most of the time, but I know for example that it doesn't work for @costrouc. Maybe he can fill in some details on when this won't be true?

Internally, we only know the correct IP after stage 4:

https://github.com/nebari-dev/nebari/blob/daecbcf37497b1b8b4e3f0f52667ababf3c023ad/nebari/deploy.py#L102-L104

Meaning, setting it upfront during `nebari init --domain` is not guaranteed to work.

### Expected behavior

I think we should allow setting no domain upfront and instead of asking the user to set up their DNS (when `--disallow-prompt` is not set)

https://github.com/nebari-dev/nebari/blob/daecbcf37497b1b8b4e3f0f52667ababf3c023ad/nebari/deploy.py#L128-L131

just use the IP we get from the load balancer.

### OS and architecture in which you are running Nebari

Linux

### How to Reproduce the problem?

Follow the [local deployment how-to](https://www.nebari.dev/docs/how-tos/nebari-local) in an environment where the load balancer doesn't use `172.18.1.100` as IP.

### Compute environment

None

[...]

Issue #1703, opened 2023-04-08
[BUG] - Installing nebari locally (local with kind) fails.


twaclaw:
### Describe the bug

Deploying nebari [locally](https://www.nebari.dev/docs/how-tos/nebari-local) fails. 

The health check for `https://domain/argo/` fails. 
Other endpoints work fine, and `argo worksflows` seems to be installed properly.

[log3.8.txt](https://github.com/nebari-dev/nebari/files/11183019/log3.8.txt)

### Expected behavior

Installation should go through.

### OS and architecture in which you are running Nebari

Archlinux: 6.2.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Wed, 22 Mar 2023 22:52:35 +0000 x86_64 GNU/Linux. I tried with both Python 3.8 and 3.11.

### How to Reproduce the problem?

Follow the steps in:
https://www.nebari.dev/docs/how-tos/nebari-local

### Command output

```bash session
nebari init local  --project projectname  --domain domain  --auth-provider password  --terraform-state=local

nebari deploy -c nebari-config.yaml --disable-prompt |tee log3.8.txt


[terraform]:   "keycloak" = {
[terraform]:     "health_url" = "https://domain/auth/realms/master"
[terraform]:     "url" = "https://domain/auth/"
[terraform]:   }
[terraform]:   "monitoring" = {
[terraform]:     "health_url" = "https://domain/monitoring/api/health"
[terraform]:     "url" = "https://domain/monitoring/"
[terraform]:   }
[terraform]: }
Attempt 1 health check failed for url=https://domain/argo/
Attempt 2 health check failed for url=https://domain/argo/
Attempt 3 health check failed for url=https://domain/argo/
Attempt 4 health check failed for url=https://domain/argo/
Attempt 5 health check failed for url=https://domain/argo/
Attempt 6 health check failed for url=https://domain/argo/
Attempt 7 health check failed for url=https://domain/argo/
Attempt 8 health check failed for url=https://domain/argo/
Attempt 9 health check failed for url=https://domain/argo/
Attempt 10 health check failed for url=https://domain/argo/
ERROR: Service argo-workflows DOWN when checking url=https://domain/argo/
```

### Versions and dependencies used.

`kind`: 0.18.0

`kubectl`:
Client Version: v1.26.3
Kustomize Version: v4.5.7
Server Version: v1.21.10

`nebari`: 2023.1.1

### Compute environment

kind

### Integrations

Argo

[...]

PR #1728, opened 2023-04-14
don't set /etc/hosts

pmeier:

There are basically two scenarios:

1. We have domain and can set the DNS. In that case there is no need to set `/etc/hosts` since the DNS is handling it for us
2. We don't have a domain. In that case setting `/etc/hosts` is not enough. See https://github.com/nebari-dev/nebari/issues/1703#issuecomment-1502988292 and #1707.

Thus, there is no need to set it at all. This PR removes it from CI to see if anything breaks (it shouldn't). If we move forward here, we also need to send a PR to the docs and remove that section there.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

### DIFF
--- a/.github/workflows/kubernetes_test.yaml
+++ b/.github/workflows/kubernetes_test.yaml
@@ -90,9 +90,6 @@ jobs:
       - name: Get routing table for docker pods
         run: |
           ip route
-      - name: Add DNS entry to hosts
-        run: |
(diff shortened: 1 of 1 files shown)


pmeier:
Closing this since the k8s tests are not run from forks:

https://github.com/nebari-dev/nebari/blob/4c64f6f4810fcf96039d326db4155eacf442a98b/.github/workflows/kubernetes_test.yaml#L35-L36

PR #1729, opened 2023-04-14
don't set /etc/hosts in CI

pmeier:

There are basically two scenarios:

1. We have domain and can set the DNS. In that case there is no need to set `/etc/hosts` since the DNS is handling it for us
2. We don't have a domain. In that case setting `/etc/hosts` is not enough. See https://github.com/nebari-dev/nebari/issues/1703#issuecomment-1502988292 and #1707.

Thus, there is no need to set it at all. This PR removes it from CI to see if anything breaks (it shouldn't). If we move forward here, we also need to send a PR to the docs and remove that section there.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?


pmeier:
It seems `github-actions.nebari.dev` 

https://github.com/nebari-dev/nebari/blob/daecbcf37497b1b8b4e3f0f52667ababf3c023ad/.github/workflows/kubernetes_test.yaml#L95

does not resolve and thus our k8s tests are failing here. 

However, this actually means that our k8s tests are not sufficient: as explained in 2. above, it is *not* sufficient to fix `/etc/hosts` for the cluster to work correctly. Since our tests are green normally, this indicates that we don't check thoroughly.

@costrouc can we have an A record for `github-actions.nebari.dev` that points to `172.18.1.100`? As explained in #1707 that is not 100% safe, but should work in our setup. I'll expand the current tests to also check the case with no domain with the fix for #1707.

iameskild:
Hey @pmeier, is this still a WIP?

costrouc:
@pmeier I've added github-actions.nebari.dev to our dns. I'll retrigger the dns.

costrouc (approved):
Yeah let's remove this since like you said this doesn't actually do anything within the Kubernetes deployment.

PR #1734, opened 2023-04-21
Fix? contributor test trigger

pmeier:

As seen in #1711 neither my maintainer status nor @iameskild's comment triggered the contributor tests. However, they are required by our branch protection rules. Meaning, #1711 is effectively stuck right now and cannot be merged. 

As explained in https://github.com/nebari-dev/nebari/pull/1734#issuecomment-1517695618, it is a pain to debug workflows that trigger on comments since they don't use the workflow file from the current PR, but only the one from the default branch. Thus, we probably need to merge this PR and try afterwards. I did that in my fork and it worked.

This PR does two things:

1. Instead of having one large condition, it is split in multiple atomic ones. Since we `&&` them together anyway, this can be trivially achieved.
2. Instead of having an `if:` trigger on multiple jobs, we define one "pre-job" and require that on all others through the `needs:` key. That makes it a lot easier if we ever want to change the condition.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

### DIFF
--- a/.github/workflows/contributor.yaml
+++ b/.github/workflows/contributor.yaml
@@ -32,7 +32,8 @@ jobs:
 
       - name: C
         if: contains(fromJson('["OWNER", "COLLABORATOR", "MEMBER"]'), github.event.comment.author_association)
-        run: echo ${{ fromJson('["OWNER", "COLLABORATOR", "MEMBER"]') }} | jq
+        run: |
+          echo '["OWNER", "COLLABORATOR", "MEMBER"]' | jq
 
 #  build:
 #    name: Build
--- a/.github/workflows/contributor.yaml
+++ b/.github/workflows/contributor.yaml
@@ -16,15 +16,24 @@ jobs:
         env:
           CTX: ${{ toJSON(github) }}
         run: |
-          set -x
-          A=${{ github.event.issue.pull_request && true}}
-          B=${{ contains(github.event.comment.body, '/bot run tests') }}
-          C1=${{ fromJson('["OWNER", "COLLABORATOR", "MEMBER"]') }}
-          C=${{ contains(fromJson('["OWNER", "COLLABORATOR", "MEMBER"]'), github.event.comment.author_association) }}
-          
-          if [[ A && B && C ]]; then
-            echo "Hooray!"
-          fi
+          exit 0
+
+      - name: A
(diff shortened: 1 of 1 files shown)


pmeier:
random comment

pmeier:
Ok, this will be a little tricky to debug, since 

https://github.com/nebari-dev/nebari/blob/4ac4693f90d16117713bd7e84bbb976b030528c2/.github/workflows/contributor.yaml#L3-L5

will only run against the latest commit of the default branch. Meaning, regardless of what changes we make in the PR in this file, they will not be used until after the PR is merged. I'll debug without the comment trigger, but we might have to merge a dummy PR and revert later if it doesn't do what we want it to do.

.github/workflows/contributor.yaml
```diff
@@ -31,25 +47,25 @@ jobs:
           reactions: 'rocket'
 
   kubernetes-tests:
+    needs: trigger-contributor-tests
     name: "Contributor #${{ github.event.issue.number }}"
-    if: ${{ github.event.issue.pull_request && contains(github.event.comment.body, '/bot run tests') && contains(fromJson('["OWNER", "COLLABORATOR", "MEMBER"]'), github.event.comment.author_association)}}
-    uses: nebari-dev/nebari/.github/workflows/kubernetes_test.yaml@release/2022.11.1
+    uses: nebari-dev/nebari/.github/workflows/kubernetes_test.yaml@release/2023.4.1
```
pmeier, line 52: Driveby. We should probably add this to our release checklist.

pmeier: One second thought: we do we want to run the "released" tests at all. Shouldn't this be `@develop`? Otherwise no changes to the default branch will be considered when running contributor tests.

iameskild: @pmeier I agree, `@develop` is what we want.

iameskild:
@pmeier do you want to work together on this? It would be great to get this working since we have new contributors opening PRs :)

pmeier:
Feel free to push to my branch

iameskild (approved):
As @pmeier pointed out in the PR description, to debug we need to merge this. As far as I can tell, this won't have any impact on the rest of the codebase/repo.

PR #1773, opened 2023-05-02
Pathlib everywhere

### LINKED ISSUE #1713: [ENH] - Use pathlib everywhere
### Feature description

Right now we are using both `os.path` and `pathlib`, e.g.

https://github.com/nebari-dev/nebari/blob/daecbcf37497b1b8b4e3f0f52667ababf3c023ad/nebari/render.py#L78

https://github.com/nebari-dev/nebari/blob/daecbcf37497b1b8b4e3f0f52667ababf3c023ad/nebari/render.py#L25

We should decide for one with a strong personal preference for `pathlib` on my side.

### Value and/or benefit

```
❯ python -c "import this" | sed '15q;d'
There should be one-- and preferably only one --obvious way to do it.
```

Although Python is giving us two ways here, we shouldn't make the same mistake. It is really annoying if one always has to check if the input is a path or a string.

### Anything else?

_No response_

pmeier:
## Reference Issues or PRs

Closes #1713.

7k951s

We are Buzz Lightyear here and Woody is someone stuck on Python < 3.4. As the title implies, this replaces `os.path` with `pathlib.Path` everywhere in the `nebari` package except for the templates. LMK if we want `pathlib` there as well.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?


pmeier:
There is one file missing that I need to refactor. After that it should only be getting the tests green. I'm not sure if I have time this week to address this. But it is still on my backlog.

nebari/cli/main.py
```diff
@@ -50,6 +50,34 @@
...
+
+
+def config_path_callback(value: Path) -> Path:
+    value = path_callback(value)
+    if not value.is_file():
+        raise ValueError(f"Passed configuration path {value} does not exist!")
+    return value
+
+
+CONFIG_PATH_OPTION: Path = typer.Option(
```
pmeier, line 64: This is the only refactor I've made. Since the configuration file and output path is used by multiple subcommands, I've defined them once at the top. This avoids duplication in the code and makes it easier to maintain.

Adam-D-Lewis:
@pmeier Looks like the Kubernetes tests fail https://github.com/nebari-dev/nebari/actions/runs/4940974457/jobs/8833155254 and also some merge conflicts

iameskild:
@pmeier it looks like we will need update this PR due to the recent changes to the API. Is this something you'll have time for?

pmeier:
I'm OOO for the better part of the next two weeks. Will pick it up after. Feel free to change the labels on the PR to indicstr that if necessary.

iameskild:
Hi @pmeier, I know we've gone back and forth on this PR but do you have time to wrap this up? It's also worth noting that many of the changes in here will likely be overridden by the work coming from the extension-mechanism. Perhaps it would be better to wait until that work is complete...

pmeier:
Unless there is some blocker I failed to see, I promise I finish it before tomorrows community meeting.

iameskild (approved):
Beautiful :)

PR #1778, opened 2023-05-03
make nebari API private

pmeier:
Implements 4. of nebari-dev/governance#37 in three steps:

1. 62379f4449c12473fa5bf04faf92b6f2710e2370: Introduce a `src/nebari` layout. Right now we have the package directly in our repository root. Since we will ultimately have a `nebari` as well as a `_nebari` package, it is useful to tuck them together in one folder. This is done by [`pytest`](https://github.com/pytest-dev/pytest/tree/main/src), which we are following with this API change, and many other Python projects as well.
2. be85b66378c9036a24eb4425361bc7956c3c6796: make `nebari` package private by renaming it to `_nebari`.
3. 3181f5779dc615278118ebd00e37f135167d5050: reintroduce a public `nebari` package that does nothing but expose the CLI and a `__version__` attribute.

I'll send one commit per step outlined above to make it easier to review. Since the diff will get quite large, I advise reviewers to go commit by commit.

This PR will conflict with basically every other open PR and every patch that was merged into `develop` after I have started. To keep disruption to a minimum, I plan to only rebase once after this PR has been approved. For all other open PRs, I'm available for helping to migrate them. Tag me there if you need help.


pmeier:
Same deal as in https://github.com/nebari-dev/nebari/pull/1778#issuecomment-1560840288 with be85b66378c9036a24eb4425361bc7956c3c6796.

costrouc (approved):
Looks great to me @pmeier and I'm in favor of merging this earlier rather than later. Once you give the merge conflicts I'm in favor of immediately merging. 

I like our minimal nebari interface now of `nebari/__init__.py` and `nebari/__main__.py`

PR #1602, opened 2022-12-14
ENH - Switch to ruff and pre-commit.ci

trallard:
This PR adds the following changes:

- Drop flake8 in favour of ruff
- Moves ruff configurations to `pyproject.toml`
- Enables pre-commit.ci for autoupdates and lining in PRs
- Removes pre-commit GitHub action in favour of pre-commit.ci: note that the `terraform_fmt` cannot be run in pre-commit.ci so this will run in GH actions anyway

## Reference Issues or PRs

Closes gh-1589

## What does this implement/fix?

_Put a `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds a feature)
- [ ] Breaking change (fix or feature that would cause existing features not to work as expected)
- [ ] Documentation Update
- [x] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Other (please describe):

## Testing

- [x] Did you test the pull request locally?
- [ ] Did you add new tests?

PR #1782, opened 2023-05-03
improve pre-commit run on GHA

pmeier:
The comment

https://github.com/nebari-dev/nebari/blob/ac981ea6fe535cdc4592bffd60a49fe9ab137028/.github/workflows/run-precommit.yaml#L22-L23

is not the full picture. The action also [installs `pre-commit` and runs it against all files by default](https://github.com/pre-commit/action/blob/5f528da5c95691c4cf42ff76a4d10854b62cbb82/action.yml#L11-L20). Meaning, we are currently installing `pre-commit` for nothing 

https://github.com/nebari-dev/nebari/blob/ac981ea6fe535cdc4592bffd60a49fe9ab137028/.github/workflows/run-precommit.yaml#L25-L26

and running the `terraform_fmt` hook twice:

https://github.com/nebari-dev/nebari/blob/ac981ea6fe535cdc4592bffd60a49fe9ab137028/.github/workflows/run-precommit.yaml#L28-L29

This PR improves this by only running the `terraform_fmt` hook through the action.

### DIFF
--- a/.github/workflows/run-precommit.yaml
+++ b/.github/workflows/run-precommit.yaml
@@ -19,11 +19,7 @@ jobs:
       - name: Checkout repository 🔔
         uses: actions/checkout@v3
 
-      # https://github.com/pre-commit/action to enable cache
-      - uses: pre-commit/action@v3.0.0
-
-      - name: Install pre-commit 📦
-        run: pip install pre-commit
-
       - name: Run terraform pre-commit ⚡️
-        run: pre-commit run terraform_fmt
+        uses: pre-commit/action@v3.0.0
+        with:
+          extra_args: --all-files terraform_fmt
--- a/nebari/template/stages/01-terraform-state/aws/main.tf
+++ b/nebari/template/stages/01-terraform-state/aws/main.tf
@@ -5,7 +5,7 @@ variable "name" {
 
 variable "namespace" {
   description = "Namespace to create Kubernetes resources"
-  type= string
+  type        = string
 }
 
 module "terraform-state" {
--- a/nebari/template/stages/01-terraform-state/aws/main.tf
+++ b/nebari/template/stages/01-terraform-state/aws/main.tf
@@ -5,7 +5,7 @@ variable "name" {
 
 variable "namespace" {
   description = "Namespace to create Kubernetes resources"
-  type        = string
+  type= string
 }
 
 module "terraform-state" {


pmeier:
1c2ba6ba7c032b32556263b0f6e0384ae30390c9 demonstrates that we still [catch formatting issues](https://github.com/nebari-dev/nebari/actions/runs/4871035759/jobs/8687511719?pr=1782#step:3:67).

.github/workflows/run-precommit.yaml
```diff
@@ -19,11 +19,7 @@ jobs:
...
-      - uses: pre-commit/action@v3.0.0
-
-      - name: Install pre-commit 📦
-        run: pip install pre-commit
-
       - name: Run terraform pre-commit ⚡️
-        run: pre-commit run terraform_fmt
+        uses: pre-commit/action@v3.0.0
+        with:
+          extra_args: --all-files terraform_fmt
```
aktech, line 25: Seems like this is only running pre-commit for terraform, not for rest of the code.

Like we did earlier: https://github.com/nebari-dev/nebari/actions/runs/4873704426/jobs/8693649224#step:3:63

```bash
Run pre-commit run --show-diff-on-failure --color=always --all-files
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
check json...............................................................Passed
check yaml...............................................................Passed
check toml...............................................................Passed
check that executables have shebangs.....................................Passed
codespell................................................................Passed
black....................................................................Passed
ruff.....................................................................Passed
isort....................................................................Passed
Terraform fmt............................................................Passed
mdformat.................................................................Passed

```

pmeier: Not sure what you are saying here. We don't want to run all pre-commit hooks here. This is what we have [pre-commit CI](https://results.pre-commit.ci/run/github/252841745/1683111024.eUdbjnNgTrW3HAHR79urTQ) for. It was introduced in #1602 and runs everything but `terraform_fmt`:

https://github.com/nebari-dev/nebari/blob/7653c6700a2bfdfb4e4c07e2e869618663dfe96f/.pre-commit-config.yaml#L15-L20

The reason for this split setup is that `terraform_fmt` requires `terraform` to be available on the system and that is not possible on pre-commit CI.

aktech: Ah, yes. I didn't realise that. Thanks for explaining.

aktech (approved):

Issue #1618, opened 2023-01-11
ENH - Updates to kubernetes tests triggers


trallard:
## TODO

- [ ] Add explicit `ci skip` in a given commit to skip the kubernetes tests
- [ ] Add workflow dispatch to trigger tests manually
- [ ] other needed improvements to triggers and optimization of our resources

kcpevey:
Check for changes added [here](https://github.com/nebari-dev/nebari/pull/1665). I just noticed we were duplicating a hard coded nebari url so I turned it into an env var. Likely won't affect this issue, but wanted to mention it in case you notice this as well :)

Adam-D-Lewis:
I believe this is out of date, closing.  Feel free to re-open if this issue and add more details if this issue is still relevant.

PR #1797, opened 2023-05-09
Update trigger configuration for kubernetes tests

viniciusdc:
## Reference Issues or PRs
Related to https://github.com/nebari-dev/nebari/issues/1618

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?


viniciusdc:
image

.github/workflows/kubernetes_test.yaml
```diff
@@ -4,33 +4,34 @@ on:
...
-      - "tests/**"
       - "tests_deployment/**"
       - "tests_e2e/**"
-      - "scripts/**"
-      - "nebari/**"
-      - "pyproject.toml"
-      - "pytest.ini"
+      - "nebari/stages/**"
+      - "nebari/template/stages/**"
+
```
Adam-D-Lewis, line 11: Why don't we test e.g. nebari/cli/** or nebari/*.py anymore?  Seems like it'd be good to run the kubernetes tests if we make changes to the cli.

viniciusdc: I am divided in that matter, some commands I definitely agree should be tested within a running instance (e.g. keycloak), but at the end of the day, the CLI is just a wrapper for deployment, which I think we could get around by writing proper tests just for it and keep the kubernetes default trigger run just for e2e. 

As a second thought, I can add a way for triggering this action by adding a label to the PR or a comment, so that it can be manually triggered when we would like a e2e but the changes were done to a untracked location

iameskild:
@viniciusdc I think this PR will need to be updated with the changes to the API that Philip made recently 👍

viniciusdc:
> @viniciusdc I think this PR will need to be updated with the changes to the API that Philip made recently +1

Thanks @iameskild I will push these updates today afternoon

.github/workflows/kubernetes_test.yaml
```diff
@@ -4,33 +4,34 @@ on:
...
       - ".github/workflows/kubernetes_test.yaml"
-      - "tests/**"
       - "tests_deployment/**"
       - "tests_e2e/**"
-      - "scripts/**"
-      - "src/**"
-      - "pyproject.toml"
-      - "pytest.ini"
+      - "src/_nebari/stages/**"
+      - "src/_nebari/template/stages/**"
```
iameskild, line 10: What do you think about running these tests for changes made to `src/**`?

viniciusdc: IMO, all the Python scripts under `src` should have their unit tests, and this action should only care about changes made directly to the components that interfere with the deployment itself. Though I might be overseeing something important there, do you have any code in mind that should be tested here and part of `src` ?

iameskild: I see what you're saying. My opinion is that the source code is so closely coupled with the stages/templates that are created. I'll let you make the final decision ;)

iameskild:
Hey @viniciusdc do you have some time to wrap this PR up? It looks like we're very close to being done. We will need to update the validate-helm chart script so that it reflects the recent API changes. If you're too busy, let me know and I can try getting to this. Thanks :)

PR #1799, opened 2023-05-10
fix deprecated file deletion

pmeier:

The function

https://github.com/nebari-dev/nebari/blob/8bd56f90b84c7686f45459ae8f7cf690721bd60a/nebari/render.py#L229-L238

is called only once from

https://github.com/nebari-dev/nebari/blob/8bd56f90b84c7686f45459ae8f7cf690721bd60a/nebari/render.py#L79-L95

As indicated by the variable name, `deleted_paths=DEPRECATED_FILE_PATHS` should schedule all deprecated files for deletion. However, inside `inspect_files`, `deleted_paths` is unconditionally overwritten:

https://github.com/nebari-dev/nebari/blob/8bd56f90b84c7686f45459ae8f7cf690721bd60a/nebari/render.py#L275-L279

Thus, the passed value of `deleted_paths` is completely ignored.

IIUC, the purpose of this block is to find all deprecated files that actually exist and return them so they can be deleted. This PR implements this.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?


costrouc (approved):
Wanted to write here just some history on this function on copying files and why I didn't choose a pre-built solution. 

1. This code was written with the idea that we would be moving towards an extension system. Meaning that  we have multiple `src` directories with one `destination` directory.  
2. Not all "files" in src are coming from files. Some are computed e.g. the terraform `*.tf.json` files
3. There are cases to consider:
   - a file exists in multiple `src` directories... (decision: which one to choose?)
   - a file from a `src` doesn't exist in `destination` (decision: easy decision just copy over)
   - a file from a `src` is different in `destination` (decision: easy copy over since `src` is source of truth but use should be notified of change)
   - a file in `destination` exists that isn't in any `src` directories (decision: keep? but notify user there is an unwatched file?)

Basically the multiple src issue and dynamic generation of files is what led me to write those functions.

PR #1803, opened 2023-05-11
make domain optional

### LINKED ISSUE #1707: [BUG] - local deploy cannot guaranteed be done without having access to a domain

pmeier:
## Reference Issues or PRs

Closes #1707.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

## Any other comments?

`nebari init` (with or without `--guided-init`) and `nebari deploy` work fine. However we get problems for everything else since we never write back the IP of the load balancer that we find during deploy. I think we have two options here:

1. Write back the IP back. This probably a bad idea, since a deploy would potentially change the config. Since this should only happen for local deployments, it isn't too bad though.
2. Make an empty `domain` field a sentinel. During the verification stage, we could inspect the cluster, if available, and extract the domain from there.

### DIFF
--- a/nebari/cli/init.py
+++ b/nebari/cli/init.py
@@ -368,15 +368,20 @@ def guided_init_wizard(ctx: typer.Context, guided_init: str):
         # DOMAIN NAME
         rich.print(
             (
-                "\n\n 🪴  Great! Now you need to provide a valid domain name (i.e. the URL) to access your Nebri instance. "
-                "This should be a domain that you own.\n\n"
+                "\n\n 🪴  Great! Now you can provide a valid domain name (i.e. the URL) to access your Nebari instance. "
+                "This should be a domain that you own. FIXME Default is the IP of the load balancer\n\n"
             )
         )
-        inputs.domain_name = questionary.text(
-            "What domain name would you like to use?",
-            qmark=qmark,
-            validate=lambda text: True if len(text) > 0 else "Please enter a value",
-        ).unsafe_ask()
+        inputs.domain_name = (
+            questionary.text(
+                "What domain name would you like to use?",
+                qmark=qmark,
+            ).unsafe_ask()
+            or None
+        )
+
+        print("#" * 60)
+        print(repr(inputs.domain_name))
 
         # AUTH PROVIDER
         rich.print(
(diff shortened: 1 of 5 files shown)


nebari/cli/init.py
```diff
@@ -368,15 +368,20 @@ def guided_init_wizard(ctx: typer.Context, guided_init: str):
         # DOMAIN NAME
         rich.print(
             (
-                "\n\n 🪴  Great! Now you need to provide a valid domain name (i.e. the URL) to access your Nebri instance. "
-                "This should be a domain that you own.\n\n"
+                "\n\n 🪴  Great! Now you can provide a valid domain name (i.e. the URL) to access your Nebari instance. "
+                "This should be a domain that you own. FIXME Default is the IP of the load balancer\n\n"
```
pmeier, line 372: Need to put something else here.

nebari/cli/init.py
```diff
@@ -368,15 +368,20 @@ def guided_init_wizard(ctx: typer.Context, guided_init: str):
...
+        inputs.domain_name = (
+            questionary.text(
+                "What domain name would you like to use?",
+                qmark=qmark,
+            ).unsafe_ask()
+            or None
+        )
+
+        print("#" * 60)
+        print(repr(inputs.domain_name))
```
pmeier, line 384: Oops

```suggestion
```

nebari/cli/init.py
```diff
@@ -545,6 +551,8 @@ def if_used(key, model=inputs, ignore_list=["cloud_provider"]):
                 if isinstance(value, bool) and value:
                     return b.format(key=key, value=value).replace("_", "-")
 
+        print(inputs.dict())
+
```
pmeier, line 555: Oops

```suggestion
```

iameskild:
Hey @pmeier, what is the status of this PR?

viniciusdc:
Hi @pmeier, regarding what we discussed during the last community meeting, for the destroy part, if you would like the load balancer address (or any information from the resources) to be exposed, right now, we are getting it from terraform [here](https://github.com/nebari-dev/nebari/blob/develop/src/_nebari/template/stages/04-kubernetes-ingress/outputs.tf) we then use [this](https://github.com/nebari-dev/nebari/blob/a6d032f532e39bbd360955f29f6a0b5cc8c87b11/src/_nebari/provider/terraform.py#L157-L166) function to catch all the outputs (that terraform saves under a json file)

viniciusdc:
Also, it might be interesting to run the Kubernetes CI on this PR as well, so that we can see the error message as well. I am a bit curious to see where terraform might be complaining about the address, as this should only be failing in the schema validation

costrouc:
Closing since implemented in https://github.com/nebari-dev/nebari/pull/1833

PR #1804, opened 2023-05-15
don't fail CI matrices fast

pmeier:
## Reference Issues or PRs

It is really annoying to see a large matrix completely cancelled just because one job failed. For example: https://github.com/nebari-dev/nebari/actions/runs/4981394220/jobs/8915592409. This is especially true, if the individual parameters aren't strongly correlated. For example, if tests fail on Python 3.8 it is likely that they also fail on 3.9. In contrast, if the provider test fails on AWS, this gives very little insight if the same would also fail on GCP or the like.

Thus, unless we have very strict budget concerns here, it would improve DevX quite a bit to not fail the matrix builds fast. The infracost CI has `fail-fast: true` set, but there is no comment in the PR that added it (#1316) as for why.

- [ ] Documentation Update
- [ ] Build related changes

### DIFF
--- a/.github/workflows/infracost.yml
+++ b/.github/workflows/infracost.yml
@@ -26,12 +26,12 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
-      fail-fast: true
       matrix:
         provider:
           - aws
           - gcp
           - azure
+      fail-fast: false
 
     env:
       # Specify the path where the Terraform state will be stored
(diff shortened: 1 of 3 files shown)


Adam-D-Lewis (approved):

Adam-D-Lewis:
> It is really annoying to see a large matrix completely cancelled just because one job failed. For example: https://github.com/nebari-dev/nebari/actions/runs/4981394220/jobs/8915592409. This is especially true, if the individual parameters aren't strongly correlated. For example, if tests fail on Python 3.8 it is likely that they also fail on 3.9. In contrast, if the provider test fails on AWS, this gives very little insight if the same would also fail on GCP or the like.

I agree with this

Issue #1817, opened 2023-05-22
[BUG] - Infracost CI and DO tests failing


pavithraes:
_Originally reported by @pmeier on Quansight Slack_

`@Adam-D-Lewis` triggered some tests on `@pmeier`'s PR through the comment and the infracost jobs on AWS and Azure both fail because no credentials can be retrieved. Since Adam should have all the permissions, we need to figure out why this fails?

Refs:
- https://github.com/nebari-dev/nebari/actions/runs/5038088999/jobs/9035348504?pr=1778
- https://github.com/nebari-dev/nebari/actions/runs/5038088999/jobs/9035348581?pr=1778

The DO provider tests failing as well: https://github.com/nebari-dev/nebari/actions/runs/5016789600/jobs/8994092014#step:11:424

kcpevey:
I'm seeing the same thing on my PR https://github.com/nebari-dev/nebari/pull/1665

pmeier:
I need functioning CI for PRs like #1773 or #1778 since they are large scale refactorings.

kcpevey:
I'm guessing its related to our recent changes to our credentials infra. @iameskild or @costrouc might know more?

pmeier:
> As for Infracost, I personally don't think it's worth the maintenance burden of keeping integrated. As Dharhas said, it was hastily added so its utility at this moment is relatively limited.

Let me send a PR for that.

pmeier:
@dharhas I talked to @iameskild offline and we interpreted your https://github.com/nebari-dev/nebari/issues/1817#issuecomment-1557262871 differently. Do you mean we should just kill the CI jobs or rip it out completely? Currently it is part of our CLI

https://github.com/nebari-dev/nebari/blob/41116257ad07c094b29cf363fe52776573c063ee/nebari/cli/main.py#L385

However, it is marked "experimental"

https://github.com/nebari-dev/nebari/blob/41116257ad07c094b29cf363fe52776573c063ee/nebari/cli/main.py#L420

and thus we can likely purge it without BC concerns.

pmeier:
There is #1820 to remove the infracost capabilities. Re-labeling as `needs: investigation`, since the DO issue is not resolved yet.

pmeier:
[...]

PR #1819, opened 2023-05-22
Simplify CI

iameskild:
## Reference Issues or PRs

We've spent a decent amount of time trying to get CI working for PRs that come from a forked repo as well as maintaining CI workflows (such as Infracost) that are essentially never used. In this PR, I propose we simplify our CI setup as much as possible to reduce the maintenance burden. This PR eliminates the need to maintain the `Contributor` 
 and `Infracost` workflows. 

From the perspective of maintainers and developers, nothing should change. The only noticeable differences are that we now explicitly agree that we won't run the `Test Nebari Providers` on PRs coming from folks (this was never working before so this isn't really a change but more of an affirmation that this expected behavior) and instead run this test on a daily basis as a general precaution. This PR itself was opened from my fork of the repo so this is what we expect new contributors to see in terms of CI; first time contributors will still need a maintainer to approve and run CI.

- [ ] Documentation Update
- [ ] Build related changes
- [x] Other (please describe):

## Testing

- [ ] Did you add new tests?


.github/workflows/test-provider.yaml
```diff
@@ -1,6 +1,9 @@
+# This is only workflow that requires cloud credentials and therefore will not run on PRs coming from forks.
 name: "Test Nebari Provider"
 
 on:
+  schedule:
+    - cron: "0 3 * * *"
```
pmeier, line 6: We need a mechanism to alert us in case of failures here. I just let a bot open an issue in case of failure. For example

[workflow](https://github.com/pmeier/light-the-torch/blob/20a75752e156651bb3dea2f497e1cde03690b8f4/.github/workflows/check-pytorch-package-indices.yml#L39-L47)

[local action](https://github.com/pmeier/light-the-torch/blob/20a75752e156651bb3dea2f497e1cde03690b8f4/.github/actions/publish-from-template/action.yml#L12-L25) with [Python script](https://github.com/pmeier/light-the-torch/blob/20a75752e156651bb3dea2f497e1cde03690b8f4/.github/actions/publish-from-template/render_template.py#L8-L12) (this is technically not needed, but believe me when I tell you I got bitten so often because the template failed to render and thus no issue was opened although the workflow failed)

[issue template](https://github.com/pmeier/light-the-torch/blob/20a75752e156651bb3dea2f497e1cde03690b8f4/.github/issue-templates/packages-out-of-sync.md?plain=1#L1-L16)

If we don't have something like that, any failure will go undetected until we have a PR from the main repo.

iameskild: I have added the `Test Nebari Provider` badge to the repo README so we can quickly check if the tests are failing. My main concern is that we're spending so much time maintaining these additional workflows for a minor benefit. 

We have a triage meeting once a week and it might be worth adding a "Validate CI is passing" task to those meetings.

iameskild:
@pmeier thanks for the feedback!

The main issue with the `Test Nebari Providers` CI is that it requires secrets (specifically OIDC environments variables) that are not going to be available on forked repos ([from GH docs](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)). This means in order for to ensure that these tests get run when a PR is opened from a fork, we would need a workflow that copies over all of those changes to a new branch and then runs CI from that branch (or something equally complex); our current `Contributor` workflow does not handle this situation so either way, we would need to create a new workflow. 

In my personal opinion, at this point, I don't think the maintenance burden of this additional workflow outweighs the potential upside of catching an error (which is now more easily visible from the repo README). This is a tradeoff I am willing to make but I'm eager to hear what other people have to say.

.github/workflows/test-provider.yaml
pmeier: > My main concern is that we're spending so much time maintaining these additional workflows for a minor benefit.

Agreed. But what I proposed above just enhances visibility after we have killed the workflows from running on PRs.

> I have added the `Test Nebari Provider` badge to the repo README so we can quickly check if the tests are failing.

A red badge on README is a lot less visible than an issue. Are you committing to check this batch on a daily basis?

pmeier:
After some offline chat with @iameskild, I've implemented my suggestion from https://github.com/nebari-dev/nebari/pull/1819#discussion_r1200854901 in 9fff9545b4f6ac6190957b976e25a0cc569d3c4a. However, since this PR is coming from a fork, it is impossible to test here :smile: So we either merge as is (after review of course, but untested by CI) and fix later if something comes up, or we revert the patch again and merge it after the main PR has been merged.

.github/workflows/infracost.yml
```diff
@@ -1,170 +0,0 @@
-name: Infracost CI
```
pmeier, line 1: Given that we want to rip out infracost completely (https://github.com/nebari-dev/nebari/issues/1817#issuecomment-1557838379), we should probably do it in a separate PR.

pmeier: Rolled back in 59ee8b7d6ac54a086f7bdee2d14f3cead6ec65d6.

viniciusdc:
This looks great, and after pondering around it for some time, it does make sense that the cloud provider tests run daily instead of per contributor PR. 

The reason for it is that pytest (specifically the `tests/test_init.py`) is intended to perform validation of the CLI commands and make sure any attempt to connect to the provider API succeeded (doing so, by running init). So, if one of those tests were to fail, this is not the responsibility of the user, as this only showcases that currently, nebari is unable to connect to a provider:
- This could be interpreted as a token/credentials issue by our side and should then be fixed as soon as possible (thus the need for the issue) or that the provider API itself is down (which also needs to be communicated)

This should not block a contribution for both scenarios as it's not part of the contributor's responsibility. 

So, with that in mind, my only suggestion to change is to update the trigger of the tests workflow to only run for a contributor in case there were changes to the tests itself.

iameskild:
Thanks for the feedback @viniciusdc! 

I paired down the trigger paths for `Test Nebari Providers` based on your comment.

pmeier (approved):
Need a decision on https://github.com/nebari-dev/nebari/pull/1819#issuecomment-1557837653, but otherwise good to go from my side. Thanks Eskild!

iameskild:
CI/CD LucidCharts have been updated. This is ready to merge.

PR #1839, opened 2023-06-21
Enable extensible schema for nebari-config

iameskild:
## Reference Issues or PRs

Extend the base plugin (`NebariStage`) to include a `stage_schema` which can be used to extend the `Main` schema.

cc @costrouc 

## What does this implement/fix?

_Put a `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds a feature)
- [ ] Breaking change (fix or feature that would cause existing features not to work as expected)
- [ ] Documentation Update
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Other (please describe):

## Testing

- [ ] Did you test the pull request locally?
- [ ] Did you add new tests?

PR #1844, opened 2023-06-26
Add NebariPluginManager, allow schema to be dynamically extended

iameskild:
## Reference Issues or PRs

Supersedes #1839. 
This PR refactors a few things and adds the ability to dynamically update the `NebariConfig` schema:
- Creates the `NebariPluginManager` class which is responsible for the following items:
  - The `pluggy.plugin_manager` and determining which plugins (stages and/or subcommands) are active.
  - The Typer CLI (which includes `info` by default) and the aforementioned pluggable subcommands.
  - The `NebariConfig` schema and extending it where needed.
    - At the moment, none of the stages have `stage_schema` specified but the `_extend_schema` method allows for plugins to the define their own schema which is then appended to the `Main` schema to create the final schema. 

In the long run, it is my opinion that Nebari codebase is mainly responsible for how the `NebariPluginManager`, which means we should think about:
- How pluggable stages are defined, namely: 
	- What other stages does this stage depend on, if any? (ie. k8s cluster created before k8s resources are created)
	- What are the inputs values needed for this stage? (defined by it's own `stage_schema`)
	- What are the other required inputs for this stage, if any? (defined in the `nebari-config.yaml`, in sections other than the one defined by it's own `stage_schema`).
- How pluggable subcommands are defined, namely:
	- What `NebariPluginManager` attributes does this subcommand require or touch, if any?

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?


src/nebari/plugins.py
```diff
@@ -18,38 +28,243 @@
...
+                    stages.name,
+                )
+        return config_schema
+
+    def _version_callback(self, value: bool):
+        if value:
+            typer.echo(__version__)
+            raise typer.Exit()
+
+    def create_cli(self) -> typer.Typer:
```
costrouc, line 166: I'd prefer if the cli is not coupled to the plugin manager. But I think it makes perfect sense to have a PluginManager.

costrouc (changes requested):

costrouc:
Merging since we can continue from this point.

PR #1848, opened 2023-06-29
Refactor render functions, order stages based on explicit dependencies

iameskild:
## Reference Issues or PRs

This PR does two things (on second thought, this should've been two separate PRs):
- Rafactors render so that all of the functionality that was in render.py is now in `NebariPluginManager.reander_stages`
- We've had a lot of discussions around how to order stages, this PR adds a `dependencies` attribute which takes a list of explicit `OutputSchemas` from previous stages. 
  - The `get_available_stages` now orders the stages based on dependencies (stages with no dependencies always go first) and if two stages have the same dependencies, then the stages are ordered based on priority.
  - The goal would be to eventually remove `stage_outputs` and have the stage grab the values it needs from the `dependencies` schema instead. I didn't include this last bit in this PR as this PR is already fairly large.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?


src/_nebari/stages/nebari_tf_extensions/__init__.py
```diff
@@ -2,6 +2,9 @@
 from typing import Any, Dict, List
 
 from _nebari.stages.base import NebariTerraformStage
+from _nebari.stages.infrastructure import InfrastructureOutputSchema
```
costrouc, line 5: Is there a way we could make this system work without requiring imports? I'm still not sure if this is good/bad. I think having a discussion about this would give me a better idea on how this might work. I'm worried that plugins will have to import their dependencies in order to list one another as a dependency.

iameskild: I'm definitely open to discussing this in more detail. This was my attempt at making the dependencies more explicit. I understand that having explicit dependencies might place more restrictions on how stages are developed. 

That said, I would argue that we have implicit dependencies currently, namely `stage_outputs` and env vars set by the `contextmanager` of the previous sage. In stages developed outside of the main Nebari codebase, they might need to know which the exact values in `stage_outputs` (ie `stage_output['<stage_name>']['value']`).

costrouc:
We had a discussion around this today for about an hour [@aktech and @iameskild  and I]. We still think there needs to be discussion around this but in general agree this is the right direction and what the nebari stages system needs to adopt. 

Key issues:
  - how to replace a given stage?
  - how to extend the output of a given stage?
  - can a stage produce multiple outputs? (I think it should)

For a concrete example

```python
class Ingress(NebariStage):
    output_schema = OutputSchema


class MyIngress(Ingress):
    output_schema = OutputSchema


class MyStage(NebariStage):
    dependencies = [ OutputSchema ]
````

Which Stage does `MyStage` choose? How is this decided. 

To put this in another context it reminds me of interfaces. We want to choose the given stages which will give us the required interfaces.

iameskild:
Closing this PR in favor of documenting the preferred changes proposed.

Issue #1862, opened 2023-07-18
META - Extension System Testing


costrouc:
### Context

[#1833](https://github.com/nebari-dev/nebari/pull/1833) is a huge PR.  It need significant testing. Here is an exhaustive checklist 

### Value and/or benefit

Ensure this PR is successful and all previous behavior is preserved.

```[tasklist]
### Providers
- [ ] https://github.com/nebari-dev/nebari/issues/1951
- [ ] https://github.com/nebari-dev/nebari/issues/1952
- [ ] https://github.com/nebari-dev/nebari/issues/1953
- [ ] https://github.com/nebari-dev/nebari/issues/1954
```

```[tasklist]
### Extra
- [ ] https://github.com/nebari-dev/nebari/issues/1955
- [ ] https://github.com/nebari-dev/nebari/issues/1958
```

When testing out a checklist item follow https://github.com/nebari-dev/nebari/issues/new?assignees=&labels=type%3A+release+%F0%9F%8F%B7&template=testing-checklist.md&title=Testing+checklist+for+%3Cversion%3E.

costrouc:
Breaking up tasks:
  - stream for cli testing "framework"
  - stream for cloud testinging "framework"
  - schema fields all having descriptions and additional testing where needed. https://github.com/nebari-dev/nebari/pull/1833/files#diff-f7cda9f473c67dd06224a3ab445bb68065976dc6593511fab331179d60a50013R69 (stretch goal) not needed for this issue.

aktech:
One note the cleaning of resources, it doesn't cleans quite well by the way:

```
2023-08-01 13:27:23      INFO   49 utils: destroying Nebari took 3220.010 [s]
2023-08-01 13:27:23     ERROR  186 destroy: Stage=stages/07-kubernetes-services failed to fully destroy
2023-08-01 13:27:23     ERROR  186 destroy: Stage=stages/05-kubernetes-keycloak failed to fully destroy
2023-08-01 13:27:23     ERROR  186 destroy: Stage=stages/04-kubernetes-ingress failed to fully destroy
2023-08-01 13:27:23     ERROR  186 destroy: Stage=stages/03-kubernetes-initialize failed to fully destroy
2023-08-01 13:27:23     ERROR  186 destroy: Stage=stages/02-infrastructure failed to fully destroy
[...]

PR #1930, opened 2023-08-23
Handle region with care, updates to test suite

### LINKED ISSUE #1824: [ENH] - Allow setting region during init
### Feature description

We currently hard-code the following regions based on the cloud provider the user selects:

https://github.com/nebari-dev/nebari/blob/1413b9535be3458bac745f3014a886efbe16ab15/nebari/initialize.py#L217-L218

https://github.com/nebari-dev/nebari/blob/1413b9535be3458bac745f3014a886efbe16ab15/nebari/initialize.py#L228-L230

https://github.com/nebari-dev/nebari/blob/1413b9535be3458bac745f3014a886efbe16ab15/nebari/initialize.py#L239-L240

https://github.com/nebari-dev/nebari/blob/1413b9535be3458bac745f3014a886efbe16ab15/nebari/initialize.py#L260-L261

However, similar to the deploy target, the region has no good default value. Worse, `nebari init` doesn't even give you the option to set the region and one has to manually edit the yaml afterwards.

1. We should expose a `--region` option on `nebari init`
2. The option should be required in case a new cloud provider, i.e. not `"local"` or `"existing"` is used as deploy target.

For the guided init, we should add a question for the region (preferably a populated list) that will be displayed if the user selects one of the cloud providers.

### Value and/or benefit

Less frustration due to failed deploys and time needed manually editing the config yaml.

### Anything else?

_No response_

### LINKED ISSUE #1969: [BUG] - Azure redeployments fail if storage_account_postfix is not explicitly set
### Describe the bug

Initial Azure deployments are successful but the subsequent deployments fail because the Nebari schema generates a new postfix if the storage_account_postfix field is not set in the nebari-config.yaml; this was how it was prior to the extension-mechanism. 

The second deployment fails because a new random postfix is generated and this causes the Terraform state to delete the storage account and the state and then the deployment grinds to a halt.

### Expected behavior

To successfully redeploy Azure.

MacOS

Deploy Azure twice from the `@develop` branch.

### Command output

_No response_

_No response_

### Compute environment

Azure

### Integrations

_No response_

### Anything else?

_No response_

iameskild:
## Reference Issues or PRs

This PR does a few things:
- ensures that the `region` value provided (or defaults to a sensible value if not) can be validated against the cloud API
  - with the exception of Azure (for now)
- validates the kubernetes_version at `nebari init`
- updates the kubernetes version to 1.26
- adds Azure to the integration tests
  - I also replaced the `@on_cloud` decorator for now because it was parameterizing the session fixture which caused the deployment to attempt to deploy all of the clouds even if the `-m <cloud>` was not specified. My alternative for now is just to add the following flag `--cloud <cloud>` when you want to run the integration tests for that particular cloud.

Closes #1824
Closes #1969 
Related to #1862

- [ ] Documentation Update
- [ ] Build related changes
- [x] Other (please describe): kubernetes upgrade

## Testing

- [x] Did you add new tests?


src/_nebari/deploy.py
```diff
@@ -73,3 +73,5 @@ def deploy_configuration(
         print(
             "Additional administration docs can be found at https://docs.nebari.dev/en/stable/source/admin_guide/"
         )
+
+    return stage_outputs
```
Adam-D-Lewis, line 77: Why do we need to return stage_outputs now?

src/_nebari/initialize.py
```diff
@@ -87,39 +88,54 @@ def render_config(
...
+        config["digital_ocean"] = {
+            "kubernetes_version": kubernetes_version,
+            "region": region,
+        }
+
         config["theme"]["jupyterhub"][
             "hub_subtitle"
         ] = f"{WELCOME_HEADER_TEXT} on Digital Ocean"
-        if kubernetes_version is not None:
-            config["digital_ocean"] = {"kubernetes_version": kubernetes_version}
```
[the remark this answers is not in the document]
Adam-D-Lewis: nevermind, my mistake

src/_nebari/initialize.py
```diff
@@ -87,39 +88,54 @@ def render_config(
...
+            "kubernetes_version": kubernetes_version,
+            "region": region,
+        }
+
         config["theme"]["jupyterhub"][
             "hub_subtitle"
         ] = f"{WELCOME_HEADER_TEXT} on Google Cloud Platform"
-        config["google_cloud_platform"] = {}
         if "PROJECT_ID" in os.environ:
             config["google_cloud_platform"]["project"] = os.environ["PROJECT_ID"]
```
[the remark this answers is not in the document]
Adam-D-Lewis: never mind, my mistake

src/_nebari/deploy.py
iameskild: For testing, it makes it easier to test the state of the deployment if we have access to this stage_outputs. See deployment_fixtures.py line 201.

iameskild:
With this PR, running these integration tests locally work for Azure and AWS. I will work on GCP and making sure these can run in CI in another PR. I will also circle back to Digital Ocean once I can resolve #1965.

src/_nebari/initialize.py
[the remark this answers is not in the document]
iameskild: When `kubernetes_version` is passed in as `None`, this tells us that we should default to the latest version we support. 

For this particular module, I wanted to ensure the kubernetes version and region are properly set before returning a config, especially since they are needed for the subsequent deployment.

src/_nebari/initialize.py
```diff
@@ -87,39 +95,74 @@ def render_config(
...
-            config["google_cloud_platform"]["kubernetes_version"] = kubernetes_version
     elif cloud_provider == ProviderEnum.azure:
+        azure_region = region or constants.AZURE_DEFAULT_REGION
+        azure_kubernetes_version = kubernetes_version or get_latest_kubernetes_version(
+            azure_cloud.kubernetes_versions(azure_region)
+        )
+        config["azure"] = {
+            "kubernetes_version": azure_kubernetes_version,
+            "region": azure_region,
+            "storage_account_postfix": random_secure_string(length=4),
```
iameskild, line 139: This resolves #1969.

kenafoster:
Testing the init with region flags and commercial vs. GovCloud partitions:

Works OK when specified region is in the used credentials' partition

Using GovCloud credentials, trying the default (commercial us-east-1) region:

```
nebari init aws -p testproj
Defaulting to `us-east-1` region.
Please double-check that the AWS credentials are valid and have the correct permissions. If you're deploying into a non-standard partition (e.g. AWS GovCloud), please ensure the region specified exists in that partition
```

Similar (expected) message when passing in another commercial region with region flag.

I tested the converse (using commercial credentials and trying GovCloud region) and got equivalent errors.

Render and deploy appear to use the following order of priority when choosing the region.  If this is the intended behavior then all looks good to me.

1. Nebari config file
2. AWS_DEFAULT_REGION env var
3. .aws/config
4. Default Nebari region for AWS (in constants.py)

src/_nebari/render.py
```diff
@@ -179,67 +177,3 @@ def hash_file(file_path: str):
...
-def get_nebari_secret(secret_var):
-    env_var = secret_var.lstrip("NEBARI_SECRET_")
-    val = os.environ.get(env_var)
-    if not val:
-        raise EnvironmentError(
-            f"Since '{secret_var}' was found in the"
-            " Nebari config, the environment variable"
-            f" '{env_var}' must be set."
-        )
-    return val
```
iameskild, line 245: None of these functions where being used anywhere. And besides, a more succinct version of this lives in `config.py::set_config_from_environment_variables`.

iameskild:
Thanks for the feedback @kenafoster! I went ahead and made region and kubernetes_version required values for all cloud providers. This means that they need to be set in the nebari-config.yaml and after which, they become the source of truth for any future commands.

As for init, both region and kubernetes_version are set according to the following order:
1. --region flag in the init
2. Env vars (only AWS currently)
3. default to value set in constants.py

This feels like the right approach but it required updating some of the tests to reflect this new standard.

kenafoster:
Test results look good to me.  One minor concern with the content of the error in the scenario where you try 'nebari deploy' without aws.region specified in config.

Note for all of the tests below, my enviornment has a valid key set for AWS GovCloud partition.  I have also tested the converse (using a Commerical partition credential and trying to deploy into GovCloud regions) and gotten similar, expected results.

init

  - No AWS_DEFAULT_REGION set and no --region config
    - OK.  Nebari does not init (message "Please double-check that the AWS credentials are valid and have the correct permissions. If you're deploying into a non-standard partition (e.g. AWS GovCloud), please ensure the region specified exists in that partition.")
  - AWS_DEFAULT_REGION=us-gov-east-1 and no --region config
    - OK.  Nebari inits in us-gov-east-1 (message "Falling back to the region found in the AWS_DEFAULT_REGION environment variable: `us-gov-east-1`")
  - AWS_DEFAULT_REGION=us-gov-east-1 and --region us-gov-west-1
    - OK.  Nebari inits in us-gov-west-1 (**--region flag overrides AWS_DEFAULT_REGION and the Nebari default**)
  - AWS_DEFAULT_REGION=us-gov-east-1 and --region us-gov-east-999 (doesn't exist)
    - OK.  Nebari does not init (Please double-check that the region specified is valid. Could not connect to the endpoint URL: "https://ec2.us-gov-east-999.amazonaws.com/"


render/deploy

- No AWS region set in nebari-config.yaml
  - PARTIAL OK.  Behavior is good - it fails regardless of whether AWS_DEFAULT_REGION is configured in env.  However, message includes full traceback plus (ValidationError: 5 validation errors for ConfigSchema amazon_web_services -> region  field required (type=value_error.missing))
  - AWS GovCloud region set in nebari-config.yaml
    - OK (renders/deploys)
  - AWS Commercial region set in nebari-config.yaml
    - OK.  Does not deploy (message: Please double-check that the AWS credentials are valid and have the correct permissions. If you're deploying into a non-standard partition (e.g. AWS GovCloud), please ensure the region specified exists in that partition.)


Also confirmed render, deploy do not take --region option

iameskild:
Thanks a lot for the detailed testing @kenafoster! Improving the traceback and error messages is definitely worth a look. Thanks to @sblair-metrostar, we're tracking this [here](https://github.com/nebari-dev/nebari/issues/1946).

@pavithraes @kcpevey when you have time, could you please review and approve this PR? Let me know if you have any quesitons or concerns. Thank you :)

pavithraes (approved):
@iameskild Thanks for working on this, I'll approve but note the small merge conflict that needs resolution. :)

Issue #1800, opened 2023-05-10
[ENH] - Add a static type checker?


pmeier:
### Feature description

Add a static type checker like [`mypy`](https://mypy-lang.org/) to our lint workflows. Note that this doesn't mean that we have to declare support for [PEP561](https://peps.python.org/pep-0561/). I mostly want to make sure that the type hints that we have in our source are correct.

### Value and/or benefit

For me personally the following holds:

correct type hints > no type hints > wrong type hints

I'm aware that type hints are not everyone's cup of tea and they prefer no type hints, but I don't think anyone can argue that wrong type hints aren't detrimental for everyone. 

We currently have type hints in a few places, but no way to enforce their correctness. One example of wrong type hints is

https://github.com/nebari-dev/nebari/blob/8bd56f90b84c7686f45459ae8f7cf690721bd60a/nebari/render.py#L229-L231

https://github.com/nebari-dev/nebari/blob/8bd56f90b84c7686f45459ae8f7cf690721bd60a/nebari/render.py#L77-L81

The type hint has to be `List[str]` in both cases.

Thus, if we want to use type hints, and I'm all for that, we should also have a checker that enforces their correctness.

### Anything else?

If we decide that we don't want to have a type checker, we should also purge all type hints from the repository and enforce that no new ones are added. Sounds harsh, but I don't think there is another option to avoid wrong ones.

iameskild:
This was discussed during today's community call and we decided that we would incrementally move towards fixing all of the type annotations. We will add mypy to CI but for the time being ignore failures. @fangchenli has already started on this effort and has raised several PRs 🙌 

cc @pavithraes @dcmcand @fangchenli

PR #1974, opened 2023-09-03
TYP: fix typing error in hookspecs, refactor NebariStage

fangchenli:
## Reference Issues or PRs
xref #1800 ##1894

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?


src/nebari/hookspecs.py
```diff
@@ -1,54 +1,37 @@
...
-        self, stage_outputs: Dict[str, Dict[str, Any]], status: Dict[str, bool]
-    ):
-        yield
+    @abstractmethod
+    def render(self):
+        ...
 
 
 @hookspec
-def nebari_stage() -> List[NebariStage]:
```
fangchenli, line 48: Empty method body causes missing return error.

src/nebari/hookspecs.py
```diff
@@ -1,54 +1,37 @@
...
-    def check(
-        self, stage_outputs: Dict[str, Dict[str, Any]], disable_prompt: bool = False
-    ) -> bool:
-        pass
-
-    @contextlib.contextmanager
-    def destroy(
-        self, stage_outputs: Dict[str, Dict[str, Any]], status: Dict[str, bool]
-    ):
-        yield
```
fangchenli, line 44: Those don't apply for BootstrapStage.

src/nebari/hookspecs.py
```diff
@@ -1,54 +1,37 @@
...
-    input_schema: pydantic.BaseModel = None
-    output_schema: pydantic.BaseModel = None
-
-    def __init__(self, output_directory: pathlib.Path, config: schema.Main):
+    def __init__(self, output_directory: Path, config: schema.Main):
         self.output_directory = output_directory
         self.config = config
 
-    def render(self) -> Dict[str, str]:
-        return {}
```
fangchenli, line 27: The return type should be `Dict[Union[str, Path], str]`. Then we need to use the overload decorator in the derived class.

src/nebari/hookspecs.py
iameskild: Removing these hookspec methods effectively removes these as pluggable methods for downstream extensions.

These methods are not used by the BootstrapStage but are used by NebariTerraformStage. 

Perhaps there is more work needed before we can tear these methods out. Either adding them as hookspecs in another class or something else.

src/nebari/hookspecs.py
iameskild: Is this something you'll handle in this PR?

iameskild (changes requested):
Thanks @fangchenli! I left a few comments that I think we should discuss a bit further.

src/nebari/hookspecs.py
fangchenli: There is no elegant way to do this currently. I think it's better to left those abstract method without return time.

fangchenli: The return type should actually be `Dict[Path, str]`.

iameskild:
@costrouc could you have a look at this PR?

viniciusdc:
This has some big refactorings in the Nebari Stage which needs to be retested now that we changed a bit the behavior for the default schema dumps -- I will follow up on this, next week

dcmcand:
@viniciusdc should we close this one or is there still value here?

viniciusdc:
Hi @dcmcand, There is a lot of value here. This is good work that we will need to encompass sooner or later (related to that schema refactoring I was talking about recently). I was waiting on this due to the exposing default config values Adam was working on.

My opinion on this would be to close, and I will incorporate and cite those changes in a follow up PR

Issue #1958, opened 2023-08-28
[META] - Nebari testing suite enhancements


iameskild:
```[tasklist]
### Unit tests
- [ ] Create a fixture to generate testable nebari-config.yaml (see this comment)  
```

```[tasklist]
### Playwright integration tests
- [ ] dask
- [ ] conda-store
- [ ] jupyter-scheduler
- [ ] code-server (vs code in the browser)
- [ ] argo-workflows
```

PR #1996, opened 2023-09-11
TST: enable timeout config in playwright notebook test

fangchenli:
## Reference Issues or PRs
xref #1971 #1958 

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

## Any other comments?
The `timeout` or (`runtime`) parameter was not passed to the inner functions, so the playwright test was always using the default value (120 seconds).


iameskild:
Thanks for working on this @fangchenli! I know this test can be flaky at times but I think I would prefer increasing the timeout instead of labeling this as xfail.

iameskild (approved):
Thanks @fangchenli! Can you update the branch and when CI passes, feel free to merge!

iameskild (commented):
This looks great @fangchenli! I think eventually we want to move these playwright tests to the `tests_integration` directory so we can retire Cypress . How much trouble would it be to make that update?

fangchenli:
@iameskild Thanks for reviewing this PR!

PR #2008, opened 2023-09-13
ENH: specify required fields when retrieving available gcp projects

fangchenli:

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

### DIFF
--- a/src/_nebari/provider/cloud/google_cloud.py
+++ b/src/_nebari/provider/cloud/google_cloud.py
@@ -21,8 +21,10 @@ def check_credentials():
 def projects() -> Dict[str, str]:
     """Return a dict of available projects."""
     check_credentials()
-    output = subprocess.check_output(["gcloud", "projects", "list", "--format=json"])
-    data = json.loads(output.decode("utf-8"))
+    output = subprocess.check_output(
+        ["gcloud", "projects", "list", "--format=json(name,projectId)"]
+    )
+    data = json.loads(output)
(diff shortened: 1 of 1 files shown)


src/_nebari/provider/cloud/google_cloud.py
```diff
@@ -21,8 +21,10 @@ def check_credentials():
 def projects() -> Dict[str, str]:
     """Return a dict of available projects."""
     check_credentials()
-    output = subprocess.check_output(["gcloud", "projects", "list", "--format=json"])
-    data = json.loads(output.decode("utf-8"))
+    output = subprocess.check_output(
+        ["gcloud", "projects", "list", "--format=json(name,projectId)"]
```
iameskild, line 25: I needed to add quotes like this: `--format="json(name, projectId)"`. I believe this is due to a difference between MacOS and Linux. Is this something you can confirm @fangchenli?

iameskild (approved):
I had one small question but otherwise, LGTM.

src/_nebari/provider/cloud/google_cloud.py
fangchenli: > I needed to add quotes like this: `--format="json(name, projectId)"`. I believe this is due to a difference between MacOS and Linux. Is this something you can confirm @fangchenli?

Yes, quotes are required when running it in terminal on Linux. But when calling from python, `"--format=json(name,projectId)"` works on both MacOS and Linux.

PR #2102, opened 2023-11-09
Fix integration tests, helm-validate script

iameskild:

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?


.github/workflows/test_helm_charts.yaml
```diff
@@ -7,9 +7,10 @@ on:
   schedule:
     # Run every Monday at 13:00 UTC
     - cron: "0 13 * * 1"
-  push:
+  pull_request:
```
dcmcand, line 10: The comment above is now incorrect, it only runs on PR's, not when the file is changed. That being said, I am curious about that change. It seems useful to have this check run if the files are modified.

dcmcand (changes requested):

.github/workflows/test_helm_charts.yaml
iameskild: I'm happy to have it run on `push` as well, I just think it's most important to run on `pull_request` to confirm the changes are valid. And then once it's merged, it will run once a week anyway.

dcmcand: Yeah, I agree it is import on `pull_request`, but since it only runs if those files are changed, I think it would be good feedback to have it run on push too. However, if you don't think it needs to run on push, that is ok with me, the comment on line 2 just needs updated to reflect when it will actually run accurately.

iameskild: It will run on `push` and `pull_request` 👍

dcmcand (approved):

Issue #2164, opened 2023-12-22
[BUG] - DNS Lookup while deploying often takes time and fails to match LB ip to Domain.


pt247:
### Describe the bug

While deploying on AWS, the nebari deployment tries to match the IP of the AWS load balancer and the host. The AWS on the other hand gives 2 IPs for the load balancer that it keeps on rotating. For example: 
```
(base)  ➜ nebari-aws (2154-aws-set-minimum-notes-to-0) ✏️  dig a2c6987d8de97475491b592d93537928-2145647921.eu-west-1.elb.amazonaws.com.

; <<>> DiG 9.10.6 <<>> a2c6987d8de97475491b592d93537928-2145647921.eu-west-1.elb.amazonaws.com.
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 3748
;; flags: qr rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;a2c6987d8de97475491b592d93537928-2145647921.eu-west-1.elb.amazonaws.com. IN A

;; ANSWER SECTION:
a2c6987d8de97475491b592d93537928-2145647921.eu-west-1.elb.amazonaws.com. 60 IN A 34.252.157.110
a2c6987d8de97475491b592d93537928-2145647921.eu-west-1.elb.amazonaws.com. 60 IN A 52.214.38.217

;; Query time: 39 msec
;; SERVER: 192.168.1.1#53(192.168.1.1)
;; WHEN: Fri Dec 22 12:31:55 GMT 2023
;; MSG SIZE  rcvd: 132

(base)  ➜ nebari-aws (2154-aws-set-minimum-notes-to-0) ✏️  

```
The above output shows that this domain name resolves to 2 IPs [34.252.157.110, 52.214.38.217].
Below is another dig this time on the hostname it tries to match this with:
```
(base)  ➜ nebari-aws (2154-aws-set-minimum-notes-to-0) ✏️  dig nebari.prashanttiwari.com

; <<>> DiG 9.10.6 <<>> nebari.prashanttiwari.com
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 5851
[...]

PR #2166, opened 2023-12-22
Fix logic for dns lookup.

pt247:
## Reference Issues or PRs
#2164 

Fix logic for matching the domain IP. 

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

### DIFF
--- a/src/_nebari/stages/kubernetes_ingress/__init__.py
+++ b/src/_nebari/stages/kubernetes_ingress/__init__.py
@@ -67,23 +67,23 @@ def check_ingress_dns(stage_outputs: Dict[str, Dict[str, Any]], disable_prompt: 
     directory = "stages/04-kubernetes-ingress"
 
     ip_or_name = stage_outputs[directory]["load_balancer_address"]["value"]
-    ip = socket.gethostbyname(ip_or_name["hostname"] or ip_or_name["ip"])
+    _, _, ips = socket.gethostbyname_ex(ip_or_name["hostname"] or ip_or_name["ip"])
     domain_name = stage_outputs[directory]["domain"]
 
     def _attempt_dns_lookup(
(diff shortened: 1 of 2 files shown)


src/_nebari/stages/kubernetes_ingress/__init__.py
```diff
@@ -67,23 +67,23 @@ def check_ingress_dns(stage_outputs: Dict[str, Dict[str, Any]], disable_prompt:
...
     def _attempt_dns_lookup(
-        domain_name, ip, num_attempts=NUM_ATTEMPTS, timeout=TIMEOUT
+        domain_name, ips, num_attempts=NUM_ATTEMPTS, timeout=TIMEOUT
     ):
         for i in range(num_attempts):
             try:
-                resolved_ip = socket.gethostbyname(domain_name)
-                if resolved_ip == ip:
+                _, _, resolved_ips = socket.gethostbyname_ex(domain_name)
+                if set(resolved_ips) == set(ips):
```
dcmcand, line 79: I think that I actually liked your original implementation better here. Even though a CNAME is better if your LB has multiple ips, by checking if the resolved ip is in the list of ips it leaves the possibility that someone could just do an A record for a single ip.

If we are going to use this approach where we check all the ips, then the error message recommend a CNAME in the case of multiple IPs.

dcmcand (changes requested):

src/_nebari/stages/kubernetes_ingress/__init__.py
pt247: Cool. I will change it back.

pt247: Done

dcmcand (approved):
@pt247 thanks for the PR. :rocket:

Issue #2154, opened 2023-12-18
[ENH] - Set minimum nodes to 0 for AWS deployment


aktech:
### Feature description

We need to set the `min_nodes` for AWS to **0** for `user` and `worker` nodes. We do have [0 for GCP](https://github.com/nebari-dev/nebari/blob/efc291339c113cd244eed4778272e77d8026b0c0/src/_nebari/stages/infrastructure/__init__.py#L329).

Otherwise this makes Nebari quite expensive (**~$625/month**) for someone trying it with default configuration:

```yaml
amazon_web_services:
  kubernetes_version: '1.26'
  region: us-east-1
```

Cost:

- general node: m5.2xlarge at $0.384/hr: 0.384*720 = $276.48
- user and worker node: 2 x m5.xlarge at $0.192/hr: 2*0.192*720 = $276.48
- Node cost $552.96
- K8s base cost ~ $72

Nebari base cost on AWS: **$625** per month with default config

This was not supported in by AWS, but its been a while [since it's supported](https://aws.amazon.com/blogs/containers/amazon-eks-now-supports-kubernetes-version-1-24/).

### Value and/or benefit

Reduction in base cost for Nebari on AWS with the default configuration.

pt247:
To replicate this, I created a Nebari cluster. The following nodes were created:
```
┌───────────────────────────────────────────────────────── Nodes(all)[1] ──────────────────────────────────────────────────────────┐
│ NAME↑                                           STATUS     ROLE        TAINTS     VERSION                       PODS AGE         │
│ ip-10-10-41-202.eu-west-1.compute.internal      Ready      <none>      0          v1.26.10-eks-e71965b             0 132m        │

└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

Default settings create a single node. All the pods are deployed on the same node. 
Do we still need this change? Am I looking in the right place?

pt247:
Thanks @aktech for pointing out that k9 is not actually showing all the hosts. On AWS EC2 console we see 3 instances.
 
<img width="1324" alt="Screenshot 2023-12-22 at 13 31 03" src="https://github.com/nebari-dev/nebari/assets/8033215/41c84d11-7b04-4a46-ae79-f342d7172047">

After setting the worker and user nodes to 0 and destroying and recreating, I can verify that there is only one general instance.
[...]

PR #2168, opened 2023-12-22
Set min nodes to 0 for worker and user.

pt247:
## Reference Issues or PRs
#2154

Reduces cost of default deployment. 

### Change
An AWS Node group can be associated with several Auto Scaling groups, but in our case, it's just one. To scale from zero in AWS, you need to follow these steps:

1. Define the service node selector, such as `dedicated=user`.
2. Ensure that the nodes groups have a corresponding label like `dedicated=user`. This will help the service decide which node group to place the pod on.
3. ASG is required to give cluster-autoscaler a hint like `k8s.io/cluster-autoscaler/node-template/label/dedicated=user`. 

- [ ] Documentation Update
- [x] Build related changes
- [x] Other (please describe):
        As a part of this change, the user scheduler is moved to the General node. This is needed to ensure the user node scales down to zero when the user is not running notebooks or desk clusters. 

## Testing

- [ ] Did you add new tests?


pt247:
I have now tested scaling up and down of user and worker nodes from and to 0. I did this by creating the following:
1. Creating a JupyterHub cluster for the user.
2. Creating a Dask-cluster for the user.
3. Deleting Dask-cluster.
4. Release the node from the Jupyter node hub. 

I had to move the user scheduler to general, though. This was needed because the user scheduler on the user node will always trigger the user node scale-up.

pt247:
Just a couple of changes are needed. 
1. Moving tagging of ASG to a later stage. As it needs to exist before it can be tagged. 
2. Revert changes to retain original node selectors.

pt247:
I need to figure out why Local Integration Tests are falling:
```
[terraform]: │ Error: failed to refresh cached credentials, no EC2 IMDS role found,
[terraform]: │ operation error ec2imds: GetMetadata, http response error StatusCode: 404,
[terraform]: │ request to EC2 IMDS failed
```
I'm still trying to figure out what the root cause is. If anyone has suggestions, please let me know. 
I will work on this later today.

costrouc:
@pt247 first off want to say thank you for the PR! We have been wanting to have autoscaling down to zero for a long time on AWS and this will have significant cost savings for the nebari deployment.

My key concern with this PR has to do with this triggering a destroy on preexisting aws deployments. This isn't anything wrong with your PR but I want to make sure Nebari has an answer for how to handle changes which will trigger new node groups to be created. If this deletes pvc's I would be concerned about this PR causing data to be deleted on upgrade.

pt247:
@costrouc  I have checked this today. I installed the develop branch first and then the node-0 branch. As I suspected, the nodes didn't scale back because of protection already in place at the `develop` branch.  [link to ignore changes to `desired_size`](https://github.com/nebari-dev/nebari/blob/e3e06a8ae296ba56fd510006f0dad932069dad2f/src/_nebari/stages/infrastructure/template/aws/modules/kubernetes/main.tf#L42C1-L42C1).

pt247:
### Update
After moving the tagging logic to the post_deploy stage of Phase 2, we were able to successfully get past the `Deploy Nebari` phase in the `test-local-integration`. However, we are now encountering a failure at the `Cypress Run` stage. Specifically, the failure is occurring during the `First Test -- Check Nebari login and start JupyterLab`. Upon reviewing the saved video, it appears that the login is functioning properly, but the JupyterLab server is not starting.

### Next steps
I have set up an EC2 instance and installed the local cluster. I will attempt to log in to it by exploring the display through SSH.

Please let me know if there is a better way.

pt247:
I have modified the code to make sure that only changes for AWS don't affect other deployments. With this, the tests are working as expected.

costrouc (approved):
Looks great to me!

pt247:
All checks pass(ed). If you are happy with the changes, please feel free to merge them @costrouc

PR #2192, opened 2024-01-10
Disable "Newer update available" popup as default setting

### LINKED ISSUE #2190: [ENH] - Remove "Newer version available" popup message
### Feature description

Originally, we've addressed the news's popup segment with the PR #1641, but the recent updates on jupyterhub/jupyterlab introduced a new notification type for the available updates: 

image

according to their codebase 

https://github.com/jupyterlab/jupyterlab/blob/13ba4cc6dad700b5df13beb8c39fb0b90bb6b1ea/packages/apputils-extension/src/announcements.ts#L222-L229

We can disable it by passing into the jupyterlab settings (json) the  `checkForUpdates: false`  like so:

``` 
 "@jupyterlab/apputils-extension:notification": {
            "fetchNews": "false",
            "checkForUpdates": "false"
        }
```

### Value and/or benefit

This will help avoid confusion by the users of the platform, who could think their environment needs some time of update or could be in a broken state.

### Anything else?

_No response_

viniciusdc:
## Reference Issues or PRs

closes #2190 

This removes the popup from showing up when the user spawns its JupyterLab instance, more details in the corresponding issue.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

## Any other comments?

Tested on a local version deployment of Nebari (based on Dev)

### DIFF
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterlab/overrides.json
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterlab/overrides.json
@@ -9,8 +9,7 @@
         "authToken": ""
     },
     "@jupyterlab/apputils-extension:notification": {
-        "fetchNews": false,
-        "checkForUpdates": "false"
+        "checkForUpdates": false,
         "fetchNews": "false"
     },
     "@jupyterlab/mainmenu-extension:plugin": {
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterlab/overrides.json
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterlab/overrides.json
@@ -9,6 +9,7 @@
         "authToken": ""
     },
     "@jupyterlab/apputils-extension:notification": {
-        "fetchNews": "false"
+        "fetchNews": "false",
+        "checkForUpdates": "false"
     }
 }


src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterlab/overrides.json
```diff
@@ -9,6 +9,7 @@
         "authToken": ""
     },
     "@jupyterlab/apputils-extension:notification": {
-        "fetchNews": "false"
+        "fetchNews": "false",
+        "checkForUpdates": "false"
```
marcelovilla, line 13: @viniciusdc shouldn't this be a boolean value (i.e., `false`) rather than a string? Looking at the default values, it is indeed a boolean. I mean, https://github.com/jupyterlab/jupyterlab/blob/13ba4cc6dad700b5df13beb8c39fb0b90bb6b1ea/packages/apputils-extension/src/announcements.ts#L222 should convert it to boolean anyways, but somehow it just feels wrong.

Makes me wonder whether `"fetchNews"` should also have a boolean value.

marcelovilla: Apparently, `"fetchNews"` does not let you pass a boolean value. However, `"checkForUpdates"` complains if you pass a string:

image

So much for consistency!

marcelovilla (changes requested):

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterlab/overrides.json
marcelovilla: @viniciusdc I'm changing this and will merge

marcelovilla (approved):

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterlab/overrides.json
viniciusdc: thanks, @marcelovilla. indeed, I went with string because of ` fetchNews.` Also, good catch. I remember spotting the error message above and manually updating the pod, but I didn't commit the change to Bolean. Thanks

PR #2200, opened 2024-01-11
Adding fields to ignore within keycloak_realm

### LINKED ISSUE #2180: [BUG] - Prevent SMTP realm settings from being wiped by terraform
### Describe the bug

Currently it is not possible to set keycloak realm settings without a redeploy wiping settings.

### Expected behavior

Don't wipe keycloak user set realm settings

linux

Redeploy nebari

### Command output

_No response_

_No response_

### Compute environment

None

### Integrations

_No response_

### Anything else?

_No response_

costrouc:
Closes #2180

This PR uses the lifecycle.ignore_changes terraform attrribute which allows setting terraform to ignore if certain attributes are modified. I have tested this locally to work.

ignore_changes also supports an `all` parameter but this would set terraform to ignore all attributes in the resource which we do not want.

### DIFF
--- a/src/_nebari/stages/kubernetes_keycloak_configuration/template/main.tf
+++ b/src/_nebari/stages/kubernetes_keycloak_configuration/template/main.tf
@@ -63,9 +63,9 @@ resource "keycloak_realm" "main" {
       otp_policy,
       default_default_client_scopes,
       default_optional_client_scopes,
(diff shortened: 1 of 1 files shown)


kenafoster (approved):

dcmcand (approved):
:rocket:

Issue #2271, opened 2024-02-22
[BUG] -  EntityAlreadyExists  errors when upgrading from 2023.11.1 to 2024.1.1


rsignell:
### Describe the bug

I upgraded my nebari config file from 2023.11.1 to nebari 2024.1.1 using `nebari upgrade` which went smoothly. 

When I ran `nebari deploy`, however, I got these terraform errors  complaining that resources already exist:
```
[terraform]: │ Error: error creating IAM Role (osc-training-dev-eks-cluster-role): EntityAlreadyExists: Role with name osc-training-dev-eks-cluster-role already exists.
[terraform]: │  status code: 409, request id: 7f06518f-7857-497c-9e82-1e9530c2eb38
[terraform]: │
[terraform]: │   with module.kubernetes.aws_iam_role.cluster,
[terraform]: │   on modules/kubernetes/policy.tf line 10, in resource "aws_iam_role" "cluster":
[terraform]: │   10: resource "aws_iam_role" "cluster" {
[terraform]: │
[terraform]: ╵
[terraform]: ╷
[terraform]: │ Error: error creating IAM Role (osc-training-dev-eks-node-group-role): EntityAlreadyExists: Role with name osc-training-dev-eks-node-group-role already exists.
[terraform]: │  status code: 409, request id: dffd675a-6907-473f-a011-ff78d633ac13
[terraform]: │
[terraform]: │   with module.kubernetes.aws_iam_role.node-group,
[terraform]: │   on modules/kubernetes/policy.tf line 39, in resource "aws_iam_role" "node-group":
[terraform]: │   39: resource "aws_iam_role" "node-group" {
[terraform]: │
[terraform]: ╵
[terraform]: ╷
[terraform]: │ Error: error creating ECR repository: RepositoryAlreadyExistsException: The repository with name 'osc-training-dev-jupyterlab' already exists in the registry with id '153379367922'
[terraform]: │
[terraform]: │   with module.registry-jupyterlab.aws_ecr_repository.main,
[terraform]: │   on modules/registry/main.tf line 1, in resource "aws_ecr_repository" "main":
[terraform]: │    1: resource "aws_ecr_repository" "main" {
```

### Expected behavior

I would expect the deployment to work, or tell me what I need to do.

### OS and architecture in which you are running Nebari

Deploy command on WSL, deployment on AWS

### How to Reproduce the problem?

see above!

### Versions and dependencies used.

conda 23.3.1
Client Version: v1.28.3
[...]

PR #2275, opened 2024-02-22
Prepare 2024.3.1rc1 release

marcelovilla:

- [ ] Documentation Update
- [x] Build related changes

## Testing

- [ ] Did you add new tests?

## Any other comments?

We still have some blockers for the next release but I want to update the docker image tags in `develop` so people can deploy from there to test.


src/_nebari/constants.py
```diff
@@ -12,8 +12,8 @@
 DEFAULT_GKE_RELEASE_CHANNEL = "UNSPECIFIED"
 
 DEFAULT_NEBARI_DASK_VERSION = CURRENT_RELEASE
-DEFAULT_NEBARI_IMAGE_TAG = CURRENT_RELEASE
-DEFAULT_NEBARI_WORKFLOW_CONTROLLER_IMAGE_TAG = "2024.1.1"
+DEFAULT_NEBARI_IMAGE_TAG = "2024.2.1rc2"
```
marcelovilla, line 15: @dcmcand In https://github.com/nebari-dev/nebari/pull/2227 you did not have any changes for `DEFAULT_NEBARI_IMAGE_TAG` but I am assuming we do want to point to the latest RC in https://github.com/nebari-dev/nebari-docker-images, right? Otherwise we would be testing against older versions of the images.

dcmcand: I specified them in the nebari-config.yaml rather than here. Bu this should be fine as long as we change it back to CURRENT_RELEASE before the release

src/_nebari/upgrade.py
```diff
@@ -705,6 +705,20 @@ def _version_specific_upgrade(
...
 
+class Upgrade_2024_2_1(UpgradeStep):
+    version = "2024.2.1"
+
+    def _version_specific_upgrade(
+        self, config, start_version, config_filename: Path, *args, **kwargs
+    ):
+        rich.print("\n ⚠️  Warning ⚠️")
+        rich.print(
+            "-> Please run the [green]rm -rf stages[/green] so that we can regenerate an updated set of Terraform scripts for your deployment."
```
dcmcand, line 716: This can cause trouble for people using local tf state. See https://github.com/nebari-dev/nebari/issues/2271. I think just runnning `nebari render` should do the same thing here. Can we verify this is actually needed?

Adam-D-Lewis: At a minimum I think we should change this to `mv stages stages.old` or something similar.  This messed me up recently.  Luckily I was able to get stages back by trash-cli

marcelovilla:
@Adam-D-Lewis and @dcmcand thank you for the comments. I want us to discuss better ways to handle upgrades in the future, but for this release I have followed your advice and just added a message to backup the original rendered files.

src/_nebari/upgrade.py
```diff
@@ -705,6 +705,20 @@ def _version_specific_upgrade(
...
 
+class Upgrade_2024_3_1(UpgradeStep):
+    version = "2024.3.1"
+
+    def _version_specific_upgrade(
+        self, config, start_version, config_filename: Path, *args, **kwargs
+    ):
+        rich.print("\n ⚠️  Warning ⚠️")
+        rich.print(
+            "-> Please run [green]mv stages stages.old[/green] so that we can keep a backup of the original Terraform scripts and regenerate an updated set of them for your deployment."
```
dcmcand, line 716: Take this comment or leave it, but I don't feel like even this is necessary for this release. I feel like this could be as simple as printing "upgrading to 2024.2.1" or something like that. I am still approving in case you decide to keep this, but I don't think we need it.

dcmcand (approved):

src/_nebari/upgrade.py
marcelovilla: It's probably not because we're not deprecating anything. I took your suggestion and incorporated it.

Issue #2122, opened 2023-11-18
[BUG] - Config generated by guided init should contain default settings


rsignell:
### Describe the bug

We generated a config using guided-init , and deployed Nebari on AWS with no apparent issues. 

We then tried to use the deployment for training and students were not able to get the 30 workers they requested.  We double checked our service quota in the region we were running the training and it was set to 500.   
So we were mystified (and frustrated) why things weren't working. 

Then after the training we realized that the generated config didn't contain the node groups section with the max limits for the different worker pools.   In other words, the amazon section just looked like:
```
amazon_web_services:
  kubernetes_version: '1.26'
  region: us-west-2
```
So I imagine we were just getting the default max, which must have been something small. 

When I added the section to:
```
amazon_web_services:
  kubernetes_version: '1.26'
  region: us-west-2
  node_groups:
    general:
      instance: m5.2xlarge
      min_nodes: 1
      max_nodes: 1
    user:
      instance: m5.xlarge
      min_nodes: 1
      max_nodes: 100
    worker:
      instance: m5.xlarge
      min_nodes: 1
      max_nodes: 450
```
 and rerendered and redeployed, multiple students could spin up clusters of 30 again. 

### Expected behavior

The expected behaviour would be to have the guided-init create a config that contains the default settings so they could (1) show the user what the defaults were, and (2) provide a template that facilitates easy modification by the user. 

### OS and architecture in which you are running Nebari

aws

### How to Reproduce the problem?

use the guided init on aws

### Versions and dependencies used.

nebari 2023.10.1

### Compute environment

AWS

aktech:
Yes, indeed. This would be super helpful.

rsignell:
@aktech and I confirmed that his public app worked, but my app (which worked fine when not public) did not work as public.
The logs seemly showed everything working okay until it suddenly closed the stream and the pod died:
```
[...]

PR #2294, opened 2024-03-06
Explicit config

Adam-D-Lewis:
## Reference Issues or PRs
Fixes https://github.com/nebari-dev/nebari/issues/2122,  based on https://github.com/nebari-dev/nebari/pull/2348 which should be merged first

Currently, we set all the nebari-config fields manually in https://github.com/nebari-dev/nebari/blob/0210a47a1acbe4940e0f6ff18fb72f630e224230/src/_nebari/initialize.py#L33

but it's confusing for users to not see some useful parts of the config (e.g. https://github.com/nebari-dev/nebari/issues/2122).  We could add more fields to render_config method, but every time we add a new field in any InputSchema, we'd have to update the render_config method which is inconvenient.  Additionally, having the fields hard coded in render_config means Nebari extensions don't have their InputSchema's written to the config file even if installed.  They would need to be added some other way (e.g. typing them in manually) in order to control the extension's settings exposed by it's InputSchema.

This PR instead writes all of the InputSchema's to the nebari-config file by default, but this led to some confusing config in cases like infrastructure InputSchema where this PR caused a section to be added for each of "local", "existing", "google_cloud_platform", etc. for each cloud prodiver and all except one of these provider sections was blank.
 https://github.com/nebari-dev/nebari/blob/0210a47a1acbe4940e0f6ff18fb72f630e224230/src/_nebari/stages/infrastructure/__init__.py#L540-L546
 
So this PR also allows InputSchemas to define a `exclude_from_config` method which returns the names of the fields to exclude from the config file to fix this issue.  We likely could address this issue by separating the different providers into separated Pydantic Models (similar to what I've done to the certificate Pydantic models), but I feel it's important to have some mechanism to manually exclude portions of the InputSchema from being written to the nebari-config file when running `nebari init`, and have to chosen to address the extra providers with `exclude_from_config`.

- [ ] Documentation Update
- [ ] Build related changes
- [x] Other (please describe):  `nebari init` results in a more complete nebari-config.yaml

## Testing

- [ ] Did you add new tests?


src/_nebari/initialize.py
```diff
@@ -18,7 +18,8 @@
 )
 from _nebari.provider.oauth.auth0 import create_client
 from _nebari.stages.bootstrap import CiEnum
-from _nebari.stages.kubernetes_ingress import CertificateEnum
+
```
Adam-D-Lewis, line 21: need to get rid of the comments in this file

src/nebari/schema.py
```diff
@@ -28,6 +28,7 @@ class Config:
         extra = "forbid"
         validate_assignment = True
         allow_population_by_field_name = True
+        smart_union = True
```
Adam-D-Lewis, line 31: This is the default behavior in Pydantic V2 and prevents some unintuitive behavior.  This is necessary for the new cert classes when loadiing in the config file to make sure pydantic loads the correct class.

viniciusdc:
Thanks for the PR @Adam-D-Lewis ! I was wonderign about this whole schema and InputVar structure during the weekend as well. Will have a look once I have some time today

pavithraes:
Notes from `@pt247` and `@dcmcand` the community meeting -- it's important to have a clear separation between the values set by user and values set by default in Nebari. This can take the form of commenting out the default options.

Adam-D-Lewis:
Opened an RFD to address new requirements - https://github.com/nebari-dev/governance/issues/48

pyproject.toml
```diff
@@ -136,21 +136,23 @@ module = [
...
-    "PTH",
+    "E",  # E: pycodestyle rules
+    "F",  # F: pyflakes rules
+    "PTH",    # PTH: flake8-use-pathlib rules
 ]
 ignore = [
     "E501", # Line too long
     "F821", # Undefined name
     "PTH123", # open() should be replaced by Path.open()
 ]
```
Adam-D-Lewis, line 155: This fixes a deprecation warning from ruff.

Adam-D-Lewis:
Okay, I've really paired down this PR to just show all the config explicitly and nothing else.  It is going to be confusing for beginners, and in some spots even for more advanced users (the certificate section mashing all the different certificates together comes to mind).  

certificate:
  type: self-signed
  secret_name:
  acme_email:
  acme_server: https://acme-v02.api.letsencrypt.org/directory

We invite users to open issues where clarification is desired.

Adam-D-Lewis:
I'll rename verbose to explicit after talking with Marcelo

Adam-D-Lewis:
I'll check if there any applicable tests to add as well

src/_nebari/subcommands/init.py
```diff
@@ -565,6 +566,13 @@ def init(
             "-o",
             help="Output file path for the rendered config file.",
         ),
+        explicit: int = typer.Option(
```
marcelovilla, line 569: @Adam-D-Lewis is there any reason to have `int` over `bool` here? I know either of those will work but I'm wondering if there is any advantage.

marcelovilla (approved):
This is great @Adam-D-Lewis 🚀 ! Thank you!

Everything works as expected so I'm approving. Here are a couple comments for discussion but feel free to ignore them if you don't find them relevant.

- See comment regarding having the `explicit` option being an `int` instead of a `bool`.
- I noticed that on the guided init, the explicit config prompt is only shown when the users confirms they want to make advanced configuration changes. While I agree this might be an option for "advanced" users, I feel it might be a little bit hidden under that section. However, I don't have any strong feelings as one can always use the `--explicit` flag so I think if we document it well enough, it should not be a problem.

src/_nebari/subcommands/init.py
Adam-D-Lewis: Making it an int allows you to specify `--explicit --explicit` or `-ee`  to have levels of explicitness.  We aren't using the levels yet, but may in the future.

Adam-D-Lewis:
> I noticed that on the guided init, the explicit config prompt is only shown when the users confirms they want to make advanced configuration changes. While I agree this might be an option for "advanced" users, I feel it might be a little bit hidden under that section. However, I don't have any strong feelings as one can always use the --explicit flag so I think if we document it well enough, it should not be a problem.

I think it's going to be confusing for users until we document it better (as done [here](https://github.com/nebari-dev/nebari-docs/pull/453)) so I think it is an advanced configuration setting that should be hidden there for the moment.  Happy to move it once it's better documented.

Adam-D-Lewis:
Thanks for reviewing!

PR #2322, opened 2024-03-14
get default variable value when following a terraform variable

Adam-D-Lewis:
## Reference Issues or PRs
[This](https://github.com/nebari-dev/nebari/actions/runs/8272452121/job/22634293608) part of the release process was failing with error `ValueError: Could not find loki:{'description': 'version to deploy for the loki helm chart', 'type': '${string}', 'default': '5.43.3'} directory in helm_charts.`.  chart_version was being set to `{'description': 'version to deploy for the loki helm chart', 'type': '${string}', 'default': '5.43.3'}` instead of '5.43.3'.  This became an issue with the recent addition of loki.

I tested the change locally and it seems to run fine.  I'm not sure if there are other use cases where we would not want to grab the default value, but it does seem to run successfully when I run it locally.  Asking @viniciusdc to review since I believe he originally implemented this and may have more insight.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests? No

### DIFF
--- a/scripts/helm-validate.py
+++ b/scripts/helm-validate.py
@@ -67,7 +67,7 @@ class HelmChartIndexer:
             var_name = self._clean_var_name(argument, "var")
             for var in parent_contents.get("variable", {}):
                 if var_name in var:
-                    return var[var_name]
+                    return var[var_name]["default"]
             else:
                 raise ValueError(f"Could not find variable {var_name}")
 


viniciusdc (approved):
LGTM, I don't see any problems with using the variable's default value. The primary purpose of this was to catch any issues with the default versions and avoid situations where the helm is no longer available.

PR #2347, opened 2024-03-19
Deduplicate conda-store in JupyterLab main menu

### LINKED ISSUE #2326: [BUG] - Conda store shows up in two places in menu bar
### Describe the bug

In 2024.3.2 we have `jupyterlab-conda-store` extension back. However, now we have two (three) menu items allowing to open conda store:
- "Services > Environment Management" coming from nebari and opening conda-store in a new tab:
  image
- "Conda-Store > Conda Store Package Manager" coming from `jupyterlab-conda-store`:
   image

I don't think there is enough value in having the separate `Conda-Store` menu with just one item, and it might be confusing for users.

### Expected behavior

 Should we replace action associated with `Services > Environment Management` to open conda-store within JupyterLab? Or should we have both in a single menu (in that case how should we distinguish them?)

2024.3.2

Open JupyterLab

### Command output

_No response_

_No response_

### Compute environment

None

### Integrations

_No response_

### Anything else?

This is a follow-up to
- https://github.com/nebari-dev/nebari/issues/2182
- https://github.com/nebari-dev/nebari/issues/2216
- https://github.com/nebari-dev/nebari/issues/2302

krassowski:
## Reference Issues or PRs

- Fixes #2326.
- Depends on https://github.com/nebari-dev/nebari-docker-images/pull/130

Uses `jupyterlab-conda-store` if available, fallbacks to opening `/conda-store` in new browser tab otherwise. The label in each case is set to "Environment Manager" (but we can adjust it).

### Before

- "Services > Environment Management" coming from nebari and opening conda-store in a new tab:
  image
- "Conda-Store > Conda Store Package Manager" coming from `jupyterlab-conda-store`:
   image

### After

- single "Services > Environment Manager" which opens `jupyterlab-conda-store` if available or `/conda-store` in new tab otherwise.

image

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

### DIFF
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterlab/overrides.json
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterlab/overrides.json
@@ -51,7 +51,23 @@
                 "rank": 1000,
                 "items": [
                     {
-                        "command": "condastore:open",
+                        "command": "nebari:run-first-enabled",
+                        "args": {
+                            "commands": [
+                                {
+                                  "id": "condastore:open",
+                                  "label": "Environment Management"
+                                },
+                                {
+                                  "id": "help:open",
+                                  "args": {
+                                    "url": "/conda-store",
+                                    "text": "Environment Management",
+                                    "newBrowserTab": true
+                                  }
+                                }
+                            ]
+                        },
                         "rank": 1
                     },
                     {
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterlab/overrides.json
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterlab/overrides.json
@@ -6,7 +6,8 @@
         "apiUrl": "/conda-store/",
         "authMethod": "cookie",
         "loginUrl": "/conda-store/login?next=",
-        "authToken": ""
+        "authToken": "",
+        "addMainMenuItem": false
     },
     "@jupyterlab/apputils-extension:notification": {
         "checkForUpdates": false,
@@ -50,13 +51,8 @@
                 "rank": 1000,
                 "items": [
                     {
-                        "command": "help:open",
-                        "rank": 1,
-                        "args": {
-                            "url": "/conda-store",
-                            "text": "Environment Management",
-                            "newBrowserTab": true
-                        }
+                        "command": "condastore:open",
+                        "rank": 1
                     },
                     {
                         "command": "help:open",


dcmcand (approved):
Thanks @krassowski

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterlab/overrides.json
```diff
@@ -50,13 +51,8 @@
                 "rank": 1000,
                 "items": [
                     {
-                        "command": "help:open",
-                        "rank": 1,
-                        "args": {
-                            "url": "/conda-store",
-                            "text": "Environment Management",
-                            "newBrowserTab": true
```
viniciusdc, line 58: Without this, does the extension still opens as a separate tab?

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterlab/overrides.json
```diff
@@ -50,13 +51,8 @@
...
-                        }
+                        "command": "condastore:open",
```
viniciusdc, line 54: Hi @krassowski thanks for this, could you explain a bit how this works? does this call the extension, or the endpoint for the conda-store webpage as well?

viniciusdc (commented):
LGTM as well, though just to make sure, this is used when the user opts for installing the extension or redirects to the main page?

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterlab/overrides.json
krassowski: It opens conda-store-ui as a tab in jupyterlab main area:

conda-store

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterlab/overrides.json
krassowski: That code was opening the `/conda-store` in a new browser tab.

krassowski:
> this is used when the user opts for installing the extension what happens when the extension is not available?

The extension is included in the default docker image, so there is no opting-in. When extension is not available because user opts-out, the menu item is not show, but the other menu items are shown.

We _could_ have it both ways: when extension is available, open it as a JupyterLab tab; when extension is not available open it in a new browser tab. I can make this work straightaway by adding some logic in `jupyterlab-nebari-mode`.

krassowski:
A dedicated logic (new command) in `jupyterlab-nebari-mode` would also allow to change the label so maybe it is desirable.

krassowski:
Done. I updated top-level issue. This now depends on https://github.com/nebari-dev/nebari-docker-images/pull/130. Tested locally with the new image. Recording:

local-test

krassowski:
Is this and https://github.com/nebari-dev/nebari-docker-images/pull/130 good to merge?

viniciusdc (approved):
LGTM!!

Issue #2174, opened 2023-12-28
[BUG] - Traefik rate limited by Let's Encrypt after several pod restarts


sblair-metrostar:
### Describe the bug

[Traefik documentation](https://doc.traefik.io/traefik/https/acme/) warns to persist the acme.json certificate storage to avoid being rate limited by Let's Encrypt, but Nebari's deployment isn't configuring that. If the traefik pod is restarted enough times within a given week without this setup, Let's Encrypt will rate limit the certificate issue request and cause the site to default back to self-signed.

This has happened to me [many, many times](https://github.com/nebari-dev/nebari/issues/1581#issuecomment-1335570056) when I'm developing or testing changes to the platform or dealing with conda storage related node crashes, usually requiring a workaround involving installation of cert-manager and switching to zerossl or another cert issuing service until the Let's Encrypt backoff period elapses.

### Expected behavior

traefik pod restarts should not result in new certificates being issued when configured for lets-encrypt.

Options:
1. Configure a persistent volume mount for the [acme.json](https://github.com/nebari-dev/nebari/blob/d612e9b5f63db0e810c1fe6fc0cd61b5bfa62641/src/_nebari/stages/kubernetes_ingress/template/modules/kubernetes/ingress/main.tf#L12) storage path.
2. Replace Traefik's native Let's Encrypt integration with cert-manager for more issuer options and improved scalability.

### OS and architecture in which you are running Nebari

Linux, x64

### How to Reproduce the problem?

1. Configure the Nebari certificate with lets-encrypt

``` yaml

certificate:
  type: lets-encrypt
  acme_email: notreal@example.com
  acme_server: https://acme-v02.api.letsencrypt.org/directory

```

2. Delete the traefik pod 5 or more times in 1 week.

### Versions and dependencies used.

Nebari 2023.11.1

### Compute environment

AWS

Princess4ogb:
I am working on this ticket, Please find the gitlab issue as below.
https://gitlab.jatic.net/jatic/team-metrostar/t-e-platform/-/issues/345

kcpevey:
Resolved by #2352

PR #2352, opened 2024-03-20
PVC for Traefik Ingress (prevent LetsEncrypt throttling)

kenafoster:

This is an interim fix for #2174 .  It sets up a PVC for the traefik ingress pod so that the cert is on persistent storage, meaning that as new pods get recreated, they will not request new certs (which leads to the throttling problem mentioned in the issue).

The "best fix" is cert-manager or another service will handle all these requests.  That's being worked on - ref #2336 

In AWS, the storage class it uses is EBS-backed and therefore specific to an AZ, which means that this is not truly HA / fault tolerant (the pod will always schedule and launch on a node in the same AZ).  However, there are many resources in the kubernetes_services stage that follow this same pattern so this is not introducing a new limitation, and is only intended as an interim fix anyway.

Note I have ONLY tested this in AWS.  Need assistance running in other clouds.  Specifically, I am not sure that the access_mode will be universal because it is relying on Kubernetes to determine the default storage class and create the volume.  This seems to work OK for helm charts that do this within the `kubernetes_services` stage but this resource gets created in a prior stage, so there could be some dependency/configuration that I am missing.

- [ X ] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [X ] Did you test the pull request locally?
Local and AWS cloud
- [ ] Did you add new tests?

### DIFF
--- a/src/_nebari/stages/kubernetes_ingress/template/modules/kubernetes/ingress/main.tf
+++ b/src/_nebari/stages/kubernetes_ingress/template/modules/kubernetes/ingress/main.tf
@@ -9,7 +9,7 @@ locals {
       "--entrypoints.minio.http.tls.certResolver=letsencrypt",
       "--certificatesresolvers.letsencrypt.acme.tlschallenge",
       "--certificatesresolvers.letsencrypt.acme.email=${var.acme-email}",
-      "--certificatesresolvers.letsencrypt.acme.storage=acme.json",
+      "--certificatesresolvers.letsencrypt.acme.storage=/mnt/acme-certificates/acme.json",
       "--certificatesresolvers.letsencrypt.acme.caserver=${var.acme-server}",
     ]
     self-signed = local.default_cert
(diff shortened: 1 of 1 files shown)


dcmcand:
@marcelovilla @viniciusdc @Adam-D-Lewis whoever has time, can we test this on azure and gcp for @tylergraff ?

viniciusdc:
I just tested this on GCP, and it works as expected; the access level is being handled correctly; my only issue is that if the user needs to update its domain, then we need to manually delete the certs.json file from the PVC, as Traefikdoes not seem to be doing it right now. (before this problem was mitigated with the absence of the file during pod init)

viniciusdc:
I've created [an issue](https://github.com/nebari-dev/nebari-docs/issues/421) in our docs repo to track the details outlined above, I will merge the PR as it is working and address the original issue.

viniciusdc (approved):
LGTM

Issue #2308, opened 2024-03-12
[ENH] - Make JupyterHub use groups and roles from Keycloak


aktech:
### Feature description

Until now we haven't been using JupyterHub groups and roles much. We have Keycloak as the identity provider and we plan to use groups and roles more in keycloak for permissions overhaul, see following issues
- #2304
- https://github.com/nebari-dev/jhub-apps/issues/11

The main motivation for this is to be able to fetch groups and roles from the JupyterHub API in jhub-apps to be able to decide permissions, since jhub-apps is not supposed to be tied to Nebari, hence would be great to be able to fetch roles and groups from JupyterHub API in jhub-apps.

Relevant links
- https://discourse.jupyter.org/t/is-jupyterhub-rbac-groups-the-same-as-oauth-groups/22412/3
- https://discourse.jupyter.org/t/jupyterhub-keycloak-auth-and-ldap-user-groups/22512
- Working configuration for generic authenticator with Keycloak jupyterhub/oauthenticator#107
- https://discourse.jupyter.org/t/oidc-rbac-possible-to-map-users-to-groups-and-groups-to-roles-where-users-and-groups-are-defined-by-oidc/22426

I reckon, we might have to make changes to our Authenticator to make this happen.
https://github.com/nebari-dev/nebari/blob/53194474dfbc8ac1ded81737edc777c17c6bbe97/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/dask-gateway/files/gateway_config.py#L63

Definition of done:
- Keycloak roles and groups are accessible from JupyterHub API

https://jupyterhub.readthedocs.io/en/stable/reference/rest-api.html#/default/get_groups

Currently, this is what I get on the fetching groups from JupyterHub:

```json
{
	"last_activity": "2024-03-11T15:59:01.194646Z",
	"server": null,
	"groups": [],
	"created": "2022-12-07T16:43:29.002132Z",
	"auth_state": null,
	"name": "akumar@quansight.com",
	"kind": "user",
	"pending": null,
	"admin": true,
	"roles": [
		"user",
		"admin"
	],
	"servers": {}
	"session_id": null,
	"scopes": ["truncated"]
}
```

You can see the groups are empty and roles are also not the ones from keycloak.

### Value and/or benefit

This will help us implement app sharing and permissioning seamlessly with keycloak.

krassowski:
Thanks for the extra details!

> I reckon, we might have to make changes to our Authenticator to make this happen.

[...]

Issue #2304, opened 2024-03-11
[META] Permissions overhaul


aktech:
### Context

This is a meta issue to define and document problems with the current permissions model in Nebari

<h1>Nebari Permissions</h1>
<h1>🛡️Authentication</h1>
<h2>Login / Logout to Services or Components</h2>

Application | Authentication Required? | Login Behavior | Logout Behavior
-- | -- | -- | --
Keycloak | Yes | - | -
Monitoring | Yes | Logs in automatically if logged in to Nebari | Nebari Home logout doesn't logs you out of Monitoring
VSCode / JupyterLab | Yes | Logs in automatically if logged in to Nebari | Nebari Home logout logs you out of VSCode/JupyterLab
conda-store | Yes | Requires another login (via Keycloak) | Redirects to localhost after logout, no actual logout (conda-store bug?)
Argo | Yes | Requires another login (via Keycloak) | Nebari Home logout doesn't logs you out of Argo
Dask Diagnostics Dashboard | No (Accessible without auth) | N/A | N/A
Extenstions (e.g. mlflow) | don't care yet | don't care yet | don't care yet

<h1>🛡️Authorization</h1>
<h2>Permission Profiles (JupyterLab Profiles)</h2>
<pre><code class="language-yaml">profiles:
  jupyterlab:
    - display_name: Small Instance
      description: Stable environment with 1 cpu / 1 GB ram
      access: all
      default: true
      kubespawner_override:
        cpu_limit: 1
        cpu_guarantee: 1
        mem_limit: 1G
        mem_guarantee: 1G
    - display_name: Medium Instance
    ...

</code></pre>
<ul>
<li>It is possible to control which users have access to which JupyterLab profiles. Each profile has a field named <code>access</code> which can be set to all (default if omitted), <code>yaml</code>, or <code>keycloak</code>.</li>
<li><code>all</code> means every user will have access to the profile (default).</li>
<li><code>keycloak</code> means that access is restricted to any user who in Keycloak has either their group(s) or user with the attribute jupyterlab_profiles containing this profile name. For example, if the user is in a Keycloak group named developers which has an attribute jupyterlab_profiles set to Large Instance, they will have access to the Large Instance profile. To specify multiple profiles for one group (or user) delimit their names using ## - for example, Large Instance##Another Instance.</li>
</ul>
<h1>Permissions via Keycloak</h1>
<h2>Groups</h2>
<ul>
<li><code>analyst</code></li>
<li><code>developer</code></li>
[...]

PR #2361, opened 2024-03-27
Get JupyterHub `groups` from Keycloak, support `oauthenticator` 16.3+

krassowski:
## Reference Issues or PRs

- A task for https://github.com/nebari-dev/nebari/issues/2308
- Can be merged straightaway but for the `groups` to get populated from keycloak requires https://github.com/nebari-dev/nebari-docker-images/pull/131

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

To test:

1. Set the image to use https://github.com/nebari-dev/nebari-docker-images/pull/131:
  ```
  default_images:
    jupyterhub: quay.io/nebari/nebari-jupyterhub:oauthenticator-16.3.0
  ```
2. Deploy
3. Open https://your-nebari-deployment/hub/token
4. Open Dev Tools in your browser, switch to Network tab
5. Press "Request new API token"
6. Right click on the request in Network tab → copy as fetch
7. Paste the request, change method to "GET", remove "body", change URL to `/hub/api/users`
8. Re-send the request
9. Click on the new request, preview the response
10. It should contain groups from keycloak e.g.:
  ```json
  [
      {
          "kind": "user",
          "auth_state": null,
          "groups": [
              "grafana_developer",
              "query-users",
              "manage-identity-providers",
              "manage-clients",
              "manage-account",
              "manage-realm",
              "view-profile",
              "argo-admin",
              "dask_gateway_developer",
              "grafana_admin",
              "view-identity-providers",
              "jupyterhub_admin",
              "view-realm",
              "view-authorization",
              "jupyterhub_developer",
              "view-clients",
              "query-groups",
              "conda_store_developer",
              "view-events",
              "query-realms",
              "impersonation",
              "realm-admin",
              "create-client",
              "conda_store_superadmin",
              "argo-viewer",
              "argo-developer",
              "manage-events",
              "grafana_viewer",
              "manage-users",
              "dask_gateway_admin",
              "manage-account-links",
              "manage-authorization",
              "query-clients",
              "view-users",
              "conda_store_admin"
          ],
          "admin": true,
          "roles": [
              "user",
              "admin"
          ],
          "name": "USER",
          "pending": null,
          "server": "/user/USER/",
          "servers": {
              "": {
                  "name": "",
                  "pending": null,
                  "ready": true,
                  "stopped": false,
                  "url": "/user/USER/",
                  "user_options": {
                      "profile": "small-instance"
                  },
                  "progress_url": "/hub/api/users/USER/server/progress",
                  "state": {
                      "pod_name": "jupyter-USER"
                  }
              }
          }
      }
  ]
  ```

### DIFF
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
@@ -150,25 +150,25 @@ resource "helm_release" "jupyterhub" {
             enable_auth_state = true
           }
           GenericOAuthenticator = {
-            client_id          = module.jupyterhub-openid-client.config.client_id
-            client_secret      = module.jupyterhub-openid-client.config.client_secret
-            oauth_callback_url = "https://${var.external-url}/hub/oauth_callback"
-            authorize_url      = module.jupyterhub-openid-client.config.authentication_url
-            token_url          = module.jupyterhub-openid-client.config.token_url
-            userdata_url       = module.jupyterhub-openid-client.config.userinfo_url
-            login_service      = "Keycloak"
-            username_claim     = "preferred_username"
-            claim_groups_key   = "groups"
-            allowed_groups     = ["jupyterhub_admin", "jupyterhub_developer"]
-            admin_groups       = ["jupyterhub_admin"]
-            manage_groups      = true
-            refresh_pre_spawn  = true
+            client_id            = module.jupyterhub-openid-client.config.client_id
+            client_secret        = module.jupyterhub-openid-client.config.client_secret
+            oauth_callback_url   = "https://${var.external-url}/hub/oauth_callback"
+            authorize_url        = module.jupyterhub-openid-client.config.authentication_url
+            token_url            = module.jupyterhub-openid-client.config.token_url
+            userdata_url         = module.jupyterhub-openid-client.config.userinfo_url
+            login_service        = "Keycloak"
+            username_claim       = "preferred_username"
+            claim_groups_key     = "groups"
+            allowed_groups       = ["jupyterhub_admin", "jupyterhub_developer"]
+            admin_groups         = ["jupyterhub_admin"]
+            manage_groups        = true
+            refresh_pre_spawn    = true
             validate_server_cert = false
 
             # deprecated, to be removed (replaced by validate_server_cert)
-            tls_verify         = false
+            tls_verify = false
             # deprecated, to be removed (replaced by username_claim)
-            username_key       = "preferred_username"
+            username_key = "preferred_username"
           }
         }
       }
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
@@ -159,8 +159,8 @@ resource "helm_release" "jupyterhub" {
             login_service        = "Keycloak"
             username_claim       = "preferred_username"
             claim_groups_key     = "groups"
-            allowed_groups       = ["jupyterhub_admin", "jupyterhub_developer"]
-            admin_groups         = ["jupyterhub_admin"]
+            allowed_groups       = ["/analyst", "/developer", "/admin"]
+            admin_groups         = ["/admin"]
             manage_groups        = true
             refresh_pre_spawn    = true
             validate_server_cert = false
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
@@ -157,11 +157,18 @@ resource "helm_release" "jupyterhub" {
             token_url          = module.jupyterhub-openid-client.config.token_url
             userdata_url       = module.jupyterhub-openid-client.config.userinfo_url
             login_service      = "Keycloak"
-            username_key       = "preferred_username"
-            claim_groups_key   = "roles"
+            username_claim     = "preferred_username"
+            claim_groups_key   = "groups"
             allowed_groups     = ["jupyterhub_admin", "jupyterhub_developer"]
             admin_groups       = ["jupyterhub_admin"]
+            manage_groups      = true
+            refresh_pre_spawn  = true
+            validate_server_cert = false
+
+            # deprecated, to be removed (replaced by validate_server_cert)
             tls_verify         = false
+            # deprecated, to be removed (replaced by username_claim)
+            username_key       = "preferred_username"
           }
         }
       }


krassowski:
The cypress failure indicates that permissions did change. It looks like we need to adjust test permissions for the test to pass:

First Test -- Check Nebari login and start JupyterLab (failed)

krassowski:
I can reproduce the failure when using the old `oauthenticator` version; the intent of this PR was to make it backward compatible, so I will see if we can change something to make it such.

krassowski:
Previously the keycloak `roles` where mapped to JupyterHub `groups`:

https://github.com/nebari-dev/nebari/blob/3e1dfde9c56cfe6c151b36333713d6fbea30bf86/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf#L247-L251

https://github.com/nebari-dev/nebari/blob/3e1dfde9c56cfe6c151b36333713d6fbea30bf86/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf#L161-L162

Of note previously extracting the group data (for some reason from roles) was working, but these groups were not persisted in JupyterHub (which is why we are moving to `oauthenticator` 16.3+).

The logs:

```
│ [I 2024-03-27 18:09:45.199 JupyterHub generic:185] Validating if user claim groups match any of ['jupyterhub_admin', 'jupyterhub_developer']                                                                                                                                                        │
│ [W 2024-03-27 18:09:45.199 JupyterHub base:843] Failed login for unknown user                                                                                                                                                                                                                       │
```

suggest that validation for non-admin users fails on `oauthenticator` < 16.3 because...  they are not in a group which is in  `allowed_groups`?. But then, it passes on 16.3 because users are in `jupyterhub_developer` group by default. This does not make sense, why would it not be there in <16.3?

krassowski:
Looking more through the codebase, it appears that the idea of mapping from keycloak group to a service-specific group roles is prevalent in `nebari`.

I think that the solution is to align the JupyterHub and keycloak roles as follows:
- change the `claim_groups_key` to "groups" (already done)
- change `allowed_groups` in JupyterHub to `['/analyst', '/developer', '/users']` (to be done)
- fetch the `roles` from keycloak in addition to "groups"

@aktech any thoughts?

aktech:
> change the claim_groups_key to "groups" (already done)

I assume this [will look for the key "`groups`"](https://oauthenticator.readthedocs.io/en/latest/reference/api/gen/oauthenticator.generic.html#oauthenticator.generic.GenericOAuthenticator.claim_groups_key) in the keycloak's userdata url to populate groups in JupyterHub? If yes, then it makes sense.

>change allowed_groups in JupyterHub to ['/analyst', '/developer', '/admin'] (to be done)

Will that prevent users in any other group to login to JupyterHub (Nebari)?

Context on this: We would want people to be able to create any number of groups with any roles (pre-existing) attached to them. The arbitrary groups creation is already possible. The arbitrary roles creation is not possible yet as they are static in nebari at the moment (explicitly mapped to specific service roles), but after the [permissions overhaul](https://github.com/nebari-dev/nebari/issues/2304), we should be able to create arbitrary roles, like say "*I want to create a role to have read access to conda environments in the xyz namespace*", hence it's important to be able to fetch all groups/roles from keycloak, irrespective of the fact they are created at the deployment time or after deployment via keycloak's GUI.


> fetch the roles from keycloak in addition to "groups"

Yes.

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
```diff
@@ -150,18 +150,25 @@ resource "helm_release" "jupyterhub" {
...
-            tls_verify         = false
+            client_id            = module.jupyterhub-openid-client.config.client_id
+            client_secret        = module.jupyterhub-openid-client.config.client_secret
+            oauth_callback_url   = "https://${var.external-url}/hub/oauth_callback"
+            authorize_url        = module.jupyterhub-openid-client.config.authentication_url
+            token_url            = module.jupyterhub-openid-client.config.token_url
+            userdata_url         = module.jupyterhub-openid-client.config.userinfo_url
+            login_service        = "Keycloak"
+            username_claim       = "preferred_username"
+            claim_groups_key     = "groups"
```
aktech, line 161: Side note:
It's quite weird on [`oauthenticator`](https://github.com/jupyterhub/oauthenticator) side, that one of them is called: `username_claim` and other one is called: `claim_groups_key` instead of consistent names.

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
```diff
@@ -150,18 +150,25 @@ resource "helm_release" "jupyterhub" {
...
+            allowed_groups       = ["/analyst", "/developer", "/admin"]
+            admin_groups         = ["/admin"]
```
aktech, line 163: Would be better if this is dynamic, is there a way to do that?

Like admin group is the one with x, y, z roles attached to it.

aktech (commented):
I have one comment regarding how `allowed_groups` and `admin_groups` is set.

krassowski:
> Will that prevent users in any other group to login to JupyterHub (Nebari)?

Yes. But! We don't need to use `allowed_groups` for allow-listing users. There is a few choices here:
- `allowed_users`/`admin_users` - well, not very flexible
- `check_allowed()`/`is_admin()`  - we can implement arbitrary logic
- do nothing and just rely on roles

Ultimately I think we want the last one, but at this time the roles syncing is work in progress.

I think I can already make it work by overriding `check_allowed()`/`is_admin()` in the `Authenticator` because the `auth_state` already does contain the roles from Keycloak.

Do we want to do that in this PR?

aktech:
I am curious, do we even need this `allowed_*`? As in, is it necessary to have them? I think it's fine to allow all groups and roles from keycloak.

> do nothing and just rely on roles

Will this fetch all the groups from keycloak? If yes, then this is what we need.

> Ultimately I think we want the last one, but at this time the roles syncing is work in progress.
> I think I can already make it work by overriding check_allowed()/is_admin() in the Authenticator because the auth_state already does contain the roles from Keycloak.

Since we're fetching all the groups as per third option, this is not required.

> Do we want to do that in this PR?

Certainly not, separate PR is better infact.

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
aktech: Just catching up, do we still need allowed group?

Ref: https://github.com/nebari-dev/nebari/pull/2361#issuecomment-2030069255

krassowski:
> I am curious, do we even need this `allowed_*`? As in, is it necessary to have them? I think it's fine to allow all groups and roles from keycloak.

By default OAuthenticator will not allow any user unless they meet one of the `allowed_*` rules, [docs](https://oauthenticator.readthedocs.io/en/latest/topic/allowing.html):

> #### Default behavior: nobody is allowed!
> The default behavior of OAuthenticator (starting with version 16) is to block all users unless explicitly authorized via some `allow` configuration. If you want anyone to be able to use your hub, you must specify at least one `allow` configuration.
>
> Changed in version 16: Prior to OAuthenticator 16, `allow_all` was implied if no other allow configuration was specified. Starting from 16, `allow_all` can only be enabled explicitly.

What you are proposing is essentially setting `allow_all = True`. We can do that. I am not quite if there are any downsides, but here is what the docs has to say about it:

> This is appropriate when you use an authentication provider (e.g. an institutional single-sign-on provider), where everyone who has an account in the provider should have access to your Hub. It may also be appropriate for unadvertised short-lived hubs, e.g. dedicated hubs for workshops that will be shutdown after a day, where you may decide it is acceptable to allow anyone who finds your hub to login.

Is there a scenario when a nebari user would be allowed access to say conda-store but not to JupyterHub? I guess it does not make sense because conda-store is a resource for JupyterHub users, and any user who is trusted with conda-store would usually also trusted to access JupyterHub... unless in a high confidentiality deployment when someone wanted to get an external eye on conda-store config, but without exposing the shared file system? But we can also have fine-grained file system permissions so I guess it does not matter if configured properly?

krassowski:
FYI while I was referencing OAuthenticator above, this behaviour is also going to be the same in JupyterHub 5.0 in general, see https://github.com/jupyterhub/jupyterhub/pull/4701.

aktech:
>By default OAuthenticator will not allow any user unless they meet one of the allowed_* rules, [docs](https://oauthenticator.readthedocs.io/en/latest/topic/allowing.html):

It's fine to have `allowed_*` as long as it can be synced with keycloak, ideally without restarting the hub. Imagine, we add a new group with x, y, z (roles) permissions and we would like to make sure it takes effect without having to restart the hub. If the user needs to logout/login, that's a reasonable scenario.

> Is there a scenario when a nebari user would be allowed access to say conda-store but not to JupyterHub?

I don't think so.

> I guess it does not make sense because conda-store is a resource for JupyterHub users, and any user who is trusted with conda-store would usually also trusted to access JupyterHub... unless in a high confidentiality deployment when someone wanted to get an external eye on conda-store config, but without exposing the shared file system? But we can also have fine-grained file system permissions so I guess it does not matter if configured properly?

Yes, that makes sense. Regarding fine-grained permissions here is a proposal: https://github.com/nebari-dev/governance/issues/47

aktech:
> It's fine to have allowed_* as long as it can be synced with keycloak, ideally without restarting the hub. Imagine, we add a new group with x, y, z (roles) permissions and we would like to make sure it takes effect without having to restart the hub. If the user needs to logout/login, that's a reasonable scenario.

Also, If this is true, this PR is good to merge.

krassowski:
> It's fine to have allowed_* as long as it can be synced with keycloak, ideally without restarting the hub. Imagine, we add a new group with x, y, z (roles) permissions and we would like to make sure it takes effect without having to restart the hub. If the user needs to logout/login, that's a reasonable scenario.

The groups are synced from keycloak, but `allowed_groups` are hard-coded. Is that what you meant?

aktech:
>The groups are synced from keycloak, but allowed_groups are hard-coded. Is that what you meant?

I meant all groups should be allowed, but I see what's happening. I understand now that we need to make sure every user who's allowed to login to Nebari (JupyterHub), needs to be in one of the `allowed_groups` (which we do already now), that's fine, we can have minimal permissions in one of the  `allowed_groups` and make sure every user is there by default so that they can login.

So, no change required now, this PR looks good to me.

PR #2384, opened 2024-04-08
Add KEDA HPA TriggerAuthentication and postgresql ScaledObject.

### LINKED ISSUE #2284: [ENH] - Add kubernetes horizontal autoscaler for conda-store workers based on queue depth
### Feature description

Currently conda-store is set to allow 4 simultaneous builds at once.  This is a bottleneck once multiple environments start getting built at once and presents a scaling challenge. If we set the simultaneous builds to 1 and autoscale based on queue depth then we should be able to handle scaling far more gracefully

### Value and/or benefit

Having the conda-store workers autoscale based on queue depth will allow larger orgs to take advantage of Nebari without hitting scale bottlenecks.

### Anything else?

https://learnk8s.io/scaling-celery-rabbitmq-kubernetes

pt247:
## Reference Issues or PRs
Fixes #2284 

Adds KEDA HPA TriggerAuthentication and PostgreSQL ScaledObject.
In the current `develop`, one pod for `conda-store-worker` is always running. With this PR, the 
`conda-store-worker` pods will scale down to zero at the start, scale up with the number of pending/queued builds, and scale down to zero when builds are complete. 

- [x] Documentation Update [PR](https://github.com/nebari-dev/nebari-docs/pull/459)
- [ ] Build related changes

## Testing

- [x] Did you add new tests?: New local integration test added. 

## Any other comments?

1. We can now set the following parameters while configuring conda-store:
```
conda_store:
  max_workers: 50
  worker_resources:
    requests:
      cpu: 1
      memory: 4Gi
```

2. The local integration test is slow and adds 10-15 minutes to local integration tests.
3. We can add follow-up PRs to expose `KEDA` and `KEDA.ScaledObject` parameters if needed.


Adam-D-Lewis:
We should also set a memory request for the conda store worker deployment (I'd suggest a minimum of 1GiB), and allow the general node group to scale above 1 by default for each provider (https://github.com/nebari-dev/nebari/blob/a5cbb138fddf1551fccaa0573b10dc43b0e8f930/src/_nebari/stages/infrastructure/__init__.py#L390).  Otherwise, if we try to build 100 conda envs simultaneously, we'll get 100 conda worker pods.  They will all be scheduled on a single general node (assuming only 1 is running).  The general node won't even try to scale up, but the general node won't have enough memory to solve all the envs at the same time.

Adam-D-Lewis (changes requested):
We should also set a memory request for the conda store worker deployment (I'd suggest a minimum of 1GiB), and allow the general node group to scale above 1 (e.g. set max_nodes default value to 5) by default for each provider e.g. https://github.com/nebari-dev/nebari/blob/a5cbb138fddf1551fccaa0573b10dc43b0e8f930/src/_nebari/stages/infrastructure/__init__.py#L390.  

Otherwise, if we try to build 100 conda envs simultaneously, we'll get 100 conda worker pods.  They will all be scheduled on a single general node (assuming only 1 is running).  The general node won't even try to scale up (b/c the conda store worker doesn't have any resource requests set), but the general node won't have enough memory to solve all the envs at the same time and many will fail.

We should also make this part of the next Nebari release to tell existing deployments that it's recommended that they increase the max_nodes setting of the nebari-config if it is set explicitly to 4 or less.

It also might be a good idea to set a maximum number of conda store workers as a safe guard (even if the max number is high).  E.g. If 100 envs are all scheduled at once, maybe we should only start a maximum of 50 workers.  (In the KEDA sql command maybe take the min of the number of queued/building envs and 50 for example).  Ideally that max would be configurable in the nebari config.

dharhas:
Are there conda env solves that require more than 1 GB of ram? I remember early on we had to keep bumping up the ram on the conda-store main node.

Adam-D-Lewis:
> Are there conda env solves that require more than 1 GB of ram? I remember early on we had to keep bumping up the ram on the conda-store main node.

Yes, there are certainly conda solves (most?) that use more than 1 GB of RAM.  I don't think there's a way to know beforehand to estimate how much RAM a conda solve will need.  Setting the memory resource request doesn't limit the pod from using more RAM if available, but it would put a limit on the number of conda store workers scheduled on the same pod.

pt247:
> It also might be a good idea to set a maximum number of conda store workers as a safe guard (even if the max number is high). E.g. If 100 envs are all scheduled at once, maybe we should only start a maximum of 50 workers. (In the KEDA sql command maybe take the min of the number of queued/building envs and 50 for example). Ideally that max would be configurable in the nebari config under the conda_store section.


Default as per the documentation - [link](https://keda.sh/docs/2.13/concepts/scaling-deployments/) is 
` maxReplicaCount:  100`

Is 100 enough?

pt247:
I have tried to create 100 conda environments for testing, and not all of them transition to a completed state. Some 24 get stuck in a queued state with the following error. 

```
[2024-04-15 00:08:26,319: WARNING/Beat]   File "/opt/conda/envs/conda-store-server/lib/python3.10/dbm/dumb.py", line 316, in open return _Database(file, mode, flag=flag)                                                                                                                   
[2024-04-15 00:08:26,320: WARNING/Beat]   File "/opt/conda/envs/conda-store-server/lib/python3.10/dbm/dumb.py", line 70, in __init__ self._create(flag)       
[2024-04-15 00:08:26,320: WARNING/Beat]   File "/opt/conda/envs/conda-store-server/lib/python3.10/dbm/dumb.py", line 86, in _create with _io.open(self._datfile, 'w', encoding="Latin-1") as f:                                                                                               
[2024-04-15 00:08:26,320: WARNING/Beat]   FileNotFoundError: [Errno 2] No such file or directory: '/root/.conda-store/celerybeat-schedule.dat'    
```
@dcmcand In this case, should we roll this out based on the manual testing we have done so far? 

@Adam-D-Lewis Are you happy with the other changes?

dcmcand:
> I have tried to create 100 conda environments for testing, and not all of them transition to a completed state. Some 24 get stuck in a queued state with the following error.
> 
> ```
> [2024-04-15 00:08:26,319: WARNING/Beat]   File "/opt/conda/envs/conda-store-server/lib/python3.10/dbm/dumb.py", line 316, in open return _Database(file, mode, flag=flag)                                                                                                                   
> [2024-04-15 00:08:26,320: WARNING/Beat]   File "/opt/conda/envs/conda-store-server/lib/python3.10/dbm/dumb.py", line 70, in __init__ self._create(flag)       
> [2024-04-15 00:08:26,320: WARNING/Beat]   File "/opt/conda/envs/conda-store-server/lib/python3.10/dbm/dumb.py", line 86, in _create with _io.open(self._datfile, 'w', encoding="Latin-1") as f:                                                                                               
> [2024-04-15 00:08:26,320: WARNING/Beat]   FileNotFoundError: [Errno 2] No such file or directory: '/root/.conda-store/celerybeat-schedule.dat'    
> ```
> 
> @dcmcand In this case, should we roll this out based on the manual testing we have done so far?
> 
> @Adam-D-Lewis Are you happy with the other changes?
@pt247 After adding the changes that @Adam-D-Lewis requested, can you redo your load test? We need to be able to gracefully scale conda-store workers. Ensuring that we have guidance around what resources/configs are needed for what scale would be important too. Finally, please also open a docs PR to document how to configure this.

pt247:
@Adam-D-Lewis, I have added the resource limits. I tried to create 200 environments at once. I can observe:
1. The `general` "nodegroup" scales up and down as expected.
2. KEDA only tries to schedule 50 nodes at once as expected.
3. Lots of pods are stuck in the `creating` state with the error `stream logs failed container "nfs-server" in pod "nebari-conda-store-worker-55f5f899c7-t748x" is waiting to start: ContainerCreating for dev/nebari-conda-store-worker-55f5f899c7-t748x (nfs-server)`
4. Some are stuck in the `running` state with the error ` WARNING/Beat] FileNotFoundError: [Errno 2] No such file or directory: '/root/.conda-store/celerybeat-schedule.dat'`
5. Some environments were successfully created.

pt247:
>  Finally, please also open a docs PR to document how to configure this.

@dcmcand None of the KEDA configurations are exposed to the Nebari config. Do we need a docs PR?

dcmcand:
> > Finally, please also open a docs PR to document how to configure this.
> 
> @dcmcand None of the KEDA configurations are exposed to the Nebari config. Do we need a docs PR? Could you let me know where we should add info about this? It can be in FAQ.

The node group options are exposed and this affects their settings. So we need documentation around how this works and how that changes your node group options.

Additionally, as @Adam-D-Lewis mentioned in https://github.com/nebari-dev/nebari/pull/2384#pullrequestreview-1989478849 the ability to set the max conda-store workers would be good to add rather than just hardcoding it. That could be a follow on pr (if so please create an issue for it), but it will require docs too.

Adam-D-Lewis:
@pt247 Here is what I'd suggest adding to the nebari-config.yaml
from src/_nebari/stages/kubernetes_services/__init__.py
image
which could then be adjusted in the nebari-config.yaml via something like:
image

Then you'd need to hook up the new pydantic attributes to the terraform so changing the values in the config actually affects the deployment.

pt247:
@Adam-D-Lewis,  @pavithraes The following overrides will soon be available in conda-config. 
```
conda_store:
  max_workers: 50
  worker_resources:
    requests:
      cpu: 500m
      memory: 1Gi
```
Where should I document these in Nebari docs? 

Is this the right place? 
https://www.nebari.dev/docs/explanations/custom-overrides-configuration

Also, KEDA scales up the workers when needed and scales them back to zero.  
Do users care about this? If so, is Nebari [FAQ](https://www.nebari.dev/docs/faq) the right place to document this?

Adam-D-Lewis:
> Where should I document these in Nebari docs?
> Is this the right place?
[nebari.dev/docs/explanations/custom-overrides-configuration](https://www.nebari.dev/docs/explanations/custom-overrides-configuration)

Yes, that seems like the best place imo for now.

Adam-D-Lewis:
> Also, KEDA scales up the workers when needed and scales them back to zero.
Do users care about this? If so, is Nebari [FAQ](https://www.nebari.dev/docs/faq) the right place to document this?

I don't think it's very important to mention this since we talk about other parts of nebari scaling dynamically for cost savings.

Adam-D-Lewis:
Looks like integration tests are failing currently.  Could you fix that before another review, @pt247 ?

pt247:
@Adam-D-Lewis The [builds](https://github.com/nebari-dev/nebari/actions/runs/8929296899/job/24526822730?pr=2384) seem to time out while creating PVC for conda-store. Would you have any idea why?

src/_nebari/stages/kubernetes_services/__init__.py
```diff
@@ -180,6 +180,8 @@ class CondaStore(schema.Base):
     image_tag: str = constants.DEFAULT_CONDA_STORE_IMAGE_TAG
     default_namespace: str = "nebari-git"
     object_storage: str = "200Gi"
+    max_workers: int = 50
+    worker_resources: dict = {"requests": {"cpu": "1", "memory": "4Gi"}}
```
pt247, line 184: @Adam-D-Lewis This are the minimum recommended resources as per conda-store [docs](https://conda.store/conda-store/references/faq#what-are-the-resource-requirements-for-conda-store-server). 

Below is the resources consumed on a build (scikit-learn, pandas) when there is no sizing is in place (in develop)

Image

As you can see that this uses about 7 GB. 
If we simply increate resource allocation to 7GB. Nebari CI test fails to deploy (timeout - typical of lack of memory).

Should the resource sizing be a separate ticket? KEDA scales on Queue so its behaviour will remain same.

tests/tests_deployment/test_conda_store_scaling.py
```diff
@@ -0,0 +1,245 @@
...
+            )
+            return api_response
+        except ApiException as e:
+            self.log.exception(
+                "Exception when calling CoreV1Api->read_namespaced_config_map: %s\n" % e
+            )
+        finally:
+            _client.close()
+
+    def patch_namespaced_config_map(self, config_map):
```
pt247, line 73: @viniciusdc This trick of patching the config is working when running tests against AWS but not in CI. You suggested deploy Nebari with admin token. Can override from Nebari CI config?

src/_nebari/stages/kubernetes_services/__init__.py
pt247: `{"cpu": "1", "memory": "4Gi"}` works in current CI.
If you assign more, it fails. My question is about deciding what is the appropriate size. And if it's okay to keep that decision as a part of this PR.

pt247:
I had to comment out the Cypress run to get the tests passing. 
https://github.com/nebari-dev/nebari/actions/runs/9082667704/job/24959793640?pr=2384

Adam-D-Lewis:
Hey, @pt247, do you feel that this PR is ready for review?  I see the autoscaling test looks like it's not passing at the moment, I believe?  Are you looking for a review at the moment?

tests/tests_deployment/test_conda_store_scaling.py
```diff
@@ -0,0 +1,234 @@
...
+import requests
+from kubernetes import client, config, dynamic
+
+from tests.tests_deployment import constants
+
+CONDA_STORE_API_ENDPOINT = "conda-store/api/v1"
+NEBARI_HOSTNAME = constants.NEBARI_HOSTNAME
+NAMESPACE = os.getenv("CONDA_STORE_SERVICE_NAMESPACE")
+TEST_CONDASTORE_WOKER_COUNT = os.getenv("TEST_CONDASTORE_WOKER_COUNT", 1)
+# NEBARI_HOSTNAME = "local.quansight.dev" ## Override for local testing
```
Adam-D-Lewis, line 19: Seems like this line should be removed, right?

tests/tests_deployment/test_conda_store_scaling.py
```diff
@@ -0,0 +1,234 @@
...
+@pytest.mark.timeout(10)
+def build_n_environments(n, builds, session):
+    log.info(f"Building {n} conda environments.")
+    for _ in range(n):
+        time.sleep(1)
+        builds.append(create_conda_store_env(session))
+    return builds
+
+
+# TODO : remove filters
```
Adam-D-Lewis, line 206: Any info on when these filters can be removed?  Consider adding a comment or link explaining why they are skipped for now.

tests/tests_deployment/test_jupyterhub_api.py
```diff
@@ -9,6 +9,7 @@
 
 
 @pytest.mark.filterwarnings("ignore::urllib3.exceptions.InsecureRequestWarning")
+@pytest.mark.filterwarnings("ignore::DeprecationWarning")
```
Adam-D-Lewis, line 12: Is this needed?  Seems like the deployment tests are passing on other branches without this

src/_nebari/stages/infrastructure/__init__.py
```diff
@@ -315,7 +315,7 @@ class GCPNodeGroup(schema.Base):
 
 
 DEFAULT_GCP_NODE_GROUPS = {
-    "general": GCPNodeGroup(instance="e2-highmem-4", min_nodes=1, max_nodes=1),
+    "general": GCPNodeGroup(instance="e2-highmem-4", min_nodes=1, max_nodes=5),
```
Adam-D-Lewis, line 318: Can you update Azure max nodes to 5 to be consistent with the other cloud providers?

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/conda-store/worker.tf
```diff
@@ -202,3 +206,71 @@ resource "kubernetes_deployment" "worker" {
...
+      scaleTargetRef = {
+        kind = "Deployment"
+        name = "nebari-conda-store-worker"
+      }
+      maxReplicaCount = var.max-worker-replica-count
+      pollingInterval = 5
+      cooldownPeriod  = 5
+      triggers = [
+        {
+          type = "postgresql"
```
Adam-D-Lewis, line 255: I wouldn't necessarily prevent merging b/c of this, but this would be more robust if we use the conda store api instead of querying the database directly (since the db design could change).  I haven't seen a way to run a curl or custom command for a trigger.  Are you familiar with any way?

Adam-D-Lewis: The metrics are exported to Prometheus so we could use the Prometheus trigger, but users can disable monitoring and we'd still want the scaling to work so not ideal.  I also see they have custom [external scalers]( https://keda.sh/docs/2.14/concepts/external-scalers/), but that seems like a pain to implement when this works as you have it.

Adam-D-Lewis (changes requested):
I don't think you've addressed any of my comments unless I'm missing something.

Adam-D-Lewis (changes requested):
I don't think you've addressed any of my comments unless I'm missing something.

Adam-D-Lewis:
Also, I'm running off a modified branch starting from yours so there could be some differences, but it appears that the conda-store workers won't actually run on more than 1 node due to the pvc that they are using being ReadWriteOnce.  They will just stay pending even if another node exists for them to scale to.
image
image

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/conda-store/worker.tf
pt247: Your comment about SQL got me thinking ... 
You are correct that SQL is not an exposed AP and, therefore, has no backward compatibility guarantees.
I found the solution in KEDA documentation about  [Metric API scaler](https://keda.sh/docs/2.12/scalers/metrics-api/).
keda-metric-api

So we need to 
1. give Keda credentials to hit
`https://nebari.quansight.dev/conda-store/api/v1/build/?status=QUEUED`
and 
`https://nebari.quansight.dev/conda-store/api/v1/build/?status=BUILDING`
2. and, modify this ScaledObject or to have 2 triggers one for each endpoint. 

Now, I just need to figure out how to grant permission to KEDA operator to access conda-store API.

tests/tests_deployment/test_jupyterhub_api.py
pt247: As it turns out these filters are used in few more deployment tests: eg [test_jupyterhub_ssh](https://github.com/nebari-dev/nebari/blob/749a1f14e2fff79bfc107f06dc7087d8727ddd5b/tests/tests_deployment/test_jupyterhub_ssh.py#L72), [test_dask_gateway](https://github.com/nebari-dev/nebari/blob/749a1f14e2fff79bfc107f06dc7087d8727ddd5b/tests/tests_deployment/test_dask_gateway.py#L25)
```
@pytest.mark.filterwarnings("ignore::urllib3.exceptions.InsecureRequestWarning")
@pytest.mark.filterwarnings("ignore::DeprecationWarning")
```
so I am simply removing the TODO.

tests/tests_deployment/test_conda_store_scaling.py
pt247: yes.

src/_nebari/stages/infrastructure/__init__.py
pt247: Done.

Adam-D-Lewis:
I've documented the scaling issue at [here](https://github.com/nebari-dev/nebari/issues/2505)

Adam-D-Lewis:
For this PR, I suggest we just use KEDA to only scale up to 4 workers as we do currently, but also have the conda-store workers scale down to 0 when not in use.  We need to test (manually) that conda envs nfs drive are still mounted to the user pod when no conda store workers are running.  I'm not sure if the nfs server needs to stay up or not.

pt247:
@Adam-D-Lewis 
> @pt247 For this PR, I suggest we just use KEDA to only scale up to 4 workers as we do currently, but also have the conda-store workers scale down to 0 when not in use to relieve the memory leak behavior seen [here](https://github.com/nebari-dev/nebari/issues/2418). We need to test (manually) that conda envs are still mounted to the user pod when no conda store workers are running. I'm not sure if that'll behave properly if the nfs server isn't running.

While doing a fresh deploy I came across a worker that won't scale down, it seems be trying to build one of the shared environments: 
Screenshot 2024-06-11 at 19 33 14

Logs show this warning:
```
│ nebari-conda-store-worker-5f5579fd9d-lgb9l conda-store-worker [2024-06-11 17:43:12,543: WARNING/ForkPoolWorker-2] /opt/conda/envs/co │
│ nda-store-server/lib/python3.10/site-packages/conda_store_server/build.py:353: UserWarning: Generating Docker images is currently no │
│ t supported, see https://github.com/conda-incubator/conda-store/issues/666                                                           │
│ nebari-conda-store-worker-5f5579fd9d-lgb9l conda-store-worker   warnings.warn(                                                       │
│ nebari-conda-store-worker-5f5579fd9d-lgb9l conda-store-worker                                                                        │
│ nebari-conda-store-worker-5f5579fd9d-lgb9l conda-store-worker [2024-06-11 17:43:12,544: INFO/ForkPoolWorker-2] Task task_build_conda │
│ _docker[build-2-docker] succeeded in 0.015599027999996906s: None                                                                     │
│ nebari-conda-store-worker-5f5579fd9d-lgb9l conda-store-worker [2024-06-11 17:43:15,151: INFO/ForkPoolWorker-2] Task task_build_conda │
│ _env_export[build-2-conda-env-export] succeeded in 2.6055338719997962s: None                                                         │
│ nebari-conda-store-worker-5f5579fd9d-lgb9l conda-store-worker [2024-06-11 17:43:15,151: INFO/MainProcess] Task task_update_storage_m │
│ etrics[build-2] received                                                                                                             │
│ nebari-conda-store-worker-5f5579fd9d-lgb9l conda-store-worker [2024-06-11 17:43:15,159: INFO/ForkPoolWorker-2] Task task_update_stor │
│ age_metrics[build-2] succeeded in 0.006750864999958139s: None  
```
It looks like KEDA won't help in this case, as the build never finishes.

pt247:
@Adam-D-Lewis I have reduced the default value of max conda-store workers to 4. User can still increase this number from the config. I can add some docs faq about it once this PR is in. Alternatively, I can simply remove that configuration and hard code it to 4. I would prefer the user had an option. What do you think?

Adam-D-Lewis:
> @Adam-D-Lewis
> 
> > @pt247 For this PR, I suggest we just use KEDA to only scale up to 4 workers as we do currently, but also have the conda-store workers scale down to 0 when not in use to relieve the memory leak behavior seen [here](https://github.com/nebari-dev/nebari/issues/2418). We need to test (manually) that conda envs are still mounted to the user pod when no conda store workers are running. I'm not sure if that'll behave properly if the nfs server isn't running.
> 
> While doing a fresh deploy I came across a worker that won't scale down, it seems be trying to build one of the shared environments: Screenshot 2024-06-11 at 19 33 14
> 
> Logs show this warning:
> 
> ```
> │ nebari-conda-store-worker-5f5579fd9d-lgb9l conda-store-worker [2024-06-11 17:43:12,543: WARNING/ForkPoolWorker-2] /opt/conda/envs/co │
> │ nda-store-server/lib/python3.10/site-packages/conda_store_server/build.py:353: UserWarning: Generating Docker images is currently no │
> │ t supported, see https://github.com/conda-incubator/conda-store/issues/666                                                           │
> │ nebari-conda-store-worker-5f5579fd9d-lgb9l conda-store-worker   warnings.warn(                                                       │
> │ nebari-conda-store-worker-5f5579fd9d-lgb9l conda-store-worker                                                                        │
> │ nebari-conda-store-worker-5f5579fd9d-lgb9l conda-store-worker [2024-06-11 17:43:12,544: INFO/ForkPoolWorker-2] Task task_build_conda │
> │ _docker[build-2-docker] succeeded in 0.015599027999996906s: None                                                                     │
> │ nebari-conda-store-worker-5f5579fd9d-lgb9l conda-store-worker [2024-06-11 17:43:15,151: INFO/ForkPoolWorker-2] Task task_build_conda │
> │ _env_export[build-2-conda-env-export] succeeded in 2.6055338719997962s: None                                                         │
> │ nebari-conda-store-worker-5f5579fd9d-lgb9l conda-store-worker [2024-06-11 17:43:15,151: INFO/MainProcess] Task task_update_storage_m │
> │ etrics[build-2] received                                                                                                             │
> │ nebari-conda-store-worker-5f5579fd9d-lgb9l conda-store-worker [2024-06-11 17:43:15,159: INFO/ForkPoolWorker-2] Task task_update_stor │
> │ age_metrics[build-2] succeeded in 0.006750864999958139s: None  
> ```
> 
> It looks like KEDA won't help in this case, as the build never finishes.

Are you sure the issue with a build never finishing is related.  I can't see how it's related.  What I'd like to see is spin down all the conda store workers after having built some envs, then start up a new jupyter user pod and run something using one of the built environments (while no conda store workers are running).

Adam-D-Lewis:
Update from our meeting: looks like we can't scale down to 0 conda store workers b/c the nfs server is on the same pod as the conda store worker container and the jupyter user pods will fail to run if the conda store nfs server isn't running.

pt247:
It turns out we can not scale conda-store workers to zero because it also hosts the nfs-share that is needed for any jupyter notebook to start. Keeping at least one worker alive could be a solution. Additionally we can not scale conda-store workers  beyond single node as the nfs share RWO and associated with PVC. As there is no advantage of getting in implimenting this PR, I am closing this PR.

Issue #2330, opened 2024-03-15
[ENH] - Move to JupyterHub 5.x (unreleased)


aktech:
### Feature description

We're using JupyterHub 4.0.1 right now and this is a proposal to upgrade JupyterHub from 4.x to 5.x (unreleased).

Steps:
- Version upgrade needs to be done in the jupyterhub docker image
https://github.com/nebari-dev/nebari-docker-images/blob/521251fb3be83e4ec96ee9a20c04c3f37d880723/jupyterhub/environment.yaml#L6
- Then we need to upgrade JupyterHub configuration in nebari core as well: https://github.com/nebari-dev/nebari/tree/366d1cd064d132477cd0e9c199fe655e40fa9d24/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub
- Need to make sure all the dependent components work with the latest JupyterHub like e.g. KubeSpawner

### Value and/or benefit

Since the last release of JupyterHub (4.0.2). JupyterHub added the ability to share user servers, which is a crucial feature we would like to harness in Nebari via jhub-apps. Upgrading Nebari to use Jupyterhub 5.x will help us implement app sharing in jhub-apps, for which Nebari is the primary consumer: https://github.com/nebari-dev/jhub-apps/issues/11

We're looking for this feature (which is not available in 4.x):
- https://github.com/jupyterhub/jupyterhub/pull/4594
- https://jupyterhub.readthedocs.io/en/latest/reference/rest-api.html#operation/post-shares-server

JupyterHub 5.x release might be around the corner, but I am not sure about the exact timeline right now. We will have to upgrade later anyways, upgrading now will unblock us from using sharing feature.

krassowski:
An RC for 5.0 might be available soon (next week). As a stepping stone we should migrate to 4.1 which is planned to be released very soon.

krassowski:
https://github.com/nebari-dev/nebari-docker-images/pull/129 is ready for review.

krassowski:
As a task for 5.0 migration, we should test setting `JUPYTERHUB_ALLOW_TOKEN_IN_URL=0` once we migrated to JupyterHub 4.1 (as this was added in 4.1 and will be a default in 5.0), see https://github.com/nebari-dev/nebari/issues/2351

krassowski:
[...]

PR #2427, opened 2024-04-27
Upgrade to JupyterHub 5.0

krassowski:
## Reference Issues or PRs

Closes https://github.com/nebari-dev/nebari/issues/2330

## Testing

- [ ] Did you add new tests?

## Any other comments?

This depends on https://github.com/nebari-dev/nebari-docker-images/pull/140.

### DIFF
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/02-spawner.py
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/02-spawner.py
@@ -73,6 +73,7 @@ if z2jh.get_config("custom.jhub-apps-enabled"):
                 "external": True,
             },
             "oauth_no_confirm": True,
+            "oauth_redirect_uri": "/services/:name/oauth_callback",
         }
 
     c.JupyterHub.services.extend(
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/02-spawner.py
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/02-spawner.py
@@ -71,9 +71,7 @@ if z2jh.get_config("custom.jhub-apps-enabled"):
                 "name": name,
                 "url": url,
                 "external": True,
-            },
-            "oauth_no_confirm": True,
-            "oauth_redirect_uri": "/services/:name/oauth_callback",
+            }
         }
 
     c.JupyterHub.services.extend(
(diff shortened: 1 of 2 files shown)


src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/02-spawner.py
```diff
@@ -73,6 +73,7 @@ def service_for_jhub_apps(name, url):
                 "external": True,
             },
             "oauth_no_confirm": True,
+            "oauth_redirect_uri": "/services/:name/oauth_callback",
```
minrk, line 76: I suspect the fix here is to _remove_ the `oauth_no_confirm` here, not add an oauth_redirect_uri, because these don't look like these are applications that oauth with JupyterHub. Is that right? The reason JupyterHub 5 raises with this configuration is that some configuration suggests that OAuth is desired, but insufficient configuration is provided for it to work. So it's rightly raising, but perhaps not informatively enough.

Do you perhaps have a suggestion for the error message that might have helped you better here?

1. authenticate users with JupyterHub as OAuth provider, or
2. communicate with the JupyterHub API via tokens issued under these service names?

Or is this perhaps using services config as a way to generate some links to things that aren't really Services at all, from JupyterHub's perspective?

krassowski: Thank you for chiming in with helpful suggestions!

> these don't look like these are applications that oauth with JupyterHub

@aktech do you know if this is right? I suspect that jhub-apps might in fact be doing these things, but I am not quite sure. Do you think we should keep/remove `oauth_no_confirm`?

aktech: Thanks again for the review @minrk 🎉 

>these don't look like these are applications that oauth with JupyterHub

Yes, they don't. They just use keycloak directly.

>Do you think we should keep/remove oauth_no_confirm?

Yes correct. I already removed them from example `jupyterhub_config.py` in jhub-apps: https://github.com/nebari-dev/jhub-apps/pull/220/files#diff-26e325937baab623ebecf944e3691cfc6d509853168cf154a403089ea6a93a1e

aktech:
Looks like doing a curl on JupyterHub gives `4.1.5`

https://github.com/nebari-dev/nebari/actions/runs/8968455291/job/24627891317?pr=2427#step:13:24

krassowski:
Could it be because the PR https://github.com/nebari-dev/nebari-docker-images/pull/140 on which this PR depends is not yet merged and not released?

aktech (approved):
Verified it works! LGTM

Issue #2452, opened 2024-05-07
[ENH] - Lower cost node sizes on GCP Nebari 


Adam-D-Lewis:
### Feature description

Note: Prices are at the time of writing and for the us-central-1 region.

The general node on GCP has 8 cpus and 52G of memory (n1-standard-8).  This node size costs (~$0.38/hr = ~$275/month), but the most of the time the cpus are doing very little and GCP has the ability to create custom node sizes at the cost of ~$25/cpu-month and ~$3/GB-month of RAM (up to 6.5GB/cpu) + ~$7/GB-month for additional RAM (beyond 6.5GB/cpu).

E.g. If we could get away with a 4 cpu 30GB of RAM node then the cost would be ~$205/month.

### Value and/or benefit

Lower fixed costs for using Nebari.  It's not clear to me what the ideal resources are for the node group, but it seems likely that we can reduce the cpu amount from 8 to 4 without ill effects.  According to Grafana, in the last 7 days, a cluster never  went beyond 1 cpu used, and the vast majority of the time needed less than 1 cpu.

![image](https://github.com/nebari-dev/nebari/assets/23342526/04b7bd26-4b46-4aee-96ca-583192f2643e)

Adam-D-Lewis:
Looking further, it seems easy to change the default node type to E2 instances on GCP and reduce costs by ~50%

PR #2453, opened 2024-05-07
Reduce GCP Fixed Costs by 50%

Adam-D-Lewis:
## Reference Issues or PRs
Fixes https://github.com/nebari-dev/nebari/issues/2452

- [ ] Documentation Update
- [ ] Build related changes
- [x] Other (please describe):  Make cost optimized E2 instances on GCP the default for node groups. 

## Testing

- [ ] Did you add new tests?

### DIFF
--- a/src/_nebari/stages/infrastructure/__init__.py
+++ b/src/_nebari/stages/infrastructure/__init__.py
@@ -314,9 +314,9 @@ class GCPNodeGroup(schema.Base):
 
 
 DEFAULT_GCP_NODE_GROUPS = {
-    "general": GCPNodeGroup(instance="n1-standard-8", min_nodes=1, max_nodes=1),
-    "user": GCPNodeGroup(instance="n1-standard-4", min_nodes=0, max_nodes=5),
-    "worker": GCPNodeGroup(instance="n1-standard-4", min_nodes=0, max_nodes=5),
+    "general": GCPNodeGroup(instance="e2-highmem-4", min_nodes=1, max_nodes=1),
+    "user": GCPNodeGroup(instance="e2-standard-4", min_nodes=0, max_nodes=5),
+    "worker": GCPNodeGroup(instance="e2-standard-4", min_nodes=0, max_nodes=5),
 }
 
 
--- a/src/_nebari/upgrade.py
+++ b/src/_nebari/upgrade.py
@@ -784,6 +784,53 @@ class Upgrade_2024_5_1(UpgradeStep):
         return config
 
 
+class Upgrade_2024_6_1(UpgradeStep):
+    version = "2024.6.1"
+
+    def _version_specific_upgrade(
+        self, config, start_version, config_filename: Path, *args, **kwargs
+    ):
+        if provider := config.get("provider", ""):
+            if provider == ProviderEnum.gcp.value:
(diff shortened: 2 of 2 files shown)


dcmcand (commented):
:question: Are we sure that the e2 is suitable for the user and worker node groups? It seems like for the worker group especially, they might handycap performance. Additionally, they do not support GPU's. I think it might be better to run the General node group on the e2-highmem-4 and the user and worker node groups on the n4-standard-4. Especially with them scaling down to zero now, I think that would be an acceptable tradeoff for performance vs price.

Adam-D-Lewis:
> ❓ Are we sure that the e2 is suitable for the user and worker node groups? It seems like for the worker group especially, they might handicap performance. Additionally, they do not support GPU's. I think it might be better to run the General node group on the e2-highmem-4 and the user and worker node groups on the n4-standard-4. Especially with them scaling down to zero now, I think that would be an acceptable tradeoff for performance vs price.

Great points, I looked into it a bit more.   

It looks like CPU performance (Coremark score) is roughly equal between the 2 types.

image
image

Also, for GPU instances, we usually create new node groups specifically for the gpu profiles although it is not fully documented at the moment e.g. (image below) so I don't see that as an issue for the user or worker node defaults since they would not use these default node groups for their gpu instances.
image

The only disadvantage I see is the maximum egress is down from 10 to 8 Gbps, but I believe the cost savings is worth the 20% reduction in bandwidth for most users though it's just a hunch.

image
image

Adam-D-Lewis:
@dcmcand any further concerns or comments?

dcmcand:
I still feel a bit of concern, specifically around the dask worker. However, I have no data to actually justify by concern.

If someone upgrades and then applies the new config, it will result in the nodes being replaced. Do we have any concerns about that?

I also feel like we should make sure we document this change and how to restore the original functionality. Maybe in the FAQ? But probably also in the release notes.

dcmcand (approved):

Adam-D-Lewis:
> I still feel a bit of concern, specifically around the dask worker. However, I have no data to actually justify by concern.
> If someone upgrades and then applies the new config, it will result in the nodes being replaced. Do we have any concerns about that?

We've added node types to the nebari config that is created when running `nebari init` so if people have node types in their config (as is the default), then they won't be affected by this change.  Assuming they don't, the nodes will be replaced.  This causes Nebari to be unusable for about ~15 minutes as the nodes are switched out, but shouldn't cause a problem otherwise.  I tested this on a deployment and it worked as expected.

> I also feel like we should make sure we document this change and how to restore the original functionality. Maybe in the FAQ? But probably also in the release notes.

I'll document it in the Nebari upgrade command so that users will be notified if this will affect them and what they need to add to their config so that it won't affect them, and we can copy something similar to the release notes.

dcmcand:
sounds good. thanks @Adam-D-Lewis

Adam-D-Lewis:
We don't know what the next Nebari version will be at the moment (2024.5.2 vs 2024.6.1) so I opened a separate PR and assigned it to the 2024.5.2 milestone. https://github.com/nebari-dev/nebari/pull/2466.  My thought is that we merge this as is and make sure to merge the other in with the appropriate version number during the next release step.

Issue #2401, opened 2024-04-15
[ENH] - Upgrade Grafana to latest


aktech:
### Feature description

We're using Grafana: v8.3.3 (30bb7a93ca) - https://github.com/grafana/grafana/releases/tag/v8.3.3

Latest Grafana: [10.4.2 (2024-04-10)](https://github.com/grafana/grafana/releases/tag/v10.4.2)

Upgrade to the Latest.

### Value and/or benefit

The current version is over 2 years old and the newer dashboard is much better and also has nicer and more intuitive UI for searching through Loki logs.

viniciusdc:
I think the main dragging point on this would be compatibility; if bumping its version does not break anything, then, at least to me, this looks like a reasonable step forward. Whoever checks this out needs to make sure this Prometheus support is compliant, as it will probably be updated as well.

kcpevey:
@marcelovilla any updates on this one?

marcelovilla:
@kcpevey I'm working on it. I have some changes in https://github.com/nebari-dev/nebari/compare/develop...upgrade-grafana that work for new deployments. However, I'm running into two issues:

* Our current Dashboards have a deprecation warning in the latest Grafana version because of https://grafana.com/docs/grafana/latest/developers/angular_deprecation/. I need to figure out the implications of this.
* When upgrading the chart for existing deployments, I'm running into the following error:
```
Error: error validating "": error validating data: ValidationError(Alertmanager.spec): unknown field "automountServiceAccountToken" in com.coreos.monitoring.v1.Alertmanager.spec
```

PR #2466, opened 2024-05-14
upgrade instructions for PR 2453

Adam-D-Lewis:
## Reference Issues or PRs
This is a companion PR to https://github.com/nebari-dev/nebari/pull/2453.  B/c we don't yet know for sure the next version of Nebari will be 2024.5.2, I've added this PR to be corrected and merged shortly before the next release of Nebari.

## What does this implement/fix?

_Put a `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds a feature)
- [ ] Breaking change (fix or feature that would cause existing features not to work as expected)
- [ ] Documentation Update
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Other (please describe):

## Testing

- [ ] Did you test the pull request locally?
- [ ] Did you add new tests?

PR #2472, opened 2024-05-17
Upgrade kube-prometheus-stack helm chart

marcelovilla:
## Reference Issues or PRs

Closes https://github.com/nebari-dev/nebari/issues/2401

- [ ] Documentation Update
- [ ] Build related changes
- [x] Other (please describe): upgrade helm chart

## Testing

- [ ] Did you add new tests?

## Any other comments?

Upgrading the `kube-prometheus-stack` chart involved adding Grafana OAuth credentials as a Kubernetes secret and passing them to the chart values using [variable expansion](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#variable-expansion).

While this upgrade will run for a fresh deployment, it will fail on existing deployments. I'll add the relevant steps that need to be run manually in order for the upgrade to work in https://github.com/nebari-dev/nebari/pull/2466

### DIFF
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/monitoring/main.tf
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/monitoring/main.tf
@@ -3,12 +3,24 @@ resource "random_password" "grafana_admin_password" {
   special = false
 }
 
+resource "kubernetes_secret" "grafana_oauth_secret" {
+  metadata {
+    name      = "grafana-oauth-secret"
+    namespace = var.namespace
+  }
+
+  data = {
+    "grafana-oauth-client-id"     = module.grafana-client-id.config.client_id
+    "grafana-oauth-client-secret" = module.grafana-client-id.config.client_secret
+  }
+}
+
 resource "helm_release" "prometheus-grafana" {
   name       = "nebari"
   namespace  = var.namespace
   repository = "https://prometheus-community.github.io/helm-charts"
   chart      = "kube-prometheus-stack"
-  version    = "30.1.0"
+  version    = "58.4.0"
 
   values = concat([
     file("${path.module}/values.yaml"),
@@ -206,6 +218,8 @@ resource "helm_release" "prometheus-grafana" {
           }
         }
 
+        envFromSecret = kubernetes_secret.grafana_oauth_secret.metadata[0].name
+
         "grafana.ini" : {
           server = {
             protocol            = "http"
@@ -222,8 +236,8 @@ resource "helm_release" "prometheus-grafana" {
             enabled                  = "true"
             name                     = "Login Keycloak"
             allow_sign_up            = "true"
-            client_id                = module.grafana-client-id.config.client_id
-            client_secret            = module.grafana-client-id.config.client_secret
+            client_id                = "$__env{grafana-oauth-client-id}"
+            client_secret            = "$__env{grafana-oauth-client-secret}"
             scopes                   = "profile"
             auth_url                 = module.grafana-client-id.config.authentication_url
             token_url                = module.grafana-client-id.config.token_url


Adam-D-Lewis:
If users enabled monitoring in the past, but have since disabled it, they will have issues when enabling monitoring after the upgrade.  Not sure the best course of action, maybe show the delete steps to all users? Maybe open an issue in the repo and put the solution there so users can find it when searching?

marcelovilla:
@Adam-D-Lewis I thought that disabling monioring means [terraform will remove that module](https://github.com/nebari-dev/nebari/blob/2bac8aae4e7755b1a3e0a2d56b58d4b52a86b81d/src/_nebari/stages/kubernetes_services/template/monitoring.tf#L7) (and hence all related resources in the cluster).

I haven't tested it myself but that would be my intuition. If that's not the case, I agree that we need to account for those users who have disabled it in the past and would be impacted in this upgrade.

Adam-D-Lewis:
> @Adam-D-Lewis I thought that disabling monitoring means [terraform will remove that module](https://github.com/nebari-dev/nebari/blob/2bac8aae4e7755b1a3e0a2d56b58d4b52a86b81d/src/_nebari/stages/kubernetes_services/template/monitoring.tf#L7) (and hence all related resources in the cluster).
> 
> I haven't tested it myself but that would be my intuition. If that's not the case, I agree that we need to account for those users who have disabled it in the past and would be impacted in this upgrade.

I'm pretty sure the CRDs aren't deleted even when you disable monitoring.  The reason helm behaves this way is because it's a cluster wide resource so other resources would be depending on that CRD.

Adam-D-Lewis:
I'm getting this error when trying to run the first upgrade command: 
```kubectl apply --server-side --force-conflicts -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.73.0/example/prometheus-operator-crd/monitoring.coreos.com_alertmanagerconfigs.yaml
error: error validating "https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.73.0/example/prometheus-operator-crd/monitoring.coreos.com_alertmanagerconfigs.yaml": error validating data: failed to download openapi: Get "https://45C28E7AE96C34F15D338C7DFB2B7A6A.gr7.us-west-2.eks.amazonaws.com/openapi/v2?timeout=32s": getting credentials: exec: executable aws not found

It looks like you are trying to use a client-go credential plugin that is not installed.
```

Adam-D-Lewis:
I would just expect us to delete the old CRDs with `kubectl` and then the new ones will get installed when you run `nebari deploy -c nebari-config`.  I'm not sure if there are any practical differences between that and doing it this way.  It seems like you've taken the commands from https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#from-50x-to-51x

marcelovilla:
> I'm pretty sure the CRDs aren't deleted even when you disable monitoring. The reason helm behaves this way is because it's a cluster wide resource so other resources would be depending on that CRD.

Makes sense. But even if the CRDs are not deleted, Terraform won't try to create that module so the upgrade will work. The problem is likely going to be if the user enables the monitoring module afterwards.

> I would just expect us to delete the old CRDs with kubectl and then the new ones will get installed when you run nebari deploy -c nebari-config. I'm not sure if there are any practical differences between that and doing it this way. It seems like you've taken the commands from https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#from-50x-to-51x

The commands are a combination of https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/README.md#from-39x-to-40x and https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/README.md#from-57x-to-58x.

When I first started working on this, I did try to remove the CRDs and upgrade but it did not work. But now that I think about it, it might have been because I did not remove the `daemonset` (as outlined in the first link). I can try again if you think it makes more sense to delete the CRDs instead of updating them (maybe it would for those people who have disabled monitoring but might potentially re-enable it in the future).

Adam-D-Lewis:
I think you're fine to keep the upgrade commands as you have them considering that's what's recommended by the helm chart.

Adam-D-Lewis:
> Makes sense. But even if the CRDs are not deleted, Terraform won't try to create that module so the upgrade will work. The problem is likely going to be if the user enables the monitoring module afterwards.

Yeah, I agree with you.

Adam-D-Lewis (approved):
LGTM

Issue #2485, opened 2024-05-27
[BUG] - IPv6 is unsupported/broken


zealousGnu:
### Describe the bug

When running on a single-stack IPv6 cluster (including NAT64 and DNS64), Nebari is broken. 

### Expected behavior

Nebari should work out of the box on a single-stack IPv6 cluster that has DNS64 and NAT64. Once the dependent sites are IPv6-compatible, it should work even without DNS64/NAT64.

### OS and architecture in which you are running Nebari

Debian 12 (bookworm) + k3s version v1.29.4+k3s1

### How to Reproduce the problem?

Run Nebari on a single-stack IPv6 cluster with DNS64 and NAT64. There are a bunch of issues. Some of them can be easily fixed. For example:

```
./src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/03-profiles.py: "image": "busybox:1.31"
```
The `busybox` image does not follow standard practices in regards to dual-stack environments. Changing all of the `busybox` images to `alpine` (which does follow proper networking practices) fixes several of the issues. 

There are a bunch of other places in the code that depend on IPv4 and as an admin, I can't fix them all.

### Compute environment

None

### Anything else?

15-20 years ago this would've been a feature request, but requiring legacy IP (aka IPv4) is a serious problem nowadays. Moreover it reflects badly on the project and makes it look amateurish.

Adam-D-Lewis:
Thanks for contributing @zealousGnu.  I'm not aware of any single-stack IPv6 Nebari deployments currently.  It would be a worthwhile to support this, but there aren't any resources dedicated to this at the moment.  I'll open an issue to add this limitation to the docs in the mean time. 

Thanks for pointing out the issues you encountered when trying a single-stack IPv6 deployment.  Any additional info on the issues encountered would be welcome, and additionally PRs fixing them would be welcome.  Tagging @dcmcand in case he has additional comments.

PR #2496, opened 2024-05-29
ipv6 compatible ingress checks

jorop:
## Reference Issues or PRs

Addresses #2485
For ipv6 working, at least an update of keycloak is also necessary.

Ingress checks will be successful on ipv6 also using [getaddrinfo](https://docs.python.org/3/library/socket.html#socket.getaddrinfo) instead of [gethostbyname](https://docs.python.org/3/library/socket.html#socket.gethostbyname).

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?


viniciusdc:
Hi @jorop, thank you for your contribution! We're currently in the midst of a significant update focused on our RBACs and JupyterHub, which might cause some delays in addressing your input. 

We discussed this in our internal sync yesterday, and I’d like to invite you to join our community sync meetings as well: https://www.nebari.dev/docs/community/#community-meetings. These meetings are open to everyone and provide a great platform for further discussion. Please note that this item is not yet on our roadmap because we are assessing the stability of specific components that support IPv6 exclusively, which would require some more thought into it...

In the meantime, for backward compatibility, it would be beneficial to include a custom field in the ingress section of the `nebari-config.yaml` file. This field would allow us to toggle between different implementations as needed.

viniciusdc:
@jorop I will be doing a check on this

viniciusdc:
Hi @jorop, sorry for the late feedback; thanks for submitting the PR! Regarding your statement about updating the keycloak, can you elaborate further on why you see that as a requirement? I assume it's due to it only formally supporting it on `>=20.x` but I also would like to hear if you encountered any other settle back while playing around with an ipv6 deployment

viniciusdc:
Can you add a test for `get_ip` as well?

marcelovilla:
@jorop thank you for your contribution! We really appreciate the effort you've put into this PR.

At the moment, full IPv6 support is not on our immediate roadmap. While your PR is a great first step in that direction, we believe there’s still a significant amount of work required to fully implement it. Given this, we’re hesitant to merge it right now without a clearer plan for future development. Would you be interested in continuing to work on this? If so, we'd be happy to discuss what a complete implementation might look like.

Additionally, we'd love to understand more about your use case. Are you running a setup where IPv6 support is essential? Or does your contribution provide value in isolation, even without full IPv6 support? Your insights would help us assess the impact of this change.

Issue #2456, opened 2024-05-09
[ENH] - Extension to present and pull a set of examples/gallery


krassowski:
### Feature description

Nebari provides `jupyterlab.initial_repositories` option which allows to pre-populate user space with content needed for them to start work, or to explore pre-defined examples:

https://github.com/nebari-dev/nebari/blob/7d0c28ef2815943f528f54541fd55e8db0c62fc9/src/_nebari/stages/kubernetes_services/__init__.py#L238

However, this requires cloning the entire set of repositories each time, is problematic to synchronise the repositories when they change, and does not provide the user with an interactive way to explore the repositories ahead of time.

We could create a JupyterLab extension that would present the examples, either in the sidebar or in the main area, together with rich metadata (icon/screenshot, title, description) and allow user to setup the example in the background.

### Value and/or benefit

- we would avoid space disk inflation due to cloning all the example repositories into the folder of each user
- users would be able to explore the available examples in a gallery
- the gallery could be governed by a separate `yaml`/`json` file that _could_ be updated independently of the deployment cycle
- in the future we could give users control over whether they want to sync the example repository with the latest upstream version (for example with `git stash`/`git pull`/`git stash pop` sequence)
- this extension could be co-maintained with broader Jupyter community as it would be also useful for the JupyterLab extensions examples
- extending the example setup process beyond the simple `git pull` would allow to also setup, or activate any required conda environments if associated with the example repository

### Anything else?

There are existing JupyterLab extensions which use git for cloning or fetching repository content:
- jupyterlab-git
- jupyterlab-github
- jupyterlab-gitlab

While there is an overlap with these extensions, the proposed gallery has specific presentation and user interface requirements which set it apart from the existing extensions; that said, we could leverage these extensions to simplify the backend implementation.

The extension could support PAT or other authentication mechanisms; the entire sequence of pulling (and optional auth) would happen in a jupyter-server extension, so the secretes would never be exposed to the end user.

viniciusdc:
I think that's an excellent idea, and while the current `superlab.initial_repositories` do overlap with the extension, I believe they aim to solve two different scenarios. So my only request on this would be that this extension complements this cloning project's workflow. Where as:
- The current available `initial_repositories` would be used by sys admins to pre-populate user-required instructions/data;
- The new extension would be used by users in their ecosystems/workflows whenever needed;

dharhas:
Also similar - https://github.com/jupyterhub/nbgitpuller

[...]

PR #2501, opened 2024-06-04
Implement support for jupyterlab-gallery config

krassowski:
## Reference Issues or PRs

- https://github.com/nebari-dev/nebari/issues/2456

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [x] Did you add new tests?

### DIFF
--- a/src/_nebari/stages/kubernetes_services/__init__.py
+++ b/src/_nebari/stages/kubernetes_services/__init__.py
@@ -246,6 +246,7 @@ class JupyterLabGalleryExhibit(schema.Base):
 class JupyterLabGallerySettings(schema.Base):
     title: str = "Examples"
     destination: str = "examples"
+    exhibits_url: Optional[str] = None
     exhibits: List[JupyterLabGalleryExhibit] = []
 
 
--- a/src/_nebari/stages/kubernetes_services/__init__.py
+++ b/src/_nebari/stages/kubernetes_services/__init__.py
@@ -235,8 +235,23 @@ class IdleCuller(schema.Base):
     server_shutdown_no_activity_timeout: int = 15
 
 
+class JupyterLabGalleryExhibit(schema.Base):
+    git: str
+    repository: str
+    title: str
+    description: str
+    icon: Optional[str] = None
+
+
+class JupyterLabGallerySettings(schema.Base):
+    title: str = "Examples"
+    destination: str = "examples"
+    exhibits: List[JupyterLabGalleryExhibit] = []
+
+
 class JupyterLab(schema.Base):
     default_settings: Dict[str, Any] = {}
+    gallery_settings: JupyterLabGallerySettings = JupyterLabGallerySettings()
     idle_culler: IdleCuller = IdleCuller()
     initial_repositories: List[Dict[str, str]] = []
     preferred_dir: Optional[str] = None
@@ -367,6 +382,9 @@ class JupyterhubInputVars(schema.Base):
     jupyterlab_image: ImageNameTag = Field(alias="jupyterlab-image")
     jupyterlab_default_settings: Dict[str, Any] = Field(
         alias="jupyterlab-default-settings"
+    )
+    jupyterlab_gallery_settings: JupyterLabGallerySettings = Field(
+        alias="jupyterlab-gallery-settings"
     )
     initial_repositories: str = Field(alias="initial-repositories")
     jupyterhub_overrides: List[str] = Field(alias="jupyterhub-overrides")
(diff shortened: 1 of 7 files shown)


kenafoster:
Any consideration yet into how we could handle authenticating to a Github/Gitlab server with this workflow?

As a workaround to date, we've put a PAT in the HTTPS URL.  But that would mean committing a PAT Into source code.  Might not be a HUGE problem if it's scoped to read-only on the needed repo and the nebari config is protected... still not ideal

krassowski:
Tested locally:

image

```yaml
jupyterlab:
  gallery_settings:
    title: Example repositories
    destination: examples
    exhibits:
      - title: Nebari
        git: https://github.com/nebari-dev/nebari.git
        repository: https://github.com/nebari-dev/nebari
        description: 🪴 Nebari - your open source data science platform
        icon: https://raw.githubusercontent.com/nebari-dev/nebari-design/main/logo-mark/horizontal/Nebari-Logo-Horizontal-Lockup-White-text.svg
      - title: PyTorch Tutorial
        git: https://github.com/yunjey/pytorch-tutorial.git
        repository: https://github.com/yunjey/pytorch-tutorial
        description: PyTorch Tutorial for Deep Learning Researchers
        icon: https://github.com/yunjey/pytorch-tutorial/raw/master/logo/pytorch_logo_2018.svg
default_images:
  jupyterlab: quay.io/nebari/nebari-jupyterlab:jupyterlab-gallery-db9c173-20240604
```

krassowski:
A limitation for now is that users can see contents `/etc/jupyter/`. If we want to put PATs in we should configure, at least for jupyter_gallery config this file to be not readable for users. Alternatively, we need to use a Python file (rather than JSON) and pass environment variables.

krassowski:
I attempted passing the `file_permission: "0400"` to the `local_sensitive_file` resource, but it is not used as a file at all, but just as a temporary variable.

Currently, all the config files get passed on to the of  JupyterHub Helm chart via [`custom`](https://z2jh.jupyter.org/en/latest/resources/reference.html#custom)`.extra-mounts` as kubernetes config maps:

https://github.com/nebari-dev/nebari/blob/749a1f14e2fff79bfc107f06dc7087d8727ddd5b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf#L85-L104

which then get populated by the custom logic during the spawn sequence. This logic also only works with k8s but this is not a problem for nebari.

The problem is that Kubernetes config map are designed to store non-confidential data as per [the documentation](https://kubernetes.io/docs/concepts/configuration/configmap/).

A possibly better solution would be to use [`singleuser.extraFiles`](https://z2jh.jupyter.org/en/latest/resources/reference.html#singleuser-extrafiles) as these allow to set the file `mode`, and is recommended for providing configuration.

krassowski:
I managed to use the dedicated k8 `secrets` via JupyterHub's `singleuser.extraFiles` but:
- there is a bug in k8s which means that read permissions from group cannot be removed: it has been open for years now and got lots of upvotes and two PRs but neither was merged, which is disappointing (https://github.com/kubernetes/kubernetes/issues/57923)
- the entire secretes `mode` handling in k8s seems to have many gotchas see https://github.com/kubernetes/kubernetes/issues/34982

krassowski:
The latest commit, [`b3a5bf9`](https://github.com/nebari-dev/nebari/pull/2501/commits/b3a5bf9e1aacac57cc89b3e2455e9ec31a752f35), includes a PoC implementation for setting the mode of config file to 0400. Unfortunately, this removes the access from server too, so the configs are not applied. We would need to spawn the server with elevated privilege as compared to the user who is accessing it, and later reduce them once creating kernels or terminals. This is possible on Linux by using [capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html) with `FOWNER` or `SETGID` but will need extensive testing to ensure that the elevated permissions are cleared for any process that user may spawn, and it might not be secure if we cannot scope it appropriately.

An alternative might be to either:
- pass the settings via an environmental variable and `unset` it immediately once consumed, or
- pass the settings as a config file and remove it immediately once the gallery extension starts.

For now we can instead revert to using the configuration files as before, and document that the PAT need to be scoped to the specific repositories that are used and to be read-only (which is a good practice anyways) and that it might be accessed by the user.

Adam-D-Lewis:
@krassowski It seems like it'd benefit from having docs.  Could you open a docs PR that explain how to use it or at a minimum point to the docs in https://github.com/nebari-dev/jupyterlab-gallery?

krassowski:
https://github.com/nebari-dev/nebari-docs/pull/478 is waiting for review over a week now.

Adam-D-Lewis:
Actually, is there a way to update the example repos?  If so and it's not obvious how to do so or if there is not a way to update the example repo could you add that to the docs?

Adam-D-Lewis (approved):

PR #2512, opened 2024-06-15
1792 Add docstrings to `upgrade.py`

### LINKED ISSUE #1792: [DOC] Add doc string to upgrade.py
### Context

A file level documentation string summarizing how the upgrade sequence works would be highly beneficial. 

### Value and/or benefit

There are some doc strings on individual methods, but devs have to do quite a bit of reading and digging around the different methods included before they can understand how it all works. A high level doc string would go a long way in helping out. 

### Anything else?

_No response_

arjxn-py:
Should fix #1792 

This PR adds file-level documentation strings for `upgrade.py`

- [x] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

### DIFF
--- a/src/_nebari/upgrade.py
+++ b/src/_nebari/upgrade.py
@@ -1,6 +1,6 @@
 """
 This file contains the upgrade logic for Nebari.
-Each release of Nebari requires an upgrade step class (which is a child class of UpgradeStep) to be created.
+Each release of Nebari requires an upgrade step class (which is a child class of UpgradeStep) to be created.  
 When a user runs `nebari upgrade  -c nebari-config.yaml`, then the do_upgrade function will then run through all required upgrade steps to bring the config file up to date with the current version of Nebari.
 """
 
@@ -12,7 +12,7 @@ import string
 import textwrap
 from abc import ABC
 from pathlib import Path
-from typing import Any, ClassVar, Dict
+from typing import Any, ClassVar, Dict, override
 
 import rich
 from packaging.version import Version
@@ -52,7 +52,7 @@ def do_upgrade(config_filename, attempt_fixes=False):
     validates the current version, and if necessary, upgrades the configuration
     to the latest version of Nebari.
 
-    Parameters:
+    Args:
     config_filename (str): The path to the configuration file.
     attempt_fixes (bool): Whether to attempt automatic fixes for validation errors.
 
@@ -114,7 +114,6 @@ class UpgradeStep(ABC):
         _steps (ClassVar[Dict[str, Any]]): Class variable holding registered upgrade steps.
         version (ClassVar[str]): The version of the upgrade step.
     """
-
     _steps: ClassVar[Dict[str, Any]] = {}
     version: ClassVar[str] = ""
 
@@ -149,7 +148,7 @@ class UpgradeStep(ABC):
         """
         Checks if there is an upgrade step for a given version.
 
-        Parameters:
+        Args:
             version (str): The version to check.
 
         Returns:
@@ -164,8 +163,8 @@ class UpgradeStep(ABC):
         """
         Runs through all required upgrade steps (i.e. relevant subclasses of UpgradeStep).
         Calls UpgradeStep.upgrade_step for each.
-
-        Parameters:
+        
+        Args:
             config (dict): The current configuration dictionary.
             start_version (str): The starting version of the configuration.
             finish_version (str): The target version for the configuration.
@@ -210,8 +209,6 @@ class UpgradeStep(ABC):
 
     def get_version(self):
         """
-        Returns the version of the upgrade step.
-
         Returns:
             str: The version of the upgrade step.
         """
@@ -239,8 +236,8 @@ class UpgradeStep(ABC):
 
         It should normally be left as-is for all upgrades. Use _version_specific_upgrade below
         for any actions that are only required for the particular upgrade you are creating.
-
-        Parameters:
+        
+        Args:
             config (dict): The current configuration dictionary.
             start_version (str): The starting version of the configuration.
             config_filename (str): The path to the configuration file.
@@ -369,7 +366,7 @@ class UpgradeStep(ABC):
 
         Override this method in subclasses if you need to do anything specific to your version.
 
-        Parameters:
+        Args:
             config (dict): The current configuration dictionary.
             start_version (str): The starting version of the configuration.
             config_filename (str): The path to the configuration file.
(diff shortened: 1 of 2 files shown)


viniciusdc:
Looking great, we will be reviewing this during the week, we're in the process of releasing, so if this does not enter in this current release, it will be in the next :rocket:

src/_nebari/upgrade.py
```diff
@@ -39,6 +45,20 @@
...
 
 def do_upgrade(config_filename, attempt_fixes=False):
+    """
+    Perform an upgrade of the Nebari configuration file.
+
+    This function loads the YAML configuration file, checks for deprecated keys,
+    validates the current version, and if necessary, upgrades the configuration
+    to the latest version of Nebari.
+
+    Parameters:
```
dcmcand, line 55: Hi @arjxn-py, in Nebari we use the Google style guide for docstrings (https://google.github.io/styleguide/pyguide.html#381-docstrings). That specifies using `Args` instead of `Parameters`

src/_nebari/upgrade.py
```diff
@@ -244,6 +320,7 @@ def set_nested_item(config: dict, config_path: list, value: str):
             config[config_path[-1]] = value
 
         def update_image_tag(config, config_path, current_image, new_version):
+            """Update the image tag in the configuration."""
```
dcmcand, line 323: This is missing the `Args` and `Returns` documentation.

src/_nebari/upgrade.py
```diff
@@ -314,13 +407,20 @@ def _version_specific_upgrade(
...
+    """
+
     version = "0.4.0"
 
     def _version_specific_upgrade(
         self, config, start_version, config_filename: Path, *args, **kwargs
     ):
         """
-        Upgrade to Keycloak.
+        This version of Nebari introduces Keycloak for authentication, removes deprecated fields,
```
dcmcand, line 422: These notes are good for all of the `_version_specific_upgrade_steps` but they should also be decorated with the `@override` decorator from `typing.override`. https://google.github.io/styleguide/pyguide.html#3831-overridden-methods

src/_nebari/upgrade.py
```diff
@@ -506,6 +618,12 @@ def _version_specific_upgrade(
 
 
 class Upgrade_2023_7_1(UpgradeStep):
+    """
+    Upgrade step for Nebari version 2023.7.1
```
dcmcand, line 622: I don't think these docstrings do much. I would suggest leaving them out unless there are specific notes for that particular upgrade step.

dcmcand (changes requested):
Hi @arjxn-py, thanks so much for your contribution. There are some minor style changes needed, but once that is done, we can get this merged!

src/_nebari/upgrade.py
arjxn-py: Have added this decorator on all of the `_version_specific_upgrade_steps` in [`52ea69e`](https://github.com/nebari-dev/nebari/pull/2512/commits/52ea69e511a1ed309f03931edf857dd0914483dc)

src/_nebari/upgrade.py
arjxn-py: I've switched to specify it as `Args` instead of `Parameters` in [`52ea69e`](https://github.com/nebari-dev/nebari/pull/2512/commits/52ea69e511a1ed309f03931edf857dd0914483dc)

src/_nebari/upgrade.py
arjxn-py: Have removed those docstrings having no specific notes

src/_nebari/upgrade.py
arjxn-py: I've added some additional docstrings in [`8f83443`](https://github.com/nebari-dev/nebari/pull/2512/commits/8f834431526654cccc52a1ac596d72d1aba9f063), Plus i've also added some typing to the args which had it missing

arjxn-py:
I'd like to know how can i run pre-commit hook correctly locally before committing my changes. 
I'm doing `pre-commit run` which is returning me - 
```python 
fix end of files.....................................(no files to check)Skipped
trim trailing whitespace.............................(no files to check)Skipped
check json...........................................(no files to check)Skipped
check yaml...........................................(no files to check)Skipped
check toml...........................................(no files to check)Skipped
check that executables have shebangs.................(no files to check)Skipped
codespell............................................(no files to check)Skipped
black................................................(no files to check)Skipped
ruff.................................................(no files to check)Skipped
isort................................................(no files to check)Skipped
Terraform fmt........................................(no files to check)Skipped
```

arjxn-py:
In addition to that, I see that `python==3.10` & `python==3.11` tests are failing because the `@override` decorator is actually available in the typing module starting from Python 3.12. See also - [https://docs.python.org/3/whatsnew/3.12.html#pep-698-override-decorator-for-static-typing](https://docs.python.org/3/whatsnew/3.12.html#pep-698-override-decorator-for-static-typing)

So should I use `typing_extensions` instead?

dcmcand:
> In addition to that, I see that `python==3.10` & `python==3.11` tests are failing because the `@override` decorator is actually available in the typing module starting from Python 3.12. See also - https://docs.python.org/3/whatsnew/3.12.html#pep-698-override-decorator-for-static-typing
> 
> So should I use `typing_extensions` instead?

ah, good catch, yes, use `typing_extensions`

dcmcand:
> Thanks for the constructive review @dcmcand, i've made the suggested changes, hope I interpreted it correctly. I'd like to know how can i run pre-commit hook correctly locally before committing my changes. I'm doing `pre-commit run` which is returning me -
> 
> ```python
> fix end of files.....................................(no files to check)Skipped
> trim trailing whitespace.............................(no files to check)Skipped
> check json...........................................(no files to check)Skipped
> check yaml...........................................(no files to check)Skipped
> check toml...........................................(no files to check)Skipped
> check that executables have shebangs.................(no files to check)Skipped
> codespell............................................(no files to check)Skipped
> black................................................(no files to check)Skipped
> ruff.................................................(no files to check)Skipped
> isort................................................(no files to check)Skipped
> Terraform fmt........................................(no files to check)Skipped
> ```

You should be able to run it. Use `pre-commit run --all-files`

arjxn-py:
> You should be able to run it. Use `pre-commit run --all-files`

Yes, thanks it works 🚀

arjxn-py:
Oh Conda build failing, I just noticed that conda has the latest release of version `4.11.0` for `typing_extensions`, on it to fix it.

dcmcand (approved):
Thanks for the quality contribution @arjxn-py :rocket:

Issue #2534, opened 2024-06-21
Investigate Ceph for use in Nebari


Adam-D-Lewis:
### Context

Rook Ceph might help with Conda store worker scaling issues caused by slow NFS as mentioned in https://github.com/nebari-dev/nebari/issues/2505.  It also might help with making the user home directory more performant and give us an option of replacing the minio pods with ceph's object storage as well.

I've successfully deployed the [rook operator helm chart](https://rook.io/docs/rook/v1.11/Helm-Charts/ceph-cluster-chart/) on an AKS cluster with Nebari and created a shared file system between 2 pods on different nodes.  The steps were roughly as follows:
- Install Rook helm chart on cluster
- create a [CephCluster object](https://github.com/rook/rook/blob/master/deploy/examples/cluster-on-pvc.yaml)
- create a CephFileSystem and CephFilesystemSubVolumeGroup (in same namespace as your cluster I think)
- create a StorageClass referencing the CephFileSystem
- create a PVC which references the StorageClass (any namespace)
- Create a pod/deployment/etc. that mounts the pvc

We'll need at least 3 nodes running for ceph to take advantage of high availability and resiliency features of Ceph.  We should be able to make them smaller than the current general node.  We may want to create a separate node group for just the ceph storage with 3 small-ish nodes.  I think min node size for Ceph would be 2vcpu, 4Gi RAM per node.

The storage class names we'll use to back Ceph storage will likely be different on each cloud provider.  This should also be tested on the local deployment.

Eventually, it'd be nice to set up auto expansion of storage as space fills up.  It seems to be possible on some cloud providers at least according to [these docs](https://rook.github.io/docs/rook/latest-release/Storage-Configuration/Advanced/ceph-configuration/?h=auto+expansion#prerequisites-for-auto-expansion-of-osds).

Adam-D-Lewis:
A few issues encountered:
- Ceph let's you make a PVC larger than you have storage backing it which is a little confusing if you aren't aware of that.
- I've had a hard time deleteing CephFilesystem objects.  They tend to linger.

Adam-D-Lewis:
I tested the PR linked to this issue and ran the following and it took 10-17 seconds on that Nebari deployment while it took 1 minute on a standard GCP deployment (no Ceph).
```python
import os

# Define the directory where the files will be created
dir_path = "/home/ad/fio-test"
# dir_path = "/tmp/fio-test"

# Create the directory if it doesn't exist
[...]

PR #2541, opened 2024-06-25
Use Rook Ceph for Jupyterhub and Conda Store drives

Adam-D-Lewis:
## Reference Issues or PRs
Related to https://github.com/nebari-dev/nebari/issues/2534
Deploys rook operator and rook-cluster helm charts, sets up some StorageClasses backed by a Rook Ceph cluster and uses those storage classes for what were previously the jupyterhub and conda store NFS drives.

Developed on Azure at the moment.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?


Adam-D-Lewis:
See TODOs in code for what is still left

src/_nebari/stages/infrastructure/__init__.py
```diff
@@ -211,22 +238,21 @@ class KeyValueDict(schema.Base):
...
 
-    instance: str
-    min_nodes: Annotated[int, Field(ge=1)] = 1
-    max_nodes: Annotated[int, Field(ge=1)] = 1
+    pass
 
 
 DEFAULT_DO_NODE_GROUPS = {
-    "general": DigitalOceanNodeGroup(instance="g-8vcpu-32gb", min_nodes=1, max_nodes=1),
+    "general": DigitalOceanNodeGroup(instance="g-8vcpu-32gb", min_nodes=1, max_nodes=5),
```
Adam-D-Lewis, line 252: No need to keep max_nodes as 5

Adam-D-Lewis:
One issue I'm hitting is the Ceph cluster deployment takes a long time but helm thinks it's done b/c it just deploys a CephCluster object which is quick and the ceph operator handles the longer task of waiting for a bunch of rook reosurces to be created.  

What this means in practice currently is that the deploy step times out, and ceph isn't ready for another 5-10 minutes even beyond the timeout.

I'm not sure the best way to address this.  Maybe some sort of readiness check can be added or increase the timeout or deploy ceph on an earlier stage to give it the extra time to get ready.

Adam-D-Lewis:
Currently, there are 3 storage nodes that require about 1cpu/2-6 GB of RAM.  On Azure, I added 3 2cpu/8GB Ram nodes for storage adding $210/month to the operating cost of Nebari.  We may be able to use a single node by using [cluster-test.yaml](https://github.com/rook/rook/blob/release-1.14/deploy/examples/cluster-test.yaml) from [here](https://rook.io/docs/rook/latest-release/Getting-Started/example-configurations/#operator), but we likely won't get the same performance speedup and would not have the redundancy benefit that a more typical multi-node ceph deployment can give.

Adam-D-Lewis:
B/c there isn't a pressing need for this, and the benefits of using ceph seem to only come with significant increased cost, I'm planning to put this on the backburner for the moment.

dcmcand:
I think we can make Ceph the first stage after the k8s cluster (so stage 3?) and we can see if that is enough.

As for costs, we could make ceph single node the default and have a HA flag in the config to move to multnode, HA setup. That way we still have the abstracted storage interface, but not the increased cost.

@Adam-D-Lewis can you test with a single node ceph and see how it runs?

Adam-D-Lewis:
Yeah, I'll test with a single node

Adam-D-Lewis:
While single node ceph seems to be working in my limited use, I'm not sure I'd be comfortable making it the default initially since a single node ceph deployment is not recommended for production use in the Rook docs.  One thing I want to check is that we understand what happens if the general node is restarted.  I also think we should use it in a test deployment for a while to ensure it's working as expected.  

Currently, we use EFS on AWS deployments, NFS on all other deployments.  I suggest we add the option to specify in the config whether or not we want to use EFS, NFS, or Ceph for the jupyterlab user and shared directories and for the conda store shared directories.  Perhaps we add an option under the current `storage` section as listed below.

Currently the storage section looks like the following by default.
```yaml
storage:
  conda_store: 200Gi
  shared_filesystem: 200Gi
```

and I think we should change it to 

```yaml
storage:
  conda_store:
    type: "NFS"  # EFS, NFS, or Ceph
    size: 200Gi
  shared_filesystem:
    type: "NFS"  # EFS, NFS, or Ceph
    size: 200Gi
```

Adam-D-Lewis:
Okay, so my plan for this PR is to modify the `nebari-config.yaml` as shown here
```
storage:
  type: "Ceph"  # NFS (default) or Ceph
    storage_class_name: "default based on provider used"
  conda_store: 200Gi
  shared_filesystem: 200Gi
```

Then, you can choose Ceph or NFS on the initial deployment.  I'll try to find a way to throw a warning if people try to change the value or I'll disallow it and say you have to do a new deployment to change that value.  Ceph probably won't be accessible on AWS at first since we use EFS on AWS not our own NFS.  I should be able to make ceph work on local, gcp, azure, and existing deployments.

src/_nebari/stages/infrastructure/template/azure/modules/kubernetes/main.tf
```diff
@@ -53,39 +54,30 @@ resource "azurerm_kubernetes_cluster" "main" {
...
-  max_count             = var.node_groups[2].max_size
+  min_count             = each.value.min_size
+  max_count             = each.value.max_size
   max_pods              = var.max_pods
   node_labels = {
-    "azure-node-pool" = var.node_groups[2].name
+    "azure-node-pool" = each.value.name
   }
   orchestrator_version = var.kubernetes_version
   tags                 = var.tags
```
Adam-D-Lewis, line 83: Fly by fix allowing additional node groups to be defined on Azure

Adam-D-Lewis:
I'm working on getting this deployed on GKE.  I had to add 
```
apiVersion: v1
kind: ResourceQuota
metadata:
  # annotations:
  labels:
    addonmanager.kubernetes.io/mode: Reconcile
  name: rook-critical-pods
  namespace: rook-ceph
spec:
  hard:
    pods: 1G
  scopeSelector:
    matchExpressions:
    - operator: In
      scopeName: PriorityClass
      values:
      - system-node-critical
      - system-cluster-critical
```

that helped get some pods deployed, but now I'm facing a new issue that the ceph pvcs won't mount on the jupyterhub and dask gateway pods.  Trying to resolve.

Adam-D-Lewis:
Okay, so everything deploys, but conda store workers aren't working correctly yet so I'll tackle that next.

Adam-D-Lewis:
Conda envs are kind of working now.  I noticed spinning up a new jupyterlab server took quite a bit longer, but I'm hoping it's just an issue right after the cluster is first created.  Also, the ceph cluster failed the first time I think due to some race condition.  I may need to spend some time ensuring terraform resources are created in the correct order.

Adam-D-Lewis:
Things are looking good on the deployment.  
- I should test deploying NFS and make sure that still works.

Ideally, nebari destroy needs a bit more work.  The problem is that ceph operator pod gets deleted before it has finished deleting these resources.  This caused deleting the rook-ceph namespace to lag a bit.  After a few minutes, the namespace and those resources were deleted anyway.
```
NAME                                DATA   AGE
configmap/rook-ceph-mon-endpoints   5      16m
NAME                   TYPE                 DATA   AGE
secret/rook-ceph-mon   kubernetes.io/rook   4      16m
NAME                                 DATADIRHOSTPATH   MONCOUNT   AGE   PHASE      MESSAGE                    HEALTH      EXTERNAL   FSID
cephcluster.ceph.rook.io/rook-ceph   /var/lib/rook     1          17m   Deleting   Deleting the CephCluster   HEALTH_OK              c53dbc2f-c8c2-42e0-85aa-221ca7133b57
NAME                                          ACTIVEMDS   AGE   PHASE
cephfilesystem.ceph.rook.io/ceph-filesystem   1           17m   Ready
NAME                                                            PHASE   FILESYSTEM        QUOTA   AGE
cephfilesystemsubvolumegroup.ceph.rook.io/ceph-filesystem-csi   Ready   ceph-filesystem           17m
```

Adam-D-Lewis:
I'm currently hitting an issue.

(dependencies on the left)
module.jupyterhub-nfs-mount -> module.jupyterhub
module.jupyterhub-cephfs-mount -> module.jupyterhub

but when I switch from nfs to cephfs (or vice versa), the pvc created in the nfs and cephfs-mount modules, can't be destroyed b/c it is used by the hub pod from the jupyterhub module.

The solution is to trigger some of the jupyterhub module resources to be recreated when the nfs and cephfs mount modules are changed perhaps by using [replace_triggered_by](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#replace_triggered_by).

Adam-D-Lewis:
Going from NFS to Ceph works, but when going from Ceph to NFS, the nebari-conda-store-storage PVC is created, but it won't bind until a consumer of the PVC is created (conda-store-worker), but conda-store-worker is waiting for the PVC to finish being created before it will recreate itself with the different settings.  if we trigger based on conda-store-fs then I think it'll work, but not quite as robust.

Adam-D-Lewis:
I think I've made all the changes necessary for ceph to run on AWS, but I haven't tested it.

Adam-D-Lewis:
Remaining issues:
- [ ] Restart the general node and see what happens
- [ ] Nebari destroy fails b/c the operator is killed before it's finished cleaning up the resources
- [ ] Clean up the code a bit (remove comments, etc.)

src/_nebari/stages/kubernetes_services/template/conda-store.tf
```diff
@@ -61,17 +61,13 @@ module "kubernetes-conda-store-server" {
   services       = var.conda-store-service-token-scopes
   extra-settings = var.conda-store-extra-settings
   extra-config   = var.conda-store-extra-config
-}
-
-module "conda-store-nfs-mount" {
```
Adam-D-Lewis, line 66: This was moved inside the conda store module

Adam-D-Lewis:
I think the PR is ready for review.  You can do a deployment by adding storage.type = cephfs to the nebari config file.  You shouldn't switch an existing nebari deployment from nfs (default) to cephfs b/c you'll lose all your files and conda envs.  I'll write up a docs PR.  I think we should start using this on Quansight's deployment (after we do a backup) to get a feel for it, and eventually make it the default if all goes well.

viniciusdc:
@Adam-D-Lewis did you workout the destroy bits?

Adam-D-Lewis:
> @Adam-D-Lewis did you workout the destroy bits?

No, not yet.  The destroy still occurs, but it does take longer at the moment, and a message saying stage 07 had problems destroying, but only stages 01 and 02 are usually necessary for a successful destroy.

Adam-D-Lewis:
I deployed on Azure today and timed out.  Redeploying fixed the issue.  I'll work on fixing that, but I'd appreciate an initial review even before then.

Adam-D-Lewis:
Docs PR - https://github.com/nebari-dev/nebari-docs/pull/493
[deploy-preview-493--nebari-docs.netlify.app](https://deploy-preview-493--nebari-docs.netlify.app/)

Adam-D-Lewis:
I attempted to just have the operator wait before it was destroyed with a local-exec provisioner, but that didn't work.  There are all kinds of protections stopping the filesystem from being deleted.  Attempting to delete the cephcluster results in a message along the lines of 
"failed to reconcile CephCluster "prod/prod". failed to clean up CephCluster "prod/prod": failed until dependent objects delted CephFileSystemSubVolumeGroup and CephFileSystem."

The only way I could delete the CephFileSystemSubVolumeGroup and CephFileSystem was through manually deleting the finalizers on them.  I tried adding the rook.io/force-deletion="true" annotation, but that didn't work.  I also tried setting the cephClusterSpec.cephConfig.cleanupPolicy.confirmation = "yes-really-destroy-data" in the cephcluster helm chart values, but that didn't seem to have an effect either.

Even after deleting the CephFileSystemSubVolumeGroup and CephFileSystem via the finalizers, the Cephcluster still won't delete.  The following logs are in the operator.
"failed to reconcile CephCluster "prod/prod". failed to clean up CephCluster "prod/prod": failed to check if volumes exist for CephCluster in namespace "prod": waiting for csi volume attachments in cluster "prod" to be cleaned up".

Overriding the Cephcluster finalizers also deleted it. So worse case, we could override the finalizers on those 3 objects in the destroy step of Nebari stage 07 and then nebari destory would succeed, but it's a bit hacky, but doesn't matter much since we would only do so when we are deleting the cluster anyway.

dcmcand:
@Adam-D-Lewis can you merge the lastest develop in?

dcmcand (commented):
@Adam-D-Lewis I think this is great work.

Some questions though:

1. Are you planning on a HA multiple ceph node option for this in the future? Is there a follow on ticket for that?
2. Is there a ticket for moving node storage to ceph? 
3. It appears this is using dynamic pvc provisioning for Ceph which would allow expanding conda-store volumes in the future, correct?
4. I didn't see any validation or data loss warning for changing an existing nfs deployment to ceph or vise versa. I think we will definitely need that, is there a follow on ticket for that?
5. Let's spend some time thinking about how to migrate from nfs to ceph and back
6. Let's spend some time thinking about how to use this same design pattern for nfs with dynamic provisioning, etc

Adam-D-Lewis:
> @Adam-D-Lewis I think this is great work.
> 
> Some questions though:
> 
> 1. Are you planning on a HA multiple ceph node option for this in the future? Is there a follow on ticket for that?

This PR provides feature parity with the existing NFS storage state in Nebari.  I'm happy to work on allowing the use of additional nodes to Ceph assuming that work remains in scope for JATIC.  I have created a follow on ticket to make the number of RookCeph backing nodes variable - https://github.com/nebari-dev/nebari/issues/2573.  I assumed we would validate the Ceph storage works and then make it the default for future deployments, and only after that point would we work on increasing the number of rook/ceph nodes.

> 2. Is there a ticket for moving node storage to ceph?

I don't have a ticket for this.  I'm not sure how we would do this, and I can't think of any strong benefits.  Can you explain your thoughts further and/or what you're looking for here?

> 3. It appears this is using dynamic pvc provisioning for Ceph which would allow expanding conda-store volumes in the future, correct?

This does use dynamic provisioning of PVs since PVCs are provisioned from storage classes.  I believe the PVC should be able to increase in size without data loss, but I'll test and get back to you.  I'm not sure how increasing the size of the disks that back the RookCeph storage would work.  I'll look into that a bit more and get back to you as well.

> 4. I didn't see any validation or data loss warning for changing an existing nfs deployment to ceph or vise versa. I think we will definitely need that, is there a follow on ticket for that?

Currently, the only warning is in the [documentation PR](https://github.com/nebari-dev/nebari-docs/pull/493/files#diff-b97453b2d484370f2f98c5b1b75cebed728366a91370339f38747d5cd0c761bcR421).  We don't have a way to recognize changes to the nebari config file during a redeploy.  I did open [this ticket](https://github.com/nebari-dev/nebari/issues/2607) which explains the issue further and proposes a way to add this ability.  I wasn't planning on working on that before this PR is merged.  As a workaround we could require user confirmation on any deploy which has cephfs set as the storage, but I think it's fair to just put the warning in the docs as well as listing it in the docs as an alpha feature.

> 5. Let's spend some time thinking about how to migrate from nfs to ceph and back

I think it's fair to not support that transition, at least initially while we prove that RookCeph is going to work well.  So in that case, users would only be able to make the decision on the initial deploy and not change it afterwards.  I do think we should run this on the Quansight deployment to help test it though.   In a case like that, I would tell an admin to copy all the files in the NFS drives, tar them and copy to cloud storage.  Then copy them to the new disk and untar them after we redeploy similar to what we do in the [manual backup docs](https://www.nebari.dev/docs/how-tos/manual-backup/).  We should test this before doing it (particularly for the conda-store NFS drive), but I think it should work fine.

> 6. Let's spend some time thinking about how to use this same design pattern for nfs with dynamic provisioning, etc

Can you elaborate here?  What else are you looking for?

dcmcand:
> > 1. Are you planning on a HA multiple ceph node option for this in the future? Is there a follow on ticket for that?
> 
> This PR provides feature parity with the existing NFS storage state in Nebari. I'm happy to work on allowing the use of additional nodes to Ceph assuming that work remains in scope for JATIC. I have created a follow on ticket to make the number of RookCeph backing nodes variable - #2573. I assumed we would validate that Ceph storage works for a single node and then make it the default for future deployments, and only after that point would we work on increasing the number of rook/ceph nodes.
> 
:+1: 
> > 2. Is there a ticket for moving node storage to ceph?
> 
> I don't have a ticket for this. I'm not sure how we would do this, and I can't think of any strong benefits. Can you explain your thoughts further and/or what you're looking for here?
> 
The benefit would be allowing the general node to spin up in a different AZ to allow multi AZ deployments. That has caused a number of folks errors when they upgraded. 
> > 3. It appears this is using dynamic pvc provisioning for Ceph which would allow expanding conda-store volumes in the future, correct?
> 
> This does use dynamic provisioning of PVs since PVCs are provisioned from storage classes. I believe the PVC should be able to increase in size without data loss, but I'll test and get back to you. I'm not sure how increasing the size of the disks that back the RookCeph storage would work. I'll look into that a bit more and get back to you as well.
> 

:+1: 

> UPDATE: I tested increasing the shared directory and the conda storage sizes on a GCP deployment. The PVC for the disk that backs the RookCeph storage was increased without a problem (verified in GCP Console as well) and the PVCs for the shared directory and conda env storage also increased without a problem.
> 
> > 4. I didn't see any validation or data loss warning for changing an existing nfs deployment to ceph or vise versa. I think we will definitely need that, is there a follow on ticket for that?
> 
> Currently, the only warning is in the [documentation PR](https://github.com/nebari-dev/nebari-docs/pull/493/files#diff-b97453b2d484370f2f98c5b1b75cebed728366a91370339f38747d5cd0c761bcR421). We don't have a way to recognize **changes** to the nebari config file during a redeploy which makes this difficult. I did open [this ticket](https://github.com/nebari-dev/nebari/issues/2607) which explains the issue further and proposes a way to add this ability. I wasn't originally planning on working on that before this PR is merged, but I'd be happy to if you think it's needed. As a workaround we could require user confirmation on any deploy which has cephfs set as the storage, but I think it's fair to just put the warning in the docs as well as listing it in the docs as an alpha feature.
> 

I think that is fine for an alpha feature, but given the risk of user data loss, I think this is a problem we need to solve before we go mainstream with ceph, either by figuring out how to migrate, or by blocking switching

> > 5. Let's spend some time thinking about how to migrate from nfs to ceph and back
> 
> I think it's fair to not support that transition, at least initially while we prove that RookCeph is going to work well. So in that case, users would only be expected to make the decision on the initial deployment and not change it afterwards. I do think we should run this on the Quansight deployment to help test it though. In a case like that, I would tell an admin to copy all the files in the NFS drives, tar them and copy to cloud storage. Then copy them to the new disk and untar them after we redeploy similar to what we do in the [manual backup docs](https://www.nebari.dev/docs/how-tos/manual-backup/). We should test this before doing it (particularly for the conda-store NFS drive), but I think it should work fine.
> 

I think that is fair now, but I would prefer to have a switching path in the future, even if it is to use backup and restore.

> > 6. Let's spend some time thinking about how to use this same design pattern for nfs with dynamic provisioning, etc
> 
> Can you elaborate here? What else are you looking for?

Basically what I talked about in the meeting, removing dedicated PV's and instead creating an NFS storage class that dynamically provisions for each cloud provider. That way nfs wouuld be exandable for all providers in the future

Adam-D-Lewis:
> > > 2. Is there a ticket for moving node storage to ceph?
> > 
> > 
> > I don't have a ticket for this. I'm not sure how we would do this, and I can't think of any strong benefits. Can you explain your thoughts further and/or what you're looking for here?
> 
> ~The benefit would be allowing the general node to spin up in a different AZ to allow multi AZ deployments. That has caused a number of folks errors when they upgraded.~ I misspoke when I said node storage, because I was remembering the issue incorrectly. Basically any PVC's made by pods in the cluster would need to move to ceph. Right now we have at least one PVC against ebs which can't cross AZ's

[Here](https://github.com/nebari-dev/nebari/issues/2633) is the issue about moving other PVCs to use ceph for storage.

> > > 5. Let's spend some time thinking about how to migrate from nfs to ceph and back
> > 
> > I think it's fair to not support that transition, at least initially while we prove that RookCeph is going to work well. So in that case, users would only be expected to make the decision on the initial deployment and not change it afterwards. I do think we should run this on the Quansight deployment to help test it though. In a case like that, I would tell an admin to copy all the files in the NFS drives, tar them and copy to cloud storage. Then copy them to the new disk and untar them after we redeploy similar to what we do in the [manual backup docs](https://www.nebari.dev/docs/how-tos/manual-backup/). We should test this before doing it (particularly for the conda-store NFS drive), but I think it should work fine.
> 
> I think that is fair now, but I would prefer to have a switching path in the future, even if it is to use backup and restore.

Yeah, I'm happy for backup and restore to be the switching plan.  I'll test this out after this is merged since I'll want us to test this on the Quansight Nebari deployment after its merged.

> > > 6. Let's spend some time thinking about how to use this same design pattern for nfs with dynamic provisioning, etc
> > 
> > 
> > Can you elaborate here? What else are you looking for?
> 
> Basically what I talked about in the meeting, removing dedicated PV's and instead creating an NFS storage class that dynamically provisions for each cloud provider. That way nfs wouuld be exandable for all providers in the future

I doubt it would be very hard to make that change, and I'd be happy to comment on an issue with more thoughts.

Adam-D-Lewis:
Thanks again for the review, @dcmcand.  Is there anything else you want to see before this is merged?

dcmcand (approved):
:tada:

Adam-D-Lewis:
The failing test in the Local Integration Tests has been seen in other PRs.  @viniciusdc is working on a fix, and it is unrelated to this PR, so I will merge this PR.

PR #2546, opened 2024-06-27
Add a default roles at initialisation

### LINKED ISSUE #2536: [ENH] - Add a default keycloak role to allow app sharing for convenience
### Feature description

We recently added ability to configure fine-grained permissions for JupyterHub via keycloak here: https://github.com/nebari-dev/nebari/pull/2471

Add a default keycloak role in the jupyterhub client to be allow app sharing.

Screenshot 2024-06-24 at 5 39 32 pm

Screenshot 2024-06-24 at 5 39 42 pm

Scopes: `shares!user,read:users:name,read:groups:name`

### Value and/or benefit

Users can create it themselves as well, but would be nice to provide this auto-created, so that users can directly assign this role to users to allow them to share apps via jhub-apps in Nebari.

### Anything else?

_No response_

aktech:
## Reference Issues or PRs

Fixes #2536 

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

### DIFF
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
@@ -283,7 +283,7 @@ module "jupyterhub-openid-client" {
     {
       "name" : "allow-app-sharing-role",
       "description" : "Allow app sharing for apps created via JupyterHub App Launcher (jhub-apps)",
-      groups : [],
+      "groups" : [],
       "attributes" : {
         # grants permissions to share server
         # grants permissions to read other user's names
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
@@ -286,6 +286,8 @@ module "jupyterhub-openid-client" {
   jupyterlab_profiles_mapper = true
   service-accounts-enabled   = true
   service-account-roles = [
+    # "manage-clients" is required for creating roles for the client
+    # "manage-users" is required for attaching roles to groups
     "view-realm", "view-users", "view-clients", "manage-clients", "manage-users"
   ]
 }
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
@@ -286,7 +286,7 @@ module "jupyterhub-openid-client" {
   jupyterlab_profiles_mapper = true
   service-accounts-enabled   = true
   service-account-roles = [
-    "view-realm", "view-users", "view-clients"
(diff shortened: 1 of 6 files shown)


src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
```diff
@@ -286,7 +286,9 @@ module "jupyterhub-openid-client" {
   jupyterlab_profiles_mapper = true
   service-accounts-enabled   = true
   service-account-roles = [
-    "view-realm", "view-users", "view-clients"
+    # "manage-clients" is required for creating roles for the client
+    # "manage-users" is required for attaching roles to groups
+    "view-realm", "view-users", "view-clients", "manage-clients", "manage-users"
```
krassowski, line 291: Why do we want to add the roles using the service account from the Authentihicator level?

Would it be possible to move the role creation to earlier in the deployment process? For example, in the place where the default roles are being currently created?

krassowski: Like in the similar way the groups are configured:

https://github.com/nebari-dev/nebari/blob/2a1b8772e3a384f5b127f92abff71d3e7ee50ff4/src/_nebari/stages/kubernetes_keycloak_configuration/template/main.tf#L71-L90

My concern is giving too much rights to the service account. It is not necessarily a blocker, I just would like to understand if you went through some thought process when developing it this way.

aktech: When I started working on it, I added the roles to jupyterhub directly (via authenticator managed roles) and then realised it's not a good idea, as there would be no way to control those at keycloak level. After that the natural progression was to change the code to add roles to keycloak instead. In this process I didn't realize doing it via terraform in the deployment process rather. 

Your concerns are very valid and is actually a blocker. I'll update this to create these at deployment as that's indeed the better way!

krassowski:
The code looks fine at first glance, but `Local Integration Tests` job is failing. I know it was flaky last time I did a PR so I restarted it.

aktech:
Yeah, the playwright test and the jupyterhub ssh tests are a bit flaky. Finally all green.

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
```diff
@@ -279,6 +279,32 @@ module "jupyterhub-openid-client" {
     "developer" = ["jupyterhub_developer", "dask_gateway_developer"]
     "analyst"   = ["jupyterhub_developer"]
   }
+  client_roles = [
+    {
+      "name" : "allow-app-sharing-role",
+      "description" : "Allow app sharing for apps created via JupyterHub App Launcher (jhub-apps)",
+      groups : [],
```
krassowski, line 286: It's odd that linter did not harmonize this

```suggestion
      "groups" : [],
```

krassowski (approved):
Looks good to me!

PR #2568, opened 2024-07-18
add typos spell checker to pre-commit

Adam-D-Lewis:
## Reference Issues or PRs
Adds [typos ](https://github.com/crate-ci/typos) to the pre-commit config to fix a lot of uncaught typos.

I think the biggest benefit of adding typos is that it caught quite a few typos that codespell doesn't catch.  I'm okay with trying to leave codespell and typos both in the pre-commit file, but they may conflict eventually.  If that's the case, perhaps we remove either codespell or typos at that time from the pre-commit file.  

Typos is faster than codespell.  Typos took 0.2 seconds to run while typos took 0.05 seconds to run in my test, but I think the time difference is not important since they are both quick to run.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?


src/_nebari/stages/infrastructure/__init__.py
```diff
@@ -606,7 +606,7 @@ class NodeSelectorKeyValue(schema.Base):
 
 class KubernetesCredentials(schema.Base):
     host: str
-    cluster_ca_certifiate: str
+    cluster_ca_certificate: str
```
Adam-D-Lewis, line 609: We should catch this in the update step and update the spelling for users if they had this set explicitly.

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/dask-gateway/controller.tf
Adam-D-Lewis, line 1: We will likely need to delete the old file during the nebari update step since otherwise it'll leave it behind in the stages folder.  We've done this in the past.

Adam-D-Lewis:
Even though this isn't done yet (see my comments), I'd be curious to get any feedback on it @dcmcand @marcelovilla @viniciusdc

marcelovilla:
@Adam-D-Lewis thank you for this! I wouldn't have thought we had that many typos in the code base and it's great you're adding a hook to catch them.

I'm not familiar with either `codespell` or `typos` but it seems `typos` worked great in this case so I think leaving it for now might be good enough.

As for fixing the typo in the configuration file, I agree we should include the logic to fix the spelling in the user's config file and delete the rendered terraform file during the upgrade step. However, I don't think it's necessarily urgent and we could have a separate PR for that if you think it's more convenient.

viniciusdc:
I completely agree on what @marcelovilla said above as well, and would vote for a separate PR to address the changes in the config file.

src/_nebari/stages/infrastructure/__init__.py
viniciusdc: I would add a `# typos: ignore` here for now, and a todo to address this for next release, so that we add as part of the update method as suggested. to keep this in mind, we could pin the issue in our notes, or in the issues tab

RELEASE.md
```diff
@@ -285,7 +285,7 @@ command and follow the instructions
 * paginator for boto3 ec2 instance types by @sblair-metrostar in https://github.com/nebari-dev/nebari/pull/1923
 * Update README.md -- fix typo. by @teoliphant in https://github.com/nebari-dev/nebari/pull/1925
 * Add more unit tests, add cleanup step for Digital Ocean integration test by @iameskild in https://github.com/nebari-dev/nebari/pull/1910
-* Add cleanup step for AWS integration test, ensure diable_prompt is passed through by @iameskild in https://github.com/nebari-dev/nebari/pull/1921
+* Add cleanup step for AWS integration test, ensure disable_prompt is passed through by @iameskild in https://github.com/nebari-dev/nebari/pull/1921
```
yarikoptic, line 288: FTR: apparently codespell doesn't work for snake_cased constructs ATM:
- https://github.com/codespell-project/codespell/issues/2730

Adam-D-Lewis:
@viniciusdc Alright, I pushed off the 2 things which would take a bit more work to resolve and opened issues for them so we can get this in sooner rather than later.  @viniciusdc or @marcelovilla Can you review again?

viniciusdc (approved):
LGTM, nice tool! 🚀

Issue #2451, opened 2024-05-07
[ENH] - Colapse Cloud provider test into single workflow using custom actions and skip conditionals


viniciusdc:
### Feature description

Some users when forking our repo, would like to have the power to skip the cloud tests when they are not required, we could use the same idea that @tylergraff suggested on #2369 

I prose we collapse all testing into a single yaml file, where we can keep it running on schedule and also add and extra conditional check as the one proposed on the above issue, so that it can be skipped. The main reason to collapse it would be that most of the current logic is mostly the same 

### Value and/or benefit

- help devex when forking the repo
- better control over the current tests

marcelovilla:
I support the idea of having one yaml file that handles cloud providers deployment—at the end the logic behind deploying is the same and the only thing that changes are configuration options and credentials. What I think we could do is abstract the logic into a reusable workflow that we then call multiple times, once for each provider.

kcpevey:
@BrianCashProf will be taking a look at this one.

PR #2569, opened 2024-07-19
Enh 2451 skip conditionals

BrianCashProf:

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

## Any other comments?

This PR is related to the issue stated in ENH: [#2451](https://github.com/nebari-dev/nebari/issues/2451)
I have added a job to the integration test related github actions. This job acts as a mediary for a boolean environment variable and a conditional statement added to integration test job at it's top level. This was made so that a user with a fork of nebari who do not have access to the credentials for the cloud resources the job needs.
This method was chosen for the following reasons:
* In testing, booleans on the step level did not prevent a failing of attempting to access the credentials.
* In testing, environment variables could not be used to do a boolean on the job at large.
* A flag that was not branch specific was requested by a few users (not stated in ticket) and this solution could provide this.
* This avoids adding "push:" statements to the YAML (as opposed to using ignore-branches) which will keep the workflow inherently manual or on a cron job as dictated thus far.


viniciusdc:
Hi, @BrianCashProf. All looks good, but I am curious about the extra steps of  `check-for-credentials_***:` I noticed you use it to calculate the output used for the extra conditional in the steps later in the pipeline. 

I don't see a problem with that, but if possible, we could instead use the env you set up in the beginning to control it directly, e.g.:

```yaml
  test-aws-integration:
    runs-on: ubuntu-latest
    if: env.NO_PROVIDER_CREDENTIALS_aws == "false"
    permissions:
      id-token: write
```

BrianCashProf:
> Hi, @BrianCashProf. All looks good, but I am curious about the extra steps of `check-for-credentials_***:` I noticed you use it to calculate the output used for the extra conditional in the steps later in the pipeline.
> 
> I don't see a problem with that, but if possible, we could instead use the env you set up in the beginning to control it directly, e.g.:
> 
> ```yaml
>   test-aws-integration:
>     runs-on: ubuntu-latest
>     if: env.NO_PROVIDER_CREDENTIALS_aws == "false"
>     permissions:
>       id-token: write
> ```

That was my initial idea, but the Environment variables were not accessible when doing a boolean at the first level of the job. 
if: env.NO_PROVIDER_CREDENTIALS_aws == "false" will produce an error that states it cannot read the env variables.
Next I tried adding the check on the token step itself, but it wouldn't stop the job from reaching out to a resource it did not have and erroring out. 
With those not being possible, I thought of what variables could be read at that top level. Given that a job output could be read in the boolean, I used the environment variable to create a job output that could then be read in the boolean statement.

viniciusdc:
Indeed, it looks like the available contexts for the `jobs.<id>.if` conditional are only `github, needs, vars, inputs`

viniciusdc (approved):
LGTM, I don't see any problems with the extra job requirement, as its default allows backward compatibility already. Also, I like the fact that this will grant forks better stability.

viniciusdc:
This will be merged after the upcoming July release.

viniciusdc:
@BrianCashProf, sorry for being pedantic here, but could you add a small comment in the new job explaining why it's in there? This way, we can easily track that in future maintenance of these files.

viniciusdc (approved):

PR #2583, opened 2024-07-26
Improve codespell support: adjust and concentrate config to pyproject.toml and fix more typos

yarikoptic:
- Concentrates codespell configuration in pyproject.toml instead of pre-commit configuration so user could simply run `codespell` and get it all clean
- Do codespell .json files (but not package-lock.json)
- Two typos fixed 
- Highly related: https://github.com/nebari-dev/nebari/pull/2568 . Apparently more typos to add to codespell ;-)

TODOs:
- [x] Check if CI workflow is actually needed -- may be pre-commit would pick up typo in that TEMP commit. pre-commit fixes it, so no CI to just detect typos is needed per se (well, pre-commit runs on changed only files AFAIK, with upgrade of codespell in pre-commit might require a run on all files... probably nothing new in procedures)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,6 +53,8 @@ repos:
             "--write",
           ]
         language: python
+        additional_dependencies:
+        - tomli
 
   # python
   - repo: https://github.com/psf/black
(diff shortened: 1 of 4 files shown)


yarikoptic:
No CI workflow detected a typo, that's odd... I will try once again with slightly adjusted/extended configuration/fixes of typos.

yarikoptic:
resolved conflict (now there is `tool.typos` config as well).  FWIW: as long as codespell is used in pre-commit (and why not typos there?) -- IMHO worth having configuration for it and thus considering this PR.

Adam-D-Lewis:
LGTM, thanks @yarikoptic

Adam-D-Lewis (approved):

PR #2584, opened 2024-07-26
Add shellcheck support (config, workflow to detect/not fix) and address some warnings

yarikoptic:
More about shellcheck https://www.shellcheck.net/ .

CI workflow has 'permissions' set only to 'read' so also should be safe.

Individual commits should have more of description.  One of the "largest" changes is explicit specification to use `bash` instead of some generic POSIX shell due to use of `echo -e` which according to shellcheck is not guaranteed to be supported by a POSIX shell... e.g. in `dash`:

```shell
❯ dash
$ echo -e "123"
-e 123
```

TODOs
- [ ] verify that CI detects warning and possibly remove custom action


yarikoptic:
FWIW, for me it is easier to just add such a PR than an issue. If would like to discuss -- I am here.

PR #2587, opened 2024-07-26
Add amazon_web_services endpoint configuration options to enable entirely private cluster on AWS EKS

### LINKED ISSUE #2586: Enable entirely private AWS EKS Clusters by adding amazon_web_services EKS cluster endpoint configuration options
### Feature description

Enable entirely private AWS EKS clusters by adding two amazon_web_services configuration options to toggle both eks_endpoint_private_access and eks_endpoint_public_access to either true or false to control whether cluster endpoint access is "Public", "Public and Private" or "Private" 

### Value and/or benefit

Nebari would be deployable into private AWS subnets with option to override the current default "Public" cluster API server endpoint to set an entirely "Private" endpoint or a "Public and Private" endpoint.

[PR#1841](https://github.com/nebari-dev/nebari/pull/1841) previously added a variable for `endpoint_private_access = var.eks_endpoint_private_access`, which is not configurable from nebari-config.yaml.  Furthermore, setting endpoint_private_access to true does not result in a truly private cluster endpoint.  The result of endpoint_private_access=true is an EKS cluster endpoint of "Public and Private".
image

To obtain an entirely private cluster endpoint, an additional variable (endpoint_public_access  = var.eks_endpoint_public_access) could be added, and configurable from nebari-config.yaml, with the option to set public access to false. 

For example:
image

### Anything else?

Updates to documentation, as proposed in [#338](https://github.com/nebari-dev/nebari-docs/issues/338), should also be provided

joneszc:
## Reference Issues or PRs
Fixes #2586

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?


dcmcand (changes requested):
@joneszc, what happens if both vars are set to false? I think we would need some validation to ensure that doesn't happen as that would be a  invalid config afaik.

tylergraff:
@dcmcand I think you make a good point, and Zach did indeed verify that AWS throws an error on the "both-false" combination.
The EKS TF module exposes the two booleans, but the AWS console provides a selector interface with 3 options for eks endpoint access: public, private, and public_and_private 

@joneszc @dcmcand How about we re-implement this feature as a single `eks_endpoint_access` variable that mimics the AWS console behavior by allowing only 1 of: `[public, private, public_and_private]` - and Nebari code translates this selection into the appropriate combination of booleans in the TF. This will align with the AWS console.

dcmcand:
I like that suggestion @tylergraff as I think it is the most clear and readable way. If we don't do that we would need to add a validator to the config to ensure they weren't both false imo. We shouldn't be allowing people to have a config that will fail to apply.

joneszc:
@tylergraff @dcmcand @viniciusdc 
Thank you for your recommendations.  I've consolidated the previous variables into option: `amazon_web_services.eks_endpoint_access`
Also added the validation steps into [\_\_init\_\_.py](https://github.com/MetroStar/nebari/blob/eks-private-endpoint/src/_nebari/stages/infrastructure/__init__.py) and logic into the [terraform](https://github.com/MetroStar/nebari/blob/eks-private-endpoint/src/_nebari/stages/infrastructure/template/aws/main.tf)

The cleanest way to set this update was to create a new PR [#2618](https://github.com/nebari-dev/nebari/pull/2618)

Issue #2507, opened 2024-06-11
[BUG] - Nodes don't scale down on GKE and AKS


Adam-D-Lewis:
### Describe the bug

I noticed that GKE won't autoscale all nodes down to 0 in some cases.  I saw that **metrics-server deployment** and the **event-exporter-gke replicaset** nodeSelector only has 
```yaml
nodeSelector:                                                                                                                                                                          
    kubernetes.io/os: linux                                                                                                                                                              
```
meaning it can be scheduled on any of the nodes preventing them from scaling down.  

Options to fix this might be 
1. Disable metrics collection - https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#enable_components
2. Set a taint on user and worker nodes (and any custom nodes groups created) to force metrics-server pod to run on general node group

I think AWS doesn't have metrics-server enabled by default so I think it's reasonable to disable it.

### Expected behavior

nodes should autoscale down

### OS and architecture in which you are running Nebari

Linux x86-64

### How to Reproduce the problem?

see above

### Compute environment

GCP

Adam-D-Lewis:
While I don't think this is the issue, it occurs to me that the other nodes might be scaling up b/c we have more pods than cpu/memory on the general node.

viniciusdc:
I think as an overall change, your 2 points seems more reasonable (to all providers). For the AWS specifically, I think the metrics is a service that you need to enable if you want to use, and costs an extra expense to keep. I also agree to disable it in such case, or make it optional....

Adam-D-Lewis:
I think the GKE event exporter replicaset is also preventing scale down

Adam-D-Lewis:
I also saw the metrics server and jupyterhub's user scheduler cause the same problem on AKS.

Adam-D-Lewis:
The solution I propose is to add a taints section to each node group class.  Thus you could specify the a taint on the user node via something like the following:
```yaml
  node_groups:
    user:
      instance: Standard_D4_v3
      taints:
        - dedicated=user:NoSchedule
```

Then, we go and make sure the corresponding toleration is added to the jupyterhub user pod so that those pods will be able to run on the user node group.  This should also work with pods started via argo-jupyter-scheduler.

PR #2605, opened 2024-08-01
Add taint to user and worker nodes

Adam-D-Lewis:
## Reference Issues or PRs
Fixes https://github.com/nebari-dev/nebari/issues/2507

- [x] I need to test running pods with Argo Workflow through Nebari Workflow Controller before merging this PR

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [x] Did you add new tests?

### How to test this PR
A few possible ways to test:
- You could not specify taints, do a deployment and then add taints, redeploy and make sure things are still working.
- You could do a deployment and specify taints and make sure things work.
- You could do a local deployment and make sure it deploys and make sure things work.

Things working is defined as:
- jupyterlab server spins up for user
- dask scheduler and worker spins up for user (when using nebari-git-dask built in env).  Use some code like [this](https://github.com/nebari-dev/nebari-demo/blob/main/04-dask.ipynb)


src/_nebari/stages/infrastructure/__init__.py
```diff
@@ -41,10 +41,33 @@ class ExistingInputVars(schema.Base):
     kube_context: str
 
 
-class DigitalOceanNodeGroup(schema.Base):
```
Adam-D-Lewis, line 44: Duplicate class

Adam-D-Lewis:
This method seems to work when tested on GCP.  One issue is that certain daemonsets won't run on the user nodes.  I saw the issue with rook ceph csi-cephfslplugin from my rook PR, but I expect it would also be an issue for the monitoring daemonset pods.

src/_nebari/stages/kubernetes_services/template/rook-ceph.tf
```diff
@@ -45,6 +45,13 @@ resource "helm_release" "rook-ceph" {
       },
       csi = {
         enableRbdDriver = false, # necessary to provision block storage, but saves some cpu and memory if not needed
+        provisionerReplicas : 1, # default is 2 on different nodes
+        pluginTolerations = [
```
Adam-D-Lewis, line 49: runs csi-driver on all nodes, even those with NoSchedule taints.  Doesn't run on nodes with NoExecute taints.  This is what the nebari-prometheus-node-exporter daemonset does so I copied it here.

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/monitoring/loki/main.tf
```diff
@@ -96,6 +96,22 @@ resource "helm_release" "grafana-promtail" {
...
+          operator = "Exists"
+          effect   = "NoSchedule"
+        },
+        {
+          key      = "node-role.kubernetes.io/control-plane"
+          operator = "Exists"
+          effect   = "NoSchedule"
+        },
+        {
+          operator = "Exists"
```
Adam-D-Lewis, line 111: runs promtail on all nodes, even those with NoSchedule taints.  Doesn't run on nodes with NoExecute taints.  This is what the nebari-prometheus-node-exporter daemonset does so I copied it here.  Promtail is what exports logs from the node so we still want it to run on the user and worker nodes.

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/monitoring/loki/main.tf
```diff
@@ -96,6 +96,22 @@ resource "helm_release" "grafana-promtail" {
...
+        {
+          key      = "node-role.kubernetes.io/master"
+          operator = "Exists"
+          effect   = "NoSchedule"
+        },
+        {
+          key      = "node-role.kubernetes.io/control-plane"
+          operator = "Exists"
+          effect   = "NoSchedule"
+        },
```
Adam-D-Lewis, line 109: These top 2 are the default value for this helm chart.

Adam-D-Lewis:
Okay, so things are working for the user node group.  I tried adding a taint to the worker node group, but the dask scheduler won't run on the tainted worker node group.  See [this commit](https://github.com/nebari-dev/nebari/pull/2605/commits/c55cd5f7ecc8911ba93d6acea6cec29114ce5bd4) to see what I tried in a quick test.  I do see the new scheduler_pod_extra_config value in /var/lib/dask-gateway/config.json, but the scheduler tolerations look like 
```
│   tolerations:                                                                                                                                                                            │
│   - effect: NoExecute                                                                                                                                                                     │
│     key: node.kubernetes.io/not-ready                                                                                                                                                     │
│     operator: Exists                                                                                                                                                                      │
│     tolerationSeconds: 300                                                                                                                                                                │
│   - effect: NoExecute                                                                                                                                                                     │
│     key: node.kubernetes.io/unreachable                                                                                                                                                   │
│     operator: Exists                                                                                                                                                                      │
│     tolerationSeconds: 300      
```
so I think possibly the merge isn't going as expected, but I need to verify.  [The docs](https://gateway.dask.org/api-server.html?highlight=scheduler_extra_pod_config) say that "This dict will be deep merged with the scheduler pod spec (a V1PodSpec object) before submission. Keys should match those in the [kubernetes spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#podspec-v1-core), and should be camelCase."

Adam-D-Lewis:
I managed to get the taints applied to the scheduler pod in [this commit](https://github.com/nebari-dev/nebari/pull/2605/commits/a1370c939823899763b9d9b21fe9a492436d36b9).  I would have expected the `c.KubeClusterConfig.scheduler_extra_pod_config` to get merged with the options returned by the function passed to c.Backend.cluster_options, but it wasn't.  

[ ] I should verify this and maybe submit an issue to dask-gateway.

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/dask-gateway/files/gateway_config.py
```diff
@@ -227,18 +229,23 @@ def base_username_mount(username, uid=1000, gid=100):
     }
 
 
-def worker_profile(options, user):
```
Adam-D-Lewis, line 230: I renamed this function since it affects the scheduler as well and not just the worker

Adam-D-Lewis:
Okay things were working as expected for the jupyterlab pod and the dask worker and scheduler pods on GKE.  I need to test on:
- [ ] AWS
- [ ] Azure.

I also need to test:
- [ ] running an Argo Workflows pod.

Adam-D-Lewis:
- [ ] We should probably disallow taints on the general node since nothing will work correctly and it'd be a lot of work to fix that and no has asked for it yet.

Adam-D-Lewis:
- [ ] make sure this won't break if using provider or local providers

Adam-D-Lewis:
- [ ] During upgrade, we'll need to add taints to each of the node groups listed or else the users will get the following error during deployment.

```
[terraform]: │ The given value is not suitable for var.node_groups declared at
[terraform]: │ variables.tf:31,1-23: element 0: attribute "node_taints" is required.
```

src/_nebari/stages/infrastructure/__init__.py
```diff
@@ -349,19 +386,26 @@ class GCPGuestAccelerator(schema.Base):
...
+        min_nodes=0,
+        max_nodes=5,
+        taints=[schema.Taint(key="dedicated", value="user", effect="NoSchedule")],
+    ),
+    "worker": GCPNodeGroup(
+        instance="e2-standard-4",
+        min_nodes=0,
+        max_nodes=5,
+        taints=[schema.Taint(key="dedicated", value="worker", effect="NoSchedule")],
+    ),
```
Adam-D-Lewis, line 408: do this for other DEFAULT_NODE_GROUPS

Adam-D-Lewis:
Okay, I think things are working on AWS.  I resolved some issues on Azure, but still need to test Azure.

Adam-D-Lewis:
Tested on GCP, Azure, and AWS and works well on all.  I did the following in my test:
- Created an admin user
- launched a dask cluster
- verified taints set on user and worker nodes
- verified toleration set on jupyter user pod and dask scheduler/worker pods

Adam-D-Lewis:
- [ ] I want to create an issue at least to prompt users on upgrade to ask if they want to add the taints for potential cost reductions.

src/_nebari/stages/infrastructure/__init__.py
```diff
@@ -150,6 +201,22 @@ class AWSNodeGroupInputVars(schema.Base):
     permissions_boundary: Optional[str] = None
     ami_type: Optional[AWSAmiTypes] = None
     launch_template: Optional[AWSNodeLaunchTemplate] = None
+    node_taints: list[dict]
+
+    @field_validator("node_taints", mode="before")
```
Adam-D-Lewis, line 206: This code is repeated for GCP and AWS NodeGroupInputVars classes, but that's b/c the format expected by GCP and AWS terraform modules for taints happens to be the same.  I think they could evolve separately and so I chose to duplicate the code in this case.

Adam-D-Lewis:
We should add some instructions to the docs about adding gpu nodes.  Users should add the user taint to other gpu user node profiles in order to prevent the same issue this PR prevents.

Adam-D-Lewis:
Rather than making the user make sure to put a consistent taint on each node group, maybe we should just have a "type" field on node groups to simplify this.  Less flexible, but I think it's flexible enough for the use cases we expect.
```yaml
google_cloud_platform:
  project: qhub-279316
  region: us-central1
  kubernetes_version: 1.28.9-gke.1289000
  tags:
  - "nebari-quansight-dev"
  node_groups:
    general:
      instance: n1-standard-8
      min_nodes: 1
      max_nodes: 1
	  type: general   <--------------------------NEW-------------------------------

    user:
      instance: n1-standard-4
      min_nodes: 0
      max_nodes: 200
      type: user   <----------------------------- NEW -----------------------------

    large:
      instance: n1-standard-8
      min_nodes: 0
      max_nodes: 200
      type: user   <----------------------------- NEW -----------------------------

    worker:
      instance: n1-standard-4
      min_nodes: 0
      max_nodes: 1000
      type: worker   <----------------------------- NEW -----------------------------
```

viniciusdc:
Hi, @Adam-D-Lewis. Were the changes we discussed last week applied? Regardingng the default taints and overrides?

Adam-D-Lewis:
> Hi, @Adam-D-Lewis. Were the changes we discussed last week applied? Regardingng the default taints and overrides?

Yes, they were in [this commit](https://github.com/nebari-dev/nebari/pull/2605/commits/747a2933b5d32b02d18d16c2a1b866e1ee6280aa).  See the set_missing_taints_to_default_taints method.

viniciusdc:
It looks like `'provider_enum_name_map'` is not recognized, though it seems it was not removed...

Adam-D-Lewis:
Failing test appears unrelated to this PR since it's a playwright test and this PR makes no changes to the UI/UX other than the command line when running `nebari init`.

viniciusdc:
Thanks @Adam-D-Lewis , thats related to the recent conda-store update.

marcelovilla:
@Adam-D-Lewis the failing tests have been addressed in https://github.com/nebari-dev/nebari/pull/2911

viniciusdc (approved):
Working great, I tested both on GCP and AWS wheres the later was where I most worried about due to EBS volume mounting shenanigans. I asserted all Taints and tolerations are being applied to all auxiliary nodes -- general is not touched by default.

Keep in mind that this addressed the original issues with the non-scaling down pods on GCP with due to how taints work, e.g, this makes sure that under our nodes only nebari resources are deployed. Though, because general does not use toleration by default, unassigned pods might end up being scaled under the general node, we just need to keep that in mind and update the docs accordingly.

viniciusdc:
To test this PR, I setup a cluster running on our previous release, thenupgradede to this branch version of nebari and redeployed, wanted to mimic as much as possible an update to existing infra

PR #2611, opened 2024-08-06
Move codespell config to pyproject.toml only

Adam-D-Lewis:
## Reference Issues or PRs
The codespell config was added to pyproject.toml in https://github.com/nebari-dev/nebari/pull/2583.  This allows users to run `codespell` from command line and pick up the config as well as be picked up by pre-commit.  Typos config is already in pyproject as well since some options couldn't be set via command line.
I think we should move the codespell config there exclusively so we don't have to update things in 2 places.  The only difference is we still want --write on pre-commit, but not when using codespell from command line.

- [ ] Documentation Update
- [ ] Build related changes
- [x] Other (please describe): cicid

## Testing

- [ ] Did you add new tests?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -47,9 +47,6 @@ repos:
       - id: codespell
         args:
           [
-        #     "--builtin=rare,clear,informal,names",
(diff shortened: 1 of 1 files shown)


marcelovilla (approved):
LGTM!

PR #2632, opened 2024-08-15
Action var tests off

BrianCashProf:

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

## Any other comments?

Following the changes introduced in https://github.com/nebari-dev/nebari/pull/2569, We have decided to use action variables instead of local environment variables. 
These action variables must be set at the repo level and can be set by forks differently to the main repository without merge conflicts. 
The action variable is set as a binary boolean (with 0 being false and 1 being true).


viniciusdc:
Hi @BrianCashProf, thanks for opening the PR; I wonder if the variable could be boolean instead of the int. Also, while reading the [docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#using-the-vars-context-to-access-configuration-variable-values) for variables again, I noticed this:
> If a configuration variable has not been set, the return value of a context referencing the variable will be an empty string.

So, one thing that we could also change is the primary condition, so instead of looking for a value of the given variable (and this requiring that to be preset), we could use:
```
if: ${{ ! vars.NO_PROVIDER_CREDENTIALS_*** }}
```

BrianCashProf:
> Hi @BrianCashProf, thanks for opening the PR; I wonder if the variable could be boolean instead of the int. Also, while reading the [docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#using-the-vars-context-to-access-configuration-variable-values) for variables again, I noticed this:
> 
> > If a configuration variable has not been set, the return value of a context referencing the variable will be an empty string.
> 
> So, one thing that we could also change is the primary condition, so instead of looking for a value of the given variable (and this requiring that to be preset), we could use:
> 
> ```
> if: ${{ ! vars.NO_PROVIDER_CREDENTIALS_*** }}
> ```
> 
> I've attested that this works as expected, but with the caveat that those variables must be set in the repository settings of a given fork.

I'll take a look into this method.

tylergraff:
> If by default we want to run the pipeline and the flag only needs to be placed to skip the pipeline; then this is my suggestion

@BrianCashProf Yes we do want to run by default and only skip if the flag is set accordingly. Please make your suggested change. Once that is committed, is there anything else needed before this PR can be merged?

viniciusdc (commented):
@BrianCashProf, make sure they follow the standard you suggested above. `NO_PROVIDER_CREDENTIALS_AWS != 'true'` to `vars.NO_PROVIDER_CREDENTIALS_AZURE == 0`. Also, now that we removed the previous `check-for-credentials` job and we are also relying on different variables for each provider anyway, I suggest that we rename the variable to `SKIP_***_INTEGRATION_TEST`.

.github/workflows/test_aws_integration.yaml
```diff
@@ -33,31 +33,12 @@ env:
...
-      - name: Check if user wants to run AWS integration based on credentials
-        id: flag-check
-        run: |
-          if [ "${{ env.NO_PROVIDER_CREDENTIALS_aws }}" == "true" ]; then
-            echo "::set-output name=provider_credentials_aws::0"
-          else
-            echo "::set-output name=provider_credentials_aws::1"
-          fi
-
+# Must have action variable NO_PROVIDER_CREDENTIALS_AWS set up in the repo as a binary boolean.
```
viniciusdc, line 38: Is this still required?

BrianCashProf: Nope, I've removed the comment.

viniciusdc (approved):
LGTM, thanks @BrianCashProf

PR #2706, opened 2024-09-04
revert breaking change to azure deployment test

Adam-D-Lewis:
## Reference Issues or PRs
https://github.com/nebari-dev/nebari/pull/2632 removed a few credentials on the Azure deployment test which now fails with
```
Error building ARM Config: Authenticating using the Azure CLI is only supported as a User (not a Service Principal).
```
This PR reverts the changes to the credentials.  

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

## How to test this PR?
If [this](https://github.com/nebari-dev/nebari/actions/runs/10706057961/job/29682871900) finishes successfully then it verifies this is the fix for the issue.

### DIFF
--- a/.github/workflows/test_azure_integration.yaml
+++ b/.github/workflows/test_azure_integration.yaml
@@ -70,6 +70,13 @@ jobs:
             kv/data/repository/nebari-dev/nebari/azure/nebari-dev-ci/github-nebari-dev-repo-ci subscription_id | ARM_SUBSCRIPTION_ID;
             kv/data/repository/nebari-dev/nebari/cloudflare/internal-devops@quansight.com/nebari-dev-ci token | CLOUDFLARE_TOKEN;
 
+      - name: 'Azure login'
+        uses: azure/login@v2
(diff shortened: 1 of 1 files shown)


Adam-D-Lewis:
Okay, the re-run failed, but the error looked like it was due to it being run from a branch and not develop.  I'd like to merge and see if it runs successfully.

Adam-D-Lewis:
I think we should merge and test, if it still fails, then revert.

viniciusdc:
@Adam-D-Lewis can we trigger the Azure workflow from this branch as a test?

viniciusdc (approved):
LGTM, This is a quick fix to test the OIDC changes, I am also not against keeping these in here as the global skip var should avoid these requirments showing up on forked repos as well.

Adam-D-Lewis:
Looks like the reason it failed before it tests is that the OIDC login only works from the develop branch currently.

Issue #2681, opened 2024-09-03
Fix code scanning alert - EKS should have the encryption of secrets enabled


dcmcand:
Tracking issue for:
- [x] https://github.com/nebari-dev/nebari/security/code-scanning/35

tylergraff:
@dcmcand the link in the description leads to a 404 page for me. Am I missing permissions to view it?

joneszc:
@dcmcand @tylergraf

Proposing [PR#2723](https://github.com/nebari-dev/nebari/pull/2723) here to add option for a user to specify a KMS-key ARN in nebari-config.yaml, which will set EKS to encrypt secrets:
```
amazon_web_services:
  eks_kms_arn: 'arn:aws:kms:us-east-1:010101010:key/3xxxxxxx-xxxxx-xxxxx-xxxxx' 
```


<img width="979" alt="image" src="https://github.com/user-attachments/assets/6a64df6c-9ce4-4198-a082-907e1c47a5df">

viniciusdc:
This relies on the user already required to set up the KMS key themselves; I don't see a problem with allowing users to provide theirs, but I would prefer that Nebari create its own (and manage it) by default. What do you think @dcmcand @joneszc

tylergraff:
For our use-case, we need to allow organizations to create their own keys to match their own policy, or possibly re-use existing KMS keys. We don't have a use-case for Nebari to create and manage its own key by default, so we are considering that to be out of scope for this issue.

viniciusdc:
> For our use case, we need to allow organizations to create their own keys to match their own policy or possibly reuse existing KMS keys. We don't have a use case for Nebari to create and manage its own key by default, so we are considering that to be out of scope for this issue.

It is entirely understandable, though. IMO, this would be another issue: allowing the user to configure custom KMS encryption keys, which is not the scope of this one. This issue attempts to solve the problem that the cluster currently does not have encryption set up by default, which is a security concern. They are similar but address different use-cases.

joneszc:
> > For our use case, we need to allow organizations to create their own keys to match their own policy or possibly reuse existing KMS keys. We don't have a use case for Nebari to create and manage its own key by default, so we are considering that to be out of scope for this issue.
> 
> It is entirely understandable, though. IMO, this would be another issue: allowing the user to configure custom KMS encryption keys.
> 
> This issue attempts to solve the problem that the cluster currently does not have encryption set up by default, which is a security concern. They are similar but address different use-cases.



[...]

PR #2723, opened 2024-09-17
Add config option to enable the encryption of AWS EKS secrets

joneszc:
## Reference Issues or PRs
Fixes [#2681](https://github.com/nebari-dev/nebari/issues/2681)

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

### Allows user to set EKS encryption of secrets by specifying a KMS key ARN in nebari-config.yaml
```
amazon_web_services:
  eks_kms_arn: 'arn:aws:kms:us-east-1:010101010:key/3xxxxxxx-xxxxx-xxxxx-xxxxx'
```
image

### The KMS key must meet the following conditions:
* Symmetric
* Can encrypt and decrypt data
* Created in the same AWS Region as the cluster
* If the KMS key was created in a different account, the [IAM principal](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles.html#iam-term-principal) must have access to the KMS key.


kenafoster (approved):
LGTM but if this is merged we should open up a nebari-docs since it's a little tricky to implement and understand

I've confirmed that most of the mistakes you could make (wrong region, bad format, key doesn't exist or Nebari deploy user can't see it) will fail in the stage where it creates the IAM policy, so at least it won't get to modifying the EKS cluster before failing

viniciusdc:
I would say that we could have a validator in the AWS Pydantic provider (similar to how we already do for the instance types and region) so that the feedback loop in case of errors is quicker for the user doing the deployment (instead of relying on terraform only for the "check").

joneszc:
> I would say that we could have a validator in the AWS Pydantic provider (similar to how we already do for the instance types and region) so that the feedback loop in case of errors is quicker for the user doing the deployment (instead of relying on terraform only for the "check").

@viniciusdc @tylergraff 
As suggested, I have added initial kms-key arn validation steps [here](https://github.com/MetroStar/nebari/commit/e13fdb311e0889601488ebe35d77951c8a49ba3c)
I'm still testing these updates

joneszc:
> I think the issues with the tests below are due to the recent move from `develop` to `main`. Can you update your fork repo and update this branch?

@viniciusdc,
This [branch](https://github.com/MetroStar/nebari/tree/eks-cluster-encryption) is now, at time of writing, up to date with nebari-dev/nebari:main.  Does this suffice with your recommended action or are we still missing something? Thanks!

joneszc:
> I think the issues with the tests below are due to the recent move from `develop` to `main`. Can you update your fork repo and update this branch?

@viniciusdc,
This [branch](https://github.com/MetroStar/nebari/tree/eks-cluster-encryption) should now, at time of writing, be up to date with nebari-dev/nebari:main. Thanks!

joneszc:
Restoring [branch](https://github.com/MetroStar/nebari/tree/eks-cluster-encryption)

PR #2738, opened 2024-09-24
fix group mounting paths

### LINKED ISSUE #2737: [BUG] - `shared` mount files not showing up on recent upgrade for `2024.9.1rc3`
### Describe the bug

After the changes  in #2593, a small bug was introduced concerning how the shared volume mount was being handled. Instead of mounting to `/shared/<group_name>`, an extra slash delimiter was unintentionally added due to a bug in one of the Keycloak to JupyterHub sections of the code. 

This led to new subpaths being included in the pod specifications as `/shared//<group_name>`. For newer deployments, this issue was not apparent and was difficult to catch unless inspecting the pod's YAML specification directly, since the subpaths and symlinks were generated correctly—albeit pointing to a double-slash folder location.

However, the problem arises in existing deployments because the symlinks would have been broken. The volume mounts were not correctly mounted since those subpaths did not exist, which would most likely generate conflicting data.

Bellow you can see a comparassion of a upsgraded cluster and a fresly created one:
image
> Here due to an issue with the setings in the volumeMoutns, the vlume was not created and thus the expected `shared` link was not correcly binding in the user jupyterlab container as seen below:
image

Here you can see the fresly instanciated EFSs from a new cluster do accept the double slashing, albeit this is wrong. 
image (1)

### Expected behavior

shared folder should exist and contain the required user groups on upgraded nebari clusters

Linux

upgrade an existing nebari deployment to 2024.9.1rc* release

### Command output

_No response_

_No response_

### Compute environment

None

### Integrations

_No response_

### Anything else?

_No response_

viniciusdc:
## Reference Issues or PRs
fixes #2737 

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

### DIFF
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/03-profiles.py
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/03-profiles.py
@@ -561,12 +561,12 @@ def render_profiles(spawner):
     # e.g. /projects/myproj -> myproj
     # and /developers -> developers
     groups = [Path(group).name for group in auth_state["oauth_user"]["groups"]]
+    groups_with_permission_to_mount = [
+        Path(group).name
+        for group in auth_state.get("groups_with_permission_to_mount", [])
+    ]
 
     keycloak_profilenames = auth_state["oauth_user"].get("jupyterlab_profiles", [])
-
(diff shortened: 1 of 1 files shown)


viniciusdc:
Before
Captura de tela de 2024-09-24 16 24 49
after these changes
Captura de tela de 2024-09-24 16 17 17

marcelovilla (approved):
Thanks @viniciusdc !

PR #2752, opened 2024-10-03
UPDATED: Add config option to enable the encryption of AWS EKS secrets

joneszc:
## Reference Issues or PRs
Fixes [#2681](https://github.com/nebari-dev/nebari/issues/2681)
Fixes [#2746](https://github.com/nebari-dev/nebari/issues/2746)
Modifies [PR#2723](https://github.com/nebari-dev/nebari/pull/2723)

## What does this implement/fix?

_Put a `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds a feature)
- [ ] Breaking change (fix or feature that would cause existing features not to work as expected)
- [ ] Documentation Update
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Other (please describe):

## Testing

- [x] Did you test the pull request locally?
- [ ] Did you add new tests?

### Allows user to set EKS encryption of secrets by specifying a KMS key ARN in nebari-config.yaml
```
amazon_web_services:
  eks_kms_arn: 'arn:aws:kms:us-east-1:010101010:key/3xxxxxxx-xxxxx-xxxxx-xxxxx'
```
image

### The KMS key must meet the following conditions:
* Symmetric
* Can encrypt and decrypt data
* Created in the same AWS Region as the cluster
* If the KMS key was created in a different account, the [IAM principal](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles.html#iam-term-principal) must have access to the KMS key.

Issue #2746, opened 2024-09-26
[ENH] Allows users to pass custom KMS keys for AWS eks cluster encryption


tylergraff:
### Feature description

For our use-case, we need to allow organizations to create their own AWS KMS keys to match their organizational policy, or possibly re-use existing KMS keys.

### Value and/or benefit

Allows organizations to address security issues related to KMS encryption.

### Anything else?

Related to: 
https://github.com/nebari-dev/nebari/issues/2681

tylergraff:
relates to: https://github.com/nebari-dev/governance/issues/55

tylergraff:
relates to: https://github.com/nebari-dev/nebari/security/code-scanning/35

tylergraff:
Proposed fix:
https://github.com/nebari-dev/nebari/pull/2723

PR #2788, opened 2024-10-22
Add config option to enable the encryption of AWS EKS secrets

joneszc:
## Reference Issues or PRs
Fixes [#2681](https://github.com/nebari-dev/nebari/issues/2681)
Fixes [#2746](https://github.com/nebari-dev/nebari/issues/2746)
Modifies [PR#2723](https://github.com/nebari-dev/nebari/pull/2723) (Failing Tests / Pytest)
Modifies [PR#2752](https://github.com/nebari-dev/nebari/pull/2752) (Failing Tests / Pytest)

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

### Allows user to set EKS encryption of secrets by specifying a KMS key ARN in nebari-config.yaml
```
amazon_web_services:
  eks_kms_arn: 'arn:aws:kms:us-east-1:010101010:key/3xxxxxxx-xxxxx-xxxxx-xxxxx'
```
image

### The KMS key must meet the following conditions:
* Symmetric
* Can encrypt and decrypt data
* Created in the same AWS Region as the cluster
* If the KMS key was created in a different account, the [IAM principal](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles.html#iam-term-principal) must have access to the KMS key.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
         exclude: "^src/_nebari/template/"
 
   - repo: https://github.com/crate-ci/typos
-    rev: v1.26.0
+    rev: v1.27.0
     hooks:
       - id: typos
 
@@ -61,7 +61,7 @@ repos:
         args: ["--line-length=88", "--exclude=/src/_nebari/template/"]
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.6.9
+    rev: v0.7.2
     hooks:
       - id: ruff
         args: ["--fix"]
@@ -77,7 +77,7 @@ repos:
 
   # terraform
   - repo: https://github.com/antonbabenko/pre-commit-terraform
-    rev: v1.96.1
+    rev: v1.96.2
     hooks:
       - id: terraform_fmt
         args:


viniciusdc:
@joneszc, there are two PRs which seem to add the same thing, this one and #2752 -- I assume the first one was the original; can you close this one? (or move any relevant changes back to the other PR?)

dcmcand:
@joneszc can we close #2752 and #2723 since we have this one?

joneszc:
> @joneszc can we close #2752 and #2723 since we have this one?

@dcmcand,
Yes, those two PRs were built on forks of the old `develop` branch that is now `main`
Thanks for help determining that the branch was not the issue causing Pytest failures.  #2752 and #2723 can be closed.

joneszc:
@viniciusdc 
I've opened [PR#537](https://github.com/nebari-dev/nebari-docs/pull/537) to update the docs per your [request](https://github.com/nebari-dev/nebari/pull/2752#issuecomment-2435310592)

src/_nebari/stages/infrastructure/__init__.py
```diff
@@ -554,6 +556,37 @@ def _check_input(cls, data: Any) -> Any:
...
+                        )
+                    else:
+                        raise ValueError(
+                            f"Amazon Web Services KMS Key with ID {key_id} is not of type Symmetric"
+                        )
+            else:
+                raise ValueError(
+                    f"Amazon Web Services KMS Key with ARN {data['eks_kms_arn']} not one of available/enabled keys={[v['Arn'] for v in available_kms_keys.values()]}"
+                )
+
```
dcmcand, line 589: You could avoid a lot of nesting here by flipping the logic on these if statements. Something like:
```python
        available_kms_keys = amazon_web_services.kms_key_arns(data["region"])
        # don't check if eks_kms_arn is not set
        if "eks_kms_arn" not in data or data["eks_kms_arn"] is None:
            return data
        key_id = [
            id for id in available_kms_keys.keys() if id in data["eks_kms_arn"]
        ]
        # Raise error if key_id is not found in available_kms_keys
        if (
            len(key_id) != 1
            or available_kms_keys[key_id[0]]["Arn"] != data["eks_kms_arn"]
        ):
            raise ValueError(
                f"Amazon Web Services KMS Key with ARN {data['eks_kms_arn']} not one of available/enabled keys={[v['Arn'] for v in available_kms_keys.values()]}"
            )

        key_id = key_id[0]
        # EKS cluster encryption requires a Symmetric key that is set to encrypt and decrypt data
        if available_kms_keys[key_id]["KeySpec"] != "SYMMETRIC_DEFAULT":
            raise ValueError(
                f"Amazon Web Services KMS Key with ID {key_id} is not of type Symmetric"
            )
        if available_kms_keys[key_id]["KeyUsage"] != "ENCRYPT_DECRYPT":
            raise ValueError(
                f"Amazon Web Services KMS Key with ID {key_id} KeyUsage not set to 'Encrypt and decrypt' data"
            )        available_kms_keys = amazon_web_services.kms_key_arns(data["region"])
        # don't check if eks_kms_arn is not set
        if "eks_kms_arn" not in data or data["eks_kms_arn"] is None:
            return data
        key_id = [
            id for id in available_kms_keys.keys() if id in data["eks_kms_arn"]
        ]
        # Raise error if key_id is not found in available_kms_keys
        if (
            len(key_id) != 1
            or available_kms_keys[key_id[0]]["Arn"] != data["eks_kms_arn"]
        ):
            raise ValueError(
                f"Amazon Web Services KMS Key with ARN {data['eks_kms_arn']} not one of available/enabled keys={[v['Arn'] for v in available_kms_keys.values()]}"
            )

        key_id = key_id[0]
        # EKS cluster encryption requires a Symmetric key that is set to encrypt and decrypt data
        if available_kms_keys[key_id]["KeySpec"] != "SYMMETRIC_DEFAULT":
            raise ValueError(
                f"Amazon Web Services KMS Key with ID {key_id} is not of type Symmetric"
            )
        if available_kms_keys[key_id]["KeyUsage"] != "ENCRYPT_DECRYPT":
            raise ValueError(
                f"Amazon Web Services KMS Key with ID {key_id} KeyUsage not set to 'Encrypt and decrypt' data"
            )
```
Has less deeply nested logic and is therefore easier to read and test.

viniciusdc:
> However, attempting thereafter to re-deploy to remove the previously set EKS secrets encryption will fail as terraform attempts to delete and rebuild the EKS cluster but cannot due to existing node groups.
Hi @joneszc, thanks for checking that out! I was already expecting it to fail, but I had another thing in mind: they might be connected. Can you post a sanitized output of the terraform error and any error messages you might encounter in the CloudTrail history? I suspect you will find something related to the KMS key in there. 

I ask for it to validate if it will be beneficial to have this as an immutable field or, depending on the error, we can add manual steps to the user in our docs to disable it.

src/_nebari/provider/cloud/amazon_web_services.py
```diff
@@ -121,6 +121,27 @@ def instances(region: str) -> Dict[str, str]:
...
+        "KeyUsage",
+        "KeySpec",
+        # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/kms/client/describe_key.html#:~:text=Response%20Structure
+    ]
+    kms_keys = [
+        client.describe_key(KeyId=j["KeyId"]).get("KeyMetadata")
+        for i in paginator.paginate()
+        for j in i["Keys"]
+    ]
+    return {i["KeyId"]: {k: i[k] for k in fields} for i in kms_keys if i["Enabled"]}
```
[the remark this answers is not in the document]
viniciusdc: I do prefer this as well!

src/_nebari/provider/cloud/amazon_web_services.py
```diff
@@ -121,6 +121,27 @@ def instances(region: str) -> Dict[str, str]:
     return {t: t for t in instance_types}
 
 
+@functools.lru_cache()
+def kms_key_arns(region: str) -> Dict[str, dict]:
+    """Return dict of available/enabled KMS key IDs and associated KeyMetadata for the AWS region."""
+    session = aws_session(region=region)
+    client = session.client("kms")
+    # https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/kms/client/list_keys.html
+    paginator = client.get_paginator("list_keys")
```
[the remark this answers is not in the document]
joneszc: updated

src/_nebari/provider/cloud/amazon_web_services.py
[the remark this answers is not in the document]
joneszc: I've made the function more readable but I did not separate into two functions.

src/_nebari/stages/infrastructure/__init__.py
joneszc: my hesitation to invoke `return data` in numerous places in this section is that could cause issues in the future if additional validators are appended after the KMS key validator lines under `_check_input` for aws services.

dcmcand: @joneszc I can see that. It might make sense to pull every validator into its own method then, and have `_check_input` just call a list of validators. That way each could exit early if they can and nothing would get skipped. I think that would also be cleaner for readability and avoid too much responsibility for a single method.

src/_nebari/provider/cloud/amazon_web_services.py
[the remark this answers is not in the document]
dcmcand: Yeah, this looks much cleaner and more readable. Nice work there.

src/_nebari/stages/infrastructure/__init__.py
joneszc: That would make sense.  I could open a separate PR for breaking out multiple methods for _check_input, if you'd like.  For now, I've updated the logic, based on your recommendation, for the first check.  I've also added a check to ensure the KMS key is a customer-managed versus aws-managed key.  I've left the final check in place with conditionals to precisely triage the error; for example, there is a case for a `SYMMETRIC_DEFAULT` key that could seem adequate but will fail if it is "GENERATE_VERIFY_MAC" rather than "ENCRYPT_DECRYPT".

dcmcand: Can you open an issue for breaking out the _check_input methods so that doesn't get lost?

dcmcand:
So @joneszc am I reading that correctly that enabling this option will destroy and replace your cluster? We should probably go ahead and make this field immutable then. We definitely don't want anyone accidentally destroying their deploy. The docs should reflect that this should only be used on fresh deploys too.

src/_nebari/provider/cloud/amazon_web_services.py
```diff
@@ -121,6 +121,27 @@ def instances(region: str) -> Dict[str, str]:
     return {t: t for t in instance_types}
 
 
+@functools.lru_cache()
```
dcmcand, line 124: This is stylistic, so it is up to you whether to adopt this. I would probably add a `Kms_Key_Info` class or something.

```python
@dataclass
class Kms_Key_Info:
    Arn: str
    KeyUsage: str
    KeySpec: str
    KeyManager: str
```

The advantage here is more clarity in your type annotations. Rather than returning `Dict[str, dict] ` you can return `Dict[str, Kms_Key_Info]` which is clearer what is actually returned.

src/_nebari/stages/infrastructure/__init__.py
```diff
@@ -174,6 +174,7 @@ class AWSInputVars(schema.Base):
     eks_endpoint_access: Optional[
         Literal["private", "public", "public_and_private"]
     ] = "public"
+    eks_kms_arn: Optional[str] = None
```
dcmcand, line 177: I think this field should be immutable to prevent clusters from being deleted.

See [src/_nebari/stages/kubernetes_services/__init__.py:66 ](https://github.com/nebari-dev/nebari/blob/88dfe2447c56689c3a9828e6abef4e1c02903faf/src/_nebari/stages/kubernetes_services/__init__.py#L66) for an example of an immutable field.

dcmcand (changes requested):
Hi @joneszc , This is looking really good. I think that the config option needs to be immutable though.

src/_nebari/stages/infrastructure/__init__.py
joneszc: > So @joneszc am I reading that correctly that enabling this option will destroy and replace your cluster? We should probably go ahead and make this field immutable then. We definitely don't want anyone accidentally destroying their deploy. The docs should reflect that this should only be used on fresh deploys too.

Hello @dcmcand 
Enabling the `eks_kms_arn` option, including doing so as a re-deploy on an existing cluster, is not a problem.  The issue arises when one attempts to disable the cluster encryption on an existing cluster, which causes a failed attempt to destroy the cluster.  I've added a warning for this behavior in the [docs update](https://github.com/nebari-dev/nebari-docs/pull/537/commits/2bbe698ea4826607c9d450401d4c81e5c73c3fa6).  Making the field immutable will block users from being able to set cluster encryption on an existing cluster, which is functional. Is there a mechanism for allowing the field to be set, if not set already, but immutable if altered from a set state?  The risk of destroying the cluster is ineluctably countered by the error:
`Error: deleting EKS Cluster (nebari-test-dev): operation error EKS: DeleteCluster, https response error StatusCode: 409, RequestID: de8f18ba-0abe-42ae-961f-86d8865fbcf3, ResourceInUseException: Cluster has nodegroups attached`

src/_nebari/stages/infrastructure/__init__.py
joneszc: yes!

src/_nebari/stages/infrastructure/__init__.py
dcmcand: @joneszc can you test the above scenario and update with the latest main?

joneszc: @dcmcand 
Yes, that is the correct logic flow.  I'm testing now.  I need to test how the components behave as well after enabling encryption and then actually encrypting the secrets, as the docs mention [this important step](https://docs.aws.amazon.com/eks/latest/userguide/enable-kms.html#:~:text=After%20you%20enabled), which might have additional repercussions on Nebari components besides the Cluster functionality.

src/_nebari/provider/cloud/amazon_web_services.py
joneszc: @dcmcand 
Updated, per your suggestion

src/_nebari/stages/infrastructure/__init__.py
joneszc: @dcmcand 
Following a failed attempt to disable encryption, by setting a previously configured `amazon_web_services.eks_kms_arn` value to `null` or removing the field altogether, a follow-up re-deploy to reenable the original encryption setting does succeed without any issues.  Also, an attempt to switch KMS keys by replacing a previously set KMS key arn with a new arn will succeed in a re-deploy, but the KMS key is not actually changed on the EKS Cluster.   Essentially, no change is made and the cluster functions as before.  I'll make a note of this in the docs PR.

viniciusdc: That's good to hear, @joneszc. If I understood correctly, if by any chance the user removes the encryption key from the config and redeploys, the user will first face an error, but what I assume a few minutes, the user may be able to retain the redeployment process and succeeding in reaching the same state before the key by re-attempting the deployment? 

If so, please add this as an [admonition](https://docusaurus.io/docs/markdown-features/admonitions) to the docs under a `warning`.

src/_nebari/stages/infrastructure/__init__.py
joneszc: @viniciusdc
yes, for sure

dcmcand (approved):

PR #2797, opened 2024-10-25
Fix immutable field validation error when a sub-schema is not Pydantic

### LINKED ISSUE #2767: [BUG] - 2024.9.1 Immutable Field Check Appears to Fail for Certain Schema Types
### Describe the bug

I'm running into an odd error with the nebari-self-registration extension and changing its config values when core nebari is 2024.9.1
Starting with a config block like the one below, I am unable to change the value of the self_registration.values.images.tag field
 self_registration:
  namespace: self-registration
  ```
  values:
    image:
      repository: quay.io/nebari/nebari-self-registration
      tag: "20240926-1841"
      #tag: "20241011-1937"
```
The end of the trace for the error is:
```
│.../_nebari/stages/terr │
│ aform_state/__init__.py:279 in check_immutable_fields                                            │
│                                                                                                  │
│   276 │   │   │   │   │   │   else:                                                              │
│   277 │   │   │   │   │   │   │   raise e                                                        │
│   278 │   │   │   extra_field_schema = schema.ExtraFieldSchema(                                  │
│ ❱ 279 │   │   │   │   **bottom_level_schema.model_fields[keys[-1]].json_schema_extra or {}       │
│   280 │   │   │   )                                                                              │
│   281 │   │   │   if extra_field_schema.immutable:                                               │
│   282 │   │   │   │   key_path = ".".join(keys)                                                  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
AttributeError: 'CommentedMap' object has no attribute 'model_fields'
```

It seems to me that the check_immutable_fields()  runs into a problem specifically because the values key[ schema definition is ](https://github.com/nebari-dev/nebari-self-registration/blob/98be35af295c0bef1cbdde167ac4a7a62d0543b3/src/nebari_plugin_self_registration/__init__.py#L31)values: Optional[Dict[str, Any]] = {}
( this is because the whole block is just passed to helm chart values so it would be a headache to define the whole thing explicilty)

I'm unsure about how all of the Pydantic + pluggy schema building magic works, but this is my conclusion since changing fields that are key-value pairs (for example self_registration.namespace ) doesn't encounter this problem

I think there are other values fields in the core Nebari schema that will have the same problem this extension is having.  I'll try and recreate the issue with one of those and comment here.

### Expected behavior

Changing the values of anything under the `values` key should be allowed

MacOS Sequoia (Apple Silicon ARM)

Deploy Nebari 2024.9.1 with the nebari-self-registration extension plugin installed and config in the `values` block like in the description above

Try to change any of the content in `self_registration.values` and redeploy

### Command output

```bash session
AttributeError: 'CommentedMap' object has no attribute 'model_fields'
```

Nebari 2024.9.1
nebari-plugin-self-registration 0.0.9

Python 3.12.5

### Compute environment

None

### Integrations

_No response_

### Anything else?

_No response_

kenafoster:
## Reference Issues or PRs
Fixes #2767

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [X] Did you add new tests?

## How to test this PR?

Deploy Nebari with a configuration that includes an "overrides" or "values" block whose structure is not defined in the schema (this occurs when passing values to Helm charts).  One example is:

```
jupyterhub:
  overrides:
    singleuser:
      extraEnv:
        ENV: "test"
```

Change the value of the "ENV" key and redeploy.  (Nebari 2024.9.1 would error out)


Adam-D-Lewis:
LGTM!

Adam-D-Lewis (approved):
LGTM!  Thanks @kenafoster!

PR #2837, opened 2024-11-08
Remove explicit branch inputs from cloud integration test workflows in GHA

### LINKED ISSUE #2836: [ENH] - Remove unnecesary branch input in GHA cloud integration tests
### Feature description

We are currently adding `branch` as an input for our cloud provider workflows as a way to specify the _Nebari branch to deploy, test, destroy_. For example: https://github.com/nebari-dev/nebari/blob/855aa14ecc32625a98ec01b017acfc8a4af3c017/.github/workflows/test_aws_integration.yaml#L6-L12

This is not needed as [workflows that are run manually](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow#running-a-workflow) (via the `workflow_dispatch` trigger) already offer a way to select the branch from which they should run:

image

### Value and/or benefit

Not having duplicate UI items to specify a branch from which to run the workflow (i.e., a dropdown and a textbox) will make things more intuitive for developers triggering the workflows manually.

### Anything else?

_No response_

marcelovilla:
## Reference Issues or PRs
Closes #2836 

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

## How to test this PR?
You won't be able to see the branch input textbox gone from the GitHub Actions UI until we merge the PR to main. However, you can compare this approach to our [local upgrade tests](https://github.com/nebari-dev/nebari/blob/855aa14ecc32625a98ec01b017acfc8a4af3c017/.github/workflows/test_local_upgrade.yaml), which don't specify the branch as an input and can still be run from an arbitrary branch:

image

### DIFF
--- a/.github/workflows/test_aws_integration.yaml
+++ b/.github/workflows/test_aws_integration.yaml
@@ -25,7 +25,7 @@ on:
 
 env:
   AWS_DEFAULT_REGION: "us-west-2"
-  NEBARI_GH_BRANCH: ${{ github.event.inputs.branch || 'main' }}
+  NEBARI_IMAGE_TAG: ${{ github.event.inputs.image-tag || 'main' }}
   TF_LOG: ${{ github.event.inputs.tf-log-level || 'info' }}∏
 
 jobs:
--- a/.github/workflows/test_aws_integration.yaml
+++ b/.github/workflows/test_aws_integration.yaml
@@ -5,11 +5,6 @@ on:
     - cron: "0 0 * * MON"
   workflow_dispatch:
     inputs:
-      branch:
-        description: 'Nebari branch to deploy, test, destroy'
-        required: true
-        default: main
-        type: string
       image-tag:
         description: 'Nebari image tag created by the nebari-docker-images repo'
         required: true
@@ -31,7 +26,6 @@ on:
 env:
   AWS_DEFAULT_REGION: "us-west-2"
   NEBARI_GH_BRANCH: ${{ github.event.inputs.branch || 'main' }}
-  NEBARI_IMAGE_TAG: ${{ github.event.inputs.image-tag || 'main' }}
   TF_LOG: ${{ github.event.inputs.tf-log-level || 'info' }}∏
 
 jobs:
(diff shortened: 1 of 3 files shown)


Adam-D-Lewis:
From the [issue](https://github.com/nebari-dev/nebari/issues/2836).  
> This is not needed as [workflows that are run manually](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow#running-a-workflow) (via the workflow_dispatch trigger) already offer a way to select the branch from which they should run:

If I understand you correctly you are saying the blue and green underlined options are identical, but that isn't the case.
image

The blue option is to affect from which branch the workflow runs.  If you are modifying the workflow itself, this is a useful option to test your changes.  You may still want the test in your workflow to checkout the main branch of the Nebari codebase however.  This isn't possible without both options.

I think the customization we currently have is useful and doesn't hurt anything as long as good options are used for the defaults which I think they are.

marcelovilla:
> You may still want the test in your workflow to checkout the main branch of the Nebari codebase however. This isn't possible without both options.

Are you referring to testing the workflow from another branch but have that deploy Nebari from the `main` branch? Honestly, I don't think this is a very common scenario but I get that if it ever happens to be the case, it would certainly be useful to have a way of choosing branches separately.

The reason I opened this PR is because I think is more intuitive to use the built-in branch selector (like we're doing in other tests), instead of adding an additional on our own. Furthermore, I'm not sure if it might be prone-error. What happens if you select a branch on the drop-down and write another on the textbox? Which one gets checked out? I'm assuming the latter.

Adam-D-Lewis:
Yeah, I agree the situation I described before is not common.

I looked up [the docs for the checkout action](https://github.com/actions/checkout#usage).
```
    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
```

I was under the impression that `main` would always be checked out if we didn't set a branch, but I now see that's not the case.  I'm okay with this PR then.

Adam-D-Lewis (approved):
LGTM, thanks @marcelovilla!

PR #2839, opened 2024-11-10
[ENH] - Include "--attempt-fixes" flag from Nebari upgrade CLI in upgrade steps logic

### LINKED ISSUE #2761: [ENH] - Include "--attempt-fixes" flag from Nebari upgrade CLI in upgrade steps logic
### Feature description

We are currently exposing an `--attempt-fixes` flag in the `nebari upgrade` subcommand: https://github.com/nebari-dev/nebari/blob/9b1310b33e89c2c11c3b39128ec792ca80342486/src/_nebari/subcommands/upgrade.py#L19-L23

This parameter can be accessed from the `UpgradeStep` class, where we define specific upgrade logic for each upgrade step between contiguous versions: https://github.com/nebari-dev/nebari/blob/9b1310b33e89c2c11c3b39128ec792ca80342486/src/_nebari/upgrade.py#L52-L111

However, we're not using this flag in the most recent upgrade steps. We need to ensure that upgrade steps use this flag and define a logical path to upgrade without user input.

### Value and/or benefit

Recently, https://github.com/nebari-dev/nebari/pull/2780 got merged, implementing the logic to add automatic upgrade tests to our GHA workflows. Because tests are running automatically, we cannot rely on user input, so they are relying on  passing the `--attempt-fixes` flag when calling the `nebari upgrade` command: https://github.com/nebari-dev/nebari/blob/364c9e3d5b37286ab32f9b13542a1d3e2cd6fb0a/.github/workflows/test_local_upgrade.yaml#L73

### Anything else?

_No response_

smokestacklightnin:
## Reference Issues or PRs

This resolves #2761 by passing the `attempt_fixes` variable from the command line option to the upgrade methods.
close #2761

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

## How to test this PR?

Run `nebari upgrade --attempt-fixes` and verify that the upgrade is successful without prompts from the user.


src/_nebari/upgrade.py
```diff
@@ -230,7 +230,15 @@ def requires_nebari_version_field(self):
...
 
-    def upgrade_step(self, config, start_version, config_filename, *args, **kwargs):
+    def upgrade_step(
+        self,
+        config,
+        start_version,
+        config_filename,
+        *args,
+        attempt_fixes=False,
+        **kwargs,
```
marcelovilla, line 240: We should already have access to `attempt_fixes` in `kwargs`. See:

https://github.com/nebari-dev/nebari/blob/3f2523aae801fa5d09c414aec6fe1eb237effc0f/src/_nebari/upgrade.py#L485-L494

Did you add this because that does not work as expected?

src/_nebari/upgrade.py
```diff
@@ -329,11 +342,11 @@ def replace_image_tag(s: str, new_version: str, config_path: str) -> str:
             if current_tag == new_version:
                 return s
             loc = f"{config_path}: {image_name}"
-            response = Prompt.ask(
-                f"\nDo you want to replace current tag [green]{current_tag}[/green] with [green]{new_version}[/green] for:\n[purple]{loc}[/purple]? [Y/n] ",
-                default="Y",
+            response = attempt_fixes or Confirm.ask(
```
marcelovilla, line 345: I thought this would still prompt the user for confirmation but it seems that if `attempt_fixes` is `True`, the right hand side of the statement won't even be executed, right?

src/_nebari/upgrade.py
```diff
@@ -364,7 +377,11 @@ def set_nested_item(config: dict, config_path: list, value: str):
             config[config_path[-1]] = value
 
         def update_image_tag(
-            config: dict, config_path: str, current_image: str, new_version: str
+            config: dict,
+            config_path: str,
+            current_image: str,
+            new_version: str,
+            attempt_fixes: bool = attempt_fixes,
```
marcelovilla, line 384: If `attempt_fixes` is exposed in `kwargs`, you can use it from there. I wouldn't even use a default value, but rather pass it explicitly in all the further calls to this function by grabbing it from `kwargs`

src/_nebari/upgrade.py
```diff
@@ -640,11 +685,11 @@ def _version_specific_upgrade(
             ""
         )
 
-        continue_ = Prompt.ask(
+        continue_ = attempt_fixes or Confirm.ask(
```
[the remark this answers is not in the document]
marcelovilla: @smokestacklightnin did you see the approach I implemented? You'd need to add the logic to automatically delete the relevant CRDs here if the `attempt_fixes` is passed. We would need an extra step or we would need to reframe the question, though.

src/_nebari/upgrade.py
```diff
@@ -678,15 +723,21 @@ class Upgrade_2023_7_2(UpgradeStep):
 
     @override
     def _version_specific_upgrade(
-        self, config, start_version, config_filename: Path, *args, **kwargs
+        self,
+        config,
+        start_version,
+        config_filename: Path,
+        *args,
+        attempt_fixes=False,
```
marcelovilla, line 731: Same thing as above: `attempt_fixes` should already be exposed in `kwargs`

src/_nebari/upgrade.py
```diff
@@ -1099,16 +1159,15 @@ def _version_specific_upgrade(
...
-                continue_ = Prompt.ask(
+                _ = attempt_fixes or Prompt.ask("Hit enter to continue")
+                continue_ = attempt_fixes or Confirm.ask(
                     f"Have you backed up your custom dashboards (if necessary), deleted the {daemonset_name} daemonset and updated the kube-prometheus-stack CRDs?",
-                    choices=["y", "N"],
-                    default="N",
+                    default=False,
                 )
-                if not continue_ == "y":
+                if not continue_:
```
marcelovilla, line 1170: No need to use `attempt_fixes` here as it will only be evaluated when `attempt_fixes` is `False` or the user wants to run the commands manually

src/_nebari/upgrade.py
```diff
@@ -1179,8 +1237,9 @@ def _version_specific_upgrade(
                     },
                     indent=4,
                 )
-                text += "\n\nHit enter to continue"
-                Prompt.ask(text)
+                rich.print(text)
+                if not attempt_fixes:
```
marcelovilla, line 1241: Same as above: we don't need `attempt_fixes` here as it will only run if it's `False` and at that point we don't really care

marcelovilla (changes requested):

marcelovilla:
@smokestacklightnin can you also make sure the tests are passing?

src/_nebari/upgrade.py
smokestacklightnin: > Did you add this because that does not work as expected?

No, I should have used it from `kwargs`. Done.

src/_nebari/upgrade.py
smokestacklightnin: Fixed

src/_nebari/upgrade.py
smokestacklightnin: >  if attempt_fixes is True, the right hand side of the statement won't even be executed, right?

Correct

src/_nebari/upgrade.py
smokestacklightnin: Fixed

src/_nebari/upgrade.py
marcelovilla: This is exactly what we need, thanks. I was just wondering if it would still try to prompt the user but that does not seem to be the case

src/_nebari/upgrade.py
```diff
@@ -629,27 +650,93 @@ def _version_specific_upgrade(
...
+                    if e.status == 404:
+                        rich.print(
+                            f"Service account [yellow]{sa}[/yellow] not found. Ignoring."
+                        )
+                    else:
+                        raise e
+                else:
+                    rich.print(
+                        f"Successfully removed service account [green]{sa}[/green]"
+                    )
```
smokestacklightnin, line 706: @marcelovilla I was able to adapt [what you did](https://github.com/smokestacklightnin/nebari/blob/88be8a4db8f2e4b6ae2dde6baf24f7c563f66d56/src/_nebari/upgrade.py#L1154-L1181)

marcelovilla:
@smokestacklightnin can you make sure tests are passing? Then, I'll review the whole PR

smokestacklightnin:
I'm getting the following test failure, which is confusing me:

```

**Comment by smokestacklightnin** [comment]
@marcelovilla All tests are passing. There is only one test that fails because it needs monkeypatching (CC: @viniciusdc). I have `xfailed` it for now.

**Comment by marcelovilla** [comment]
> All tests are passing. There is only one test that fails because it needs monkeypatching (CC: @viniciusdc). I have [xfailed](https://github.com/nebari-dev/nebari/pull/2839/files#diff-67d09a1e15674ea4c0a84542d85d4a339a6b9d1b2a38438ffe2122944956b930) it for now.

@smokestacklightnin I think that should have been addressed by https://github.com/nebari-dev/nebari/pull/2879. Or am I confusing tests here?

File: src/_nebari/upgrade.py [hunk 9]
```diff
@@ -1283,13 +1497,10 @@ def _version_specific_upgrade(
...
 
         # Prompt the user for role assignment (if yes, transforms the response into bool)
-        assign_roles = (
-            Prompt.ask(
-                "[bold]Would you like Nebari to assign the corresponding role/scopes to all of your current groups automatically?[/bold]",
-                choices=["y", "N"],
-                default="N",
-            ).lower()
-            == "y"
+        # This needs to be monkeypatched and will be addressed in a future PR. Until then, this causes test failures.
```
marcelovilla, line 1500: I believe this has been solved in https://github.com/nebari-dev/nebari/pull/2879/files

marcelovilla (dismissed):

marcelovilla (commented):
@smokestacklightnin I just tested an upgrade and it seems the flag is working as expected 🚀 

The only thing I want to sort out before approving the PR is the test you xfailed for now. Is that still needed?

smokestacklightnin:
@marcelovilla I was able to to reproduce [the error](https://github.com/nebari-dev/nebari/actions/runs/12595215940/job/35104059569) locally by first running `$ rm -rf ~/.kube`. 

I was able to [monkeypatch the parts that were causing errors](https://github.com/nebari-dev/nebari/pull/2839/commits/c1ac429f755b0df5f2f216916a6b6fcc0df28127), and now all unit tests are passing

marcelovilla (approved):
Looks good, thanks! 🚀

PR #2840, opened 2024-11-11
allow passing X.XX or X.XX.XX as k8s versions

### LINKED ISSUE #2822: [BUG] - Invalid `kubernetes-version` provided error on GCP because of GKE patch release version
### Describe the bug

On several GCP deployments, I've encountered a validation issue regarding the Kubernetes version when trying to re-deploy Nebari after changing my config file. For example, this is the latest error I got:

```
ValidationError: 1 validation error for ConfigSchema
google_cloud_platform
  Value error,
Invalid `kubernetes-version` provided: 1.28.14-gke.1004000.
Please select from one of the following supported Kubernetes versions:
['1.28.14-gke.1099000', '1.28.14-gke.1217000', '1.28.14-gke.1340000',
'1.28.14-gke.1376000', '1.28.15-gke.1015000', '1.28.15-gke.1020000',
'1.29.10-gke.1043000', '1.29.10-gke.1054000'] or omit flag to use latest
Kubernetes version available. [type=value_error, input_value={'project':
'quansight-tr... 0, 'max_nodes': 1000}}}, input_type=CommentedMap]
    For further information visit https://errors.pydantic.dev/2.4/v/value_error
```

As you can see, I was using version `1.28.14-gke.1004000` for my deployment but now Nebari is forcing me to upgrade to another version, with many options just having a different GKE patch release number. While I appreciate [these GKE patch releases can include security updates and bug fixes](https://cloud.google.com/kubernetes-engine/versioning#versioning_scheme), it seems Nebari (or GKE) is deprecating them very fast, forcing users with a GCP cluste to update their k8s version fairly often (a couple weeks or months) apart. This is in some cases an incovenience because a small change in the config implies a k8s version upgrade, which takes longer and has some down time associated with it.

I haven't encountered this issue with AWS, where k8s versioning scheme only includes a major version, a minor version and a patch release.

### Expected behavior

I'd expect Nebari to force a k8s version upgrade only when it's really necessary, like for example a version entering the extended support period or end-of-life.

Not relevant

This is a rather cumbersome problem to reproduce as it depends on GKE version support. But the general idea is:

1. Run `nebari init --guided-init` to initialize a GCP deployment. This will grab a default k8s version.
2. Deploy Nebari
3. After a few weeks or months, do a small change on your config and try to run `nebari deploy -c nebari-config.yaml`
4. If you run into this issue, you'll get a message complaining that your current k8s version does not match the list of allowed options

### Command output

_No response_

_No response_

### Compute environment

GCP

### Integrations

_No response_

### Anything else?

This is [the output I get](https://pastebin.com/NB7ajh48) when running part of the logic found in Nebari to get available GKE k8s versions for GKE: https://github.com/nebari-dev/nebari/blob/9b1310b33e89c2c11c3b39128ec792ca80342486/src/_nebari/provider/cloud/google_cloud.py#L55-L65

The issue arises from version `1.28.14-gke.1004000` not being part of `valid_master_versions`. However, it is part of `valid_node_versions`, so I think it should be reasonable to expect Nebari not to force upgrading my nodes to a newer version.

dcmcand:
## Reference Issues or PRs
Fixes #2822 

Allows specifiying either major and minor versions, major minor and patch versions, or full versions of kubernetes for gcp. Also changes guided init so that if no version is specified, the latest patch version is used.

- [X] New feature (non-breaking change which adds a feature)
- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

## How to test this PR?

1. Run either `nebari init --guided-init` and don't specify a kubernetes version or `nebari init gcp -p your_project_name -d your_domain_name.org` 
2.  To validate, you can deploy to GCP or locally using `1.X`, `1.X.X`, and `1.X.X-gke.x`
2. Use

### DIFF
--- a/src/_nebari/provider/cloud/google_cloud.py
+++ b/src/_nebari/provider/cloud/google_cloud.py
@@ -64,6 +64,16 @@ def kubernetes_versions(region: str) -> List[str]:
     return filter_by_highest_supported_k8s_version(supported_kubernetes_versions)
 
 
+def get_patch_version(full_version: str) -> str:
+    return full_version.split("-")[0]
+
+
+def get_minor_version(full_version: str) -> str:
+    patch_version = get_patch_version(full_version)
+    parts = patch_version.split(".")
+    return f"{parts[0]}.{parts[1]}"
+
+
 def cluster_exists(cluster_name: str, region: str) -> bool:
     """Check if a GKE cluster exists."""
     credentials, project_id = load_credentials()
--- a/src/_nebari/provider/cloud/google_cloud.py
+++ b/src/_nebari/provider/cloud/google_cloud.py
@@ -57,7 +57,7 @@ def kubernetes_versions(region: str) -> List[str]:
     credentials, project_id = load_credentials()
     client = container_v1.ClusterManagerClient(credentials=credentials)
     response = client.get_server_config(
-        name=f"projects/{project_id}/locations/{region}"
+        name=f"projects/{project_id}/locations/{region}", timeout=300
     )
     supported_kubernetes_versions = response.valid_master_versions
 
--- a/src/_nebari/stages/infrastructure/__init__.py
+++ b/src/_nebari/stages/infrastructure/__init__.py
@@ -8,7 +8,7 @@ import sys
 import tempfile
 from typing import Annotated, Any, Dict, List, Literal, Optional, Tuple, Type, Union
 
-from pydantic import Field, field_validator, model_validator
+from pydantic import ConfigDict, Field, field_validator, model_validator
 
 from _nebari import constants
 from _nebari.provider import terraform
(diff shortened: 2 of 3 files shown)


src/_nebari/provider/cloud/google_cloud.py
```diff
@@ -57,7 +57,7 @@ def kubernetes_versions(region: str) -> List[str]:
     credentials, project_id = load_credentials()
     client = container_v1.ClusterManagerClient(credentials=credentials)
     response = client.get_server_config(
-        name=f"projects/{project_id}/locations/{region}"
+        name=f"projects/{project_id}/locations/{region}", timeout=300
```
dcmcand, line 60: I was running into GRPC timeouts using the default 20s context timeout, so I set it to 5 minutes

src/_nebari/stages/infrastructure/__init__.py
```diff
@@ -359,6 +359,7 @@ class GCPNodeGroup(schema.Base):
 
 
 class GoogleCloudPlatformProvider(schema.Base):
+    model_config = ConfigDict(coerce_numbers_to_str=True)
```
dcmcand, line 362: It was interpreting `1.29` as a float instead of a string, so I had to make it explicitly use a string

viniciusdc (approved):

PR #2842, opened 2024-11-11
Set `launch_template.ami_id` attrs to private

viniciusdc:
This PR will disable the configuration of `node_groups.*.launch_template.ami_id` specifically, until #2832 is addressed. 

## What does this implement/fix?

_Put a `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds a feature)
- [ ] Breaking change (fix or feature that would cause existing features not to work as expected)
- [ ] Documentation Update
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Other (please describe):

## Testing

- [ ] Did you test the pull request locally?
- [ ] Did you add new tests?

PR #2621, opened 2024-08-13
Add aws_launch_template & config vars to enable options to specify pre-bootstrap commands and/or custom AMI for EKS nodes

joneszc:
## Reference Issues or PRs
Fixes [#2603](https://github.com/nebari-dev/nebari/issues/2603)
Fixes [#2604](https://github.com/nebari-dev/nebari/issues/2604)

## What does this implement/fix?

_Put a `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds a feature)
- [ ] Breaking change (fix or feature that would cause existing features not to work as expected)
- [ ] Documentation Update
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Other (please describe):

## Testing

- [x] Did you test the pull request locally?
- [ ] Did you add new tests?

PR #2856, opened 2024-11-15
Disable AWS `launch_template` from nebari-config schema

viniciusdc:

This PR disables the user-facing object from the nebari-config schema while still keeping the logic introduced initially in #2621. 

### Motivation

The #2842 attempted to disable the `ami_id` setting from the `launch_template` configuration under the node groups. However, this change caused Terraform deployment issues because the aws node_group relied on the `ami_id` attribute internaly, leading to a failed Nebari deployments when it was removed in the linked PR.

Additionally, we've encountered recent internal connection issues with the nodes using just the `pre_bootstrap_command.` These issues indicated that the current configuration for `launch_template` is not functioning correctly.

It also includes:
- a custom error message in the schema validator for letting users know this is a temporary removal;
- A better docstring to the auxiliary `construct_aws_ami_type` function since the choice for the CUSTOM instance type depends on the presence of both `launch_template` and `ami_id`.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

## How to test this PR?

- init an aws config file
- (optional) run a deployment;
- Include a new node group with the `launch_template` field. This would simulate a scenario where the user has such a config already (for example):
```yaml
      launch_template:
        # Replace with your custom AMI ID
        ami_id: ami-0abcdef1234567890
        # Command to run before the node joins the cluster
        pre_bootstrap_command: |
          #!/bin/bash
          # This script is executed before the node is bootstrapped
          # You can use this script to install additional packages or configure the node
          # For example, to install the `htop` package, you can run:
          # sudo apt-get update
          # sudo apt-get install -y htop"
```
Though, just adding `launch_template` will trigger the error.
- run `nebari render` or `nebari validate` to trigger the load of the schemas. This should lead you to a ValueError

### DIFF
--- a/src/_nebari/stages/infrastructure/__init__.py
+++ b/src/_nebari/stages/infrastructure/__init__.py
@@ -8,7 +8,7 @@ import sys
 import tempfile
 from typing import Annotated, Any, Dict, List, Literal, Optional, Tuple, Type, Union
 
-from pydantic import ConfigDict, Field, PrivateAttr, field_validator, model_validator
+from pydantic import ConfigDict, Field, field_validator, model_validator
 
 from _nebari import constants
 from _nebari.provider import terraform
--- a/src/_nebari/stages/infrastructure/__init__.py
+++ b/src/_nebari/stages/infrastructure/__init__.py
@@ -136,7 +136,7 @@ class AWSAmiTypes(str, enum.Enum):
 
 class AWSNodeLaunchTemplate(schema.Base):
     pre_bootstrap_command: Optional[str] = None
-    _ami_id: Optional[str] = PrivateAttr(default=None)
+    ami_id: Optional[str] = None
 
 
 class AWSNodeGroupInputVars(schema.Base):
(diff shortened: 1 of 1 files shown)


viniciusdc:
## Tested as follows:

This PR was tested in a previously deployed nebari instance on AWS that had node groups using the `launch_template`. When attempting the re-deployment from this version, the following (expected) message occurs:

image

As the `launch_template` configuration is heavily coupled with the instance type used for both normal nodes and GPU utilization, I also tested to make sure both could be fully utilized:
- Usual node group with GPU showing correctly assigned AWS instance type 
```yaml
    gpu:
      instance: g4dn.xlarge
      min_nodes: 0
      max_nodes: 5
      gpu: true
      single_subnet: false
      permissions_boundary:
```
image

dcmcand (approved):

viniciusdc:
I am testing a fresh init and deploy just to be safe

PR #2887, opened 2024-12-17
Add Azure AAD/EntraID access settings

viniciusdc:

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

### DIFF
--- a/src/_nebari/stages/infrastructure/__init__.py
+++ b/src/_nebari/stages/infrastructure/__init__.py
@@ -90,6 +90,27 @@ class AzureNodeGroupInputVars(schema.Base):
     max_nodes: int
 
 
+class AzureRBAC(schema.Base):
+    """
+    Represents the configuration for Azure Active Directory Role-Based Access Control
+    (RBAC) integration in a Kubernetes cluster.
+
+    Attributes:
+        enabled (bool): Indicates whether Azure AD-based Role-Based Access Control is
+        enabled.
+        managed (bool): Specifies if the Azure AD integration is managed by Azure.
+            When set to True, Azure creates and manages the Service Principal used for
+            integration.
+        admin_group_object_ids (List[str]): A list of Object IDs of Azure AD groups assigned
+            administrative roles on the cluster. This property is only applicable when
+            `managed` is set to True.
+    """
+
+    enabled: bool
+    managed: bool
+    admin_group_object_ids: List[str]
+
+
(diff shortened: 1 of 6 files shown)


viniciusdc:
I just need to rename the enabled field since its misleading, its intent is to enable the usage of RBAC roles from Azure instead of the default kuberntes behavior which is an authorization change, works different then the others enabled flags on nebari right now (this does not refer to the actual toggle of this property in the cluster)

dcmcand:
@viniciusdc is this ready for review?

viniciusdc:
I discovered an issue with the previous `azure_addon_policy` settings: Terraform did not expect a null value in one case. This PR will also address that oversight. 

Subsequently, while re-testing on a different machine (with an Azure RBAC deployment already confirmed to work), I encountered a new strange error related to the `kube_config` artifact in the AKS cluster resource. It returned a key error when retrieving the Kubernetes credentials from the cluster. This may indicate the need to use the `exec` block for the Kubernetes provider when AAD/RBAC is enabled. I’m still investigating the problem, as this behavior wasn’t observed before.

viniciusdc:
My assumption seems correct; after manually enabling `local_accounts` through the Azure portal, the deployed could continue without encountering the problem above. I already have a possible remedy for this scenario, which I will attempt soon.

viniciusdc:
Currently addressing an inconsistency with the admin kubeconfig and the exec command, while making sure that the appropriate python data object under tf_objects is correctly handled as well. The main issue, seems to be the current level of nesting associated with the credentials outputting logic, to properly handle the on/off variation of enabling RBAC we need to workaround uinsg a new child module for returning the appropriate credentials attributes later on used by the kubernetes provider, this is a bit tricky, since they change not only in naming but also in functionality.

src/_nebari/stages/tf_objects.py
```diff
@@ -22,6 +22,31 @@ def NebariKubernetesProvider(nebari_config: schema.Main):
...
+    #                     "get-credentials",
+    #                     "--resource-group",
+    #                     nebari_config.azure.resource_group_name,
+    #                     "--name",
+    #                     nebari_config.azure.cluster_name,
+    #                     "--overwrite-existing",
+    #                 ],
+    #             },
+    #         ),
+    #     )
```
Adam-D-Lewis, line 49: need to modify this

Issue #2760, opened 2024-10-09
[ENH] - add conda-store black-box testing to ci


dcmcand:
### Feature description

Conda-store has black box user tests available at https://github.com/conda-incubator/conda-store/tree/main/conda-store-server/tests/user_journeys. 

### Value and/or benefit

Adding this to our CI pipeline will ensure that conda-store functionality works.

PR #2895, opened 2025-01-06
Add workflow to run conda-store user journey tests

soapy1:
## Reference Issues or PRs

fixes https://github.com/nebari-dev/nebari/issues/2760

* Runs conda-store integration tests as part of `test_local_integration`
* Adds the option to specify a `--group` flag when adding a keycloak user

To create a keycloak user with access to a particular group, run the command
```
nebari keycloak adduser --user username password --config nebari-config.yaml --groups mygroup

```

- [ ] Documentation Update
- [ ] Build related changes
- [x] Other (please describe): CI change


viniciusdc:
Since these tests depend on a live cluster, I suggest extending the already existing local integration tests with this extra check https://github.com/nebari-dev/nebari/blob/main/.github/workflows/test_local_integration.yaml -- since it uses CiRun under the hood you should not be affected by any flakiness of the GH runner in case of constrained resources.

aktech:
My 2 cents on this: Iterating on these tests can be painful as it takes 20+ min to get the feedback (assuming deployment succeeded), I have found https://github.com/marketplace/actions/debugging-with-tmate to be quite useful to ssh into the runner and debug failures.

soapy1:
Also depends on https://github.com/conda-incubator/conda-store/pull/1050

.github/workflows/test_local_integration.yaml
```diff
@@ -144,6 +145,58 @@ jobs:
...
+        run: |
+          which python
+          # install conda-store-server
+          python -m pip install conda-store/conda-store-server
+
+      - name: Get conda-store token from terraform state
+        id: conda-store-token
+        working-directory: ${{ steps.init.outputs.directory }}
+        run: |
+          kubectl get secret -n default tfstate-default-${{ steps.init.outputs.project }}-dev-07-kubernetes-services --template={{.data}} | cut -d ":" -f2 | cut -d "]" -f1 | base64 --decode | gzip -d --to-stdout > terraform_state
```
soapy1, line 179: I'm not very happy about this one. The user journey tests need a token to auth against conda-store. I tried a few different approaches to finding a token or generating a token but this is the only way that seemed to consistently work and could be done in CI. Definitely open to suggestions on how to make this better.

aktech: I have not yet looked into the whole pr yet, just looking at this comment, is this any good?

https://github.com/nebari-dev/nebari/blob/43e63fe7f21c68ec6b4ecf7a117b62f8937691fe/tests/common/conda_store_utils.py#L8

https://github.com/nebari-dev/nebari/blob/43e63fe7f21c68ec6b4ecf7a117b62f8937691fe/tests/common/conda_store_utils.py#L34

soapy1: hmmm, I did try this approach (but with curl). I think I was maybe missing this `credentialId` piece though. Will give that another try. Thanks for the tip!

soapy1: ok so, i think this is not going to work because the conda-store integration tests need an admin token. However, the logged in user does not have admin permissions, so can not create an admin token. For reference, the logged in user has the permissions
```
{
  "status": "ok",
  "data": {
    "authenticated": true,
    "primary_namespace": "global",
    "entity_permissions": {
      "default/*": [
        "environment::read",
        "namespace-role-mapping::read",
        "namespace::read"
      ],
      "filesystem/*": [
        "environment::read",
        "namespace-role-mapping::read",
        "namespace::read"
      ]
    },
    "entity_roles": {
      "default/*": [
        "viewer"
      ],
      "filesystem/*": [
        "viewer"
      ]
    },
    "expiration": "20[25](https://github.com/nebari-dev/nebari/actions/runs/13120537340/job/36605155832?pr=2895#step:19:26)-02-04T19:04:11Z"
  },
  "message": null
}
```

## other options
1. Another ok option might be to seed this nebari deployment with a service token for the test. Looking around the nebari code base, it doesn't look like there are any mechanisms for that.  

2. ?

viniciusdc:
Hey @soapy1 no rush on this, but could you update us on this? Just considering adding to our next `Feb` release

soapy1:
Heya @viniciusdc I've gotten a bit stuck on this one. I'm not able to find a way to pull an admin token for conda-store from nebari in order to run the tests (full notes in this comment - https://github.com/nebari-dev/nebari/pull/2895#discussion_r1939911006).

Things that I've tried:
* pulling a service token from the terraform state - this works but is a little janky
* logging in with the test user and generating a token - this user does not have enough permissions to create an admin token
* logging in with the root keycloak user to generate a token - this user also doesn't have enough permissions to create an admin token

Other options I've considered (but haven't tried yet):
* is there a way to seed a new service token for conda store (like a config option to add some service tokens)
* is there a way to use the keycloak api to add a user with admin permissions, and then generate a token from there

Definitely open to other ideas if you have a suggestion!

viniciusdc:
I left a comment to your thread there, but basically I think your best options are those two:
> is there a way to seed a new service token for conda store (like a config option to add some service tokens)
is there a way to use the keycloak api to add a user with admin permissions, and then generate a token from there

1. For the first one, conda-store has a service token attribute that could be used, though this would require some extension to allow this to be customizable by the user to include new services:

https://github.com/nebari-dev/nebari/blob/ecaa94b32a341a722e84ed0d77b46ec7293c6690/src/_nebari/stages/kubernetes_services/__init__.py#L446-L448

https://github.com/nebari-dev/nebari/blob/ecaa94b32a341a722e84ed0d77b46ec7293c6690/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/conda-store/server.tf#L31-L33

Used for example here:
https://github.com/nebari-dev/nebari/blob/ecaa94b32a341a722e84ed0d77b46ec7293c6690/src/_nebari/stages/kubernetes_services/template/jupyterhub.tf#L181

2. I think this one might be easier to get going, nebari already has a command to create users from CLI -- `nebari keycloak --adduser`, which could be used for creating the new user. Right now, there is no way of programmatically changing the group, we can have a script under tests that do this:
```python
try:
    # Get user ID
    users = keycloak_admin.get_users({"username": USER_TO_ADD})
    if not users:
        print(f"User '{USER_TO_ADD}' not found.")
        exit(1)
    
    user_id = users[0]["id"]
    
    # Get group ID
    groups = keycloak_admin.get_groups()
    group_id = next((g["id"] for g in groups if g["name"] == "superadmin"), None)
    
    if not group_id:
        print(f"Group 'superadmin' not found.")
        exit(1)
    
    # Add user to group
    keycloak_admin.group_user_add(user_id, group_id)
    print(f"User '{USER_TO_ADD}' successfully added to 'superadmin' group.")

except KeycloakError as e:
    print(f"Error: {e}")
```

for the main API `KeycloakAdmin` class, you might be able to use this function 
https://github.com/nebari-dev/nebari/blob/ecaa94b32a341a722e84ed0d77b46ec7293c6690/src/_nebari/keycloak.py#L104

soapy1:
> I wonder if this should be configured directly trough nebari's CLI anyway

ya, I was thinking about adding this. Something like `nebari keycloak  adduser -c nebari-config.yaml --user test password --role superadmin`. I think it would be good to have some mechanism so that users can't create a user with more permissions than they ought to be able to.

> so it would not be a security overlook to allow the group in which the user will be created to appear as a custom flag as well

could make a v1 of this be something like only root users are allowed to set a role?

src/_nebari/keycloak.py
```diff
@@ -29,7 +29,10 @@ def do_keycloak(config: schema.Main, *args):
 
         username = args[1]
         password = args[2] if len(args) >= 3 else None
-        create_user(keycloak_admin, username, password, domain=config.domain)
+        groups = args[3] if len(args) >= 4 else None
```
soapy1, line 32: I'm not a fan of this approach to passing arguments around but did this to follow the pattern already defined in this part of the code. 
I would be happy to refactor this module to make it a more developer friendly if that's something the nebari team is into.

dcmcand: why don't you split that into a different pr. Thanks!

dcmcand (approved):

Adam-D-Lewis:
> @soapy1 this requires the nebari keycloak command update to work, right? If that's the case, @Adam-D-Lewis do you mind splitting that ENH into another PR to decouple from the service-account one? We can also keep it in here as well

Yeah, I can split that out into a separate PR today.

Adam-D-Lewis:
Here's the keycloak cli refactor PR - https://github.com/nebari-dev/nebari/pull/2968

dcmcand:
@soapy1 please let me know when this is ready to go so we can get it merged

marcelovilla:
This looks great @soapy1 ! 

In the spirit of achieving testing parity, I think it would be worthwhile to also run these tests on cloud deployments. That's beyond the scope of this PR but something to keep in mind when we [start running other test suites in cloud deployments](https://github.com/nebari-dev/nebari/issues/2975) too.

PR #2917, opened 2025-01-21
add jhub apps service account with admin permissions

Adam-D-Lewis:
## Reference Issues or PRs
In order for [startup apps](https://github.com/nebari-dev/jhub-apps/blob/494a5837ec18097d2b1f7e7ff0f9d317490ec7ab/jhub_apps/config_utils.py#L125) to be used by nebari, we need to create a service account with appropriate permissions to create the apps and share them with others.  

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

### DIFF
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/main.tf
@@ -23,6 +23,7 @@ locals {
   userscheduler_nodeselector_value = var.general-node-group.value
 }
 
+
 resource "kubernetes_secret" "jhub_apps_secrets" {
   metadata {
     name      = local.jhub_apps_secrets_name
@@ -34,6 +35,26 @@ resource "kubernetes_secret" "jhub_apps_secrets" {
   }
 
   type = "Opaque"
+}
+
+resource "keycloak_user" "jhub_apps_service_account" {
+  realm_id = var.realm_id
+  username = "jhub-apps-sa"
+  # email    = "jhub-apps-sa@${var.external-url}"
+  enabled = true # not sure if they need to be enabled,  TODO: check
+}
+
+data "keycloak_group" "admin_group" {
+  realm_id = var.realm_id
+  name     = "admin"
+}
+
+
+resource "keycloak_user_groups" "jhub_apps_service_account_groups" {
+  realm_id   = var.realm_id
+  user_id    = keycloak_user.jhub_apps_service_account.id
+  group_ids  = [data.keycloak_group.admin_group.id]
+  exhaustive = true # remove all other groups
 }
 
 locals {
--- a/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/versions.tf
+++ b/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/versions.tf
@@ -1,0 +1,9 @@
+terraform {
+  required_providers {
+    keycloak = {
+      source  = "mrparkers/keycloak"
+      version = "3.7.0"
+    }
+  }
+  required_version = ">= 1.0"
+}


Adam-D-Lewis:
I still need to check if the user needs to be enabled.  It'd be nice if the user didn't have admin permissions.

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/keycloak-client/outputs.tf
```diff
@@ -12,3 +12,10 @@ output "config" {
     callback_urls      = var.callback-url-paths
   }
 }
+
+output "client_role_ids" {
```
Adam-D-Lewis, line 16: I need to get the role id for allow-app-sharing-role so I can assign it to service-account-jhub-apps user.

Adam-D-Lewis:
I'm seeing some issues right now.  The issue is the user I've created needs to have logged in before the app can be started successfully.  In the current design, the "user" is meant to act more like a service account and not log in.  The issues I've seen so far are in the code that is run by the Spawner to [set preferred username](https://github.com/nebari-dev/nebari/blob/06f2030888b0f149df0dc2b90560d1a40f61f8f6/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/02-spawner.py#L17) and [render profiles](https://github.com/nebari-dev/nebari/blob/06f2030888b0f149df0dc2b90560d1a40f61f8f6/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/03-profiles.py#L556), but it's possible there are more.

The auth state for the "jhub-apps-sa" user is None before initial log in so an error is thrown once we try to access info in the auth_state object in those methods.

krassowski:
This might be too radical and not the kind of suggestion you are asking for, but could it be solved on the jhub-apps level? I mean jhub-apps is meant to be auth-provider agnostic so it should be possible to make this work without touching keycloak at all? This might be just my PTSD from figuring out keycloak piping last time speaking. I recall @aktech also had pleasure to work on keycloak integration - he may have better ideas.

krassowski:
> Some possible ideas on how to fix this:
>
>- if the auth_state is None, then go ask keycloak for the needed info? (login on behalf of the user maybe or some other way?)
>- log the "jhub-apps-sa" user in somehow on startup app creation?
>- I could add startup_app: True to the user options dict used by jhub_apps. I could then modify our problematic spawner code which requires `auth_state` to only run if the server starting up isn't a startup app.

If staying with this approach I would probably try the solutions in that exact order. I am not sure if the last one will work if you need to do anything beyond the configuration being set (i.e. whether server will actually spawn if you do not have `auth_state`).

aktech:
> This might be too radical and not the kind of suggestion you are asking for, but could it be solved on the jhub-apps level? I mean jhub-apps is meant to be auth-provider agnostic so it should be possible to make this work without touching keycloak at all?

That would be ideal, indeed. The way we are using the spawner here, expects the user to be logged in (hence populating `auth_state`) once before creating a server, which makes sense from jupyterhub pov as the servers are created by humans instead of robots. 

This (startup apps) works in jhub-apps (without nebari) by default unless the spawner needs `auth_state` (which is the case here). Since its a feature of jhub-apps to provide the ability to create init apps on startup regardless of Authenticator or spawner, this should be handled in jhub-apps if possible (this is a big if though), you might need to pass in some kind of keycloak auth details in the JApps config, that might make it possible.

We can try to see if we can somehow call the `authenticate` method of the Authenticator in jhub-apps for the user, to populate the auth state before starting startup apps, but yes this might not be possible at all, in that case what you suggested in this comment https://github.com/nebari-dev/nebari/pull/2917#issuecomment-2607625067, sounds like a good approach, and I agree with @krassowski the last one won't work, as you need groups info to create nfs mounts.

Also, another thing to note here, is supporting this in jhub-apps might be tricky (if its possible), its probably ok to just go for your approach and we can tackle it in jhub-apps later, if time is of essence here.

Adam-D-Lewis:
I think the ideal solution is to create the startup apps up using the jupyterhub-service-account which is service account associated with the jupyterhub Keycloak client.  We basically already do the authenticatation needed in Nebari's [KeycloakOAuthenticator](https://github.com/nebari-dev/nebari/blob/43e63fe7f21c68ec6b4ecf7a117b62f8937691fe/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/04-auth.py#L310-L326).  We just need to set auth_state for that service account.

Adam-D-Lewis:
I haven't been able to successfully [set auth state for the service account](https://github.com/nebari-dev/nebari/pull/2917/commits/2a3e49b68131914b069ca8bdc08f184bdeca6e76), but just looking up the service account's info during the spawner code in [this commit](https://github.com/nebari-dev/nebari/pull/2917/commits/5f9834ac92e9d49263dcc440967ce3c903aa7387) seems to work.

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/keycloak-client/main.tf
```diff
@@ -67,29 +67,54 @@ data "keycloak_realm" "master" {
...
 
-data "keycloak_openid_client" "realm_management" {
-  realm_id  = var.realm_id
-  client_id = "realm-management"
-}
 
-data "keycloak_role" "main-service" {
-  for_each = toset(var.service-account-roles)
+# Get client data for each service account client
+data "keycloak_openid_client" "service_clients" {
```
Adam-D-Lewis, line 72: Before we only allowed service accounts to get roles from the realm-management client.  This PR allows us to set roles by any client.  This functionality was needed to be able to set the allow-app-sharing-role on the jupyterhub service account.

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/02-spawner.py
```diff
@@ -11,9 +10,8 @@
 from kubespawner import KubeSpawner  # noqa: E402
 
 
-@gen.coroutine
-def get_username_hook(spawner):
-    auth_state = yield spawner.user.get_auth_state()
+async def get_username_hook(spawner):
```
Adam-D-Lewis, line 13: flyby: tornado coroutine -> native coroutine.  We don't need to use a tornado coroutine.

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/02-spawner.py
```diff
@@ -23,6 +21,13 @@ def get_username_hook(spawner):
     )
 
 
+async def pre_spawn_hook(spawner):
+    # if we are starting a service account pod, set/update auth_state
+    if spawner.user.name == "service-account-jupyterhub":
```
Adam-D-Lewis, line 26: make service account name a variable?

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/03-profiles.py
```diff
@@ -547,14 +546,12 @@ def preserve_envvars(spawner):
     return profile
 
 
-@gen.coroutine
-def render_profiles(spawner):
+async def render_profiles(spawner):
```
Adam-D-Lewis, line 549: flyby: tornado coroutine to native coroutine

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/04-auth.py
```diff
@@ -46,15 +84,15 @@ async def update_auth_model(self, auth_model):
         user_id = auth_model["auth_state"]["oauth_user"]["sub"]
         token = await self._get_token()
 
-        jupyterhub_client_id = await self._get_jupyterhub_client_id(token=token)
+        jupyterhub_client_uuid = await self._get_jupyterhub_client_uuid(token=token)
```
Adam-D-Lewis, line 87: flyby: rename jupyterhub_client_id to jupyterhub_client_uuid so devs don't get confused with jupyterhub's client_id which is different.

Adam-D-Lewis:
The local integration test seems to be failing for unrelated reasons.

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/04-auth.py
```diff
@@ -18,6 +19,12 @@ class KeyCloakOAuthenticator(GenericOAuthenticator):
     feature added in JupyterHub 5.0 (https://github.com/jupyterhub/jupyterhub/pull/4748).
     """
 
+    JHUB_SERVICE_ACCOUNT_NAME = Unicode()
+
+    @default("JHUB_SERVICE_ACCOUNT_NAME")
+    def _default_jhub_service_account_name(self):
+        return f"service-account-{self.client_id}"
```
aktech, line 26: We should add a line above on why the service account is named so, as there is no other usage of this expression anywhere else in the code.

Worth putting this link too:
https://github.com/keycloak/keycloak/blob/5e6bb9f7bd2c83febd12668f2605aa8ecbdcf130/docs/documentation/server_admin/topics/admin-cli.adoc?plain=1#L1008

aktech (changes requested):
Looks good overall, I haven't had the chance to test it locally due to M1 issues, but will try to sort that out tomorrow.

How do you feel about adding a test [here](https://github.com/nebari-dev/nebari/blob/f4ca133af4b2bde7a48f4a65b18e03bd787607f9/tests/tests_deployment/test_jupyterhub_api.py) verifying the service account has correct roles?

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/03-profiles.py
Adam-D-Lewis: this was needed

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/04-auth.py
```diff
@@ -30,6 +37,36 @@ class KeyCloakOAuthenticator(GenericOAuthenticator):
...
+        username = self.normalize_username(username)
+
+        # Build auth model similar to OAuth flow
+        auth_model = {
+            "name": username,
+            "admin": True if username in self.admin_users else None,
+            "auth_state": self.build_auth_state_dict(token_info, user_info),
+        }
+
+        auth_model = await self.update_auth_model(auth_model)
```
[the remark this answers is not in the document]
Adam-D-Lewis: done

src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/04-auth.py
Adam-D-Lewis: done

Adam-D-Lewis:
I was originally planning on adding a test once I had a ragna distribution running that would test this, but I'll add a test for just this part.

Adam-D-Lewis:
> Looks good overall, I haven't had the chance to test it locally due to M1 issues, but will try to sort that out tomorrow.
> 
> How do you feel about adding a test [here](https://github.com/nebari-dev/nebari/blob/f4ca133af4b2bde7a48f4a65b18e03bd787607f9/tests/tests_deployment/test_jupyterhub_api.py) verifying the service account has correct roles?

I added this test now in [80456c5](https://github.com/nebari-dev/nebari/pull/2917/commits/80456c597142dd239bc15547805757423a46eac9)

aktech:
Looks like its still failing: https://github.com/nebari-dev/nebari/actions/runs/13252428164/job/36993020527#step:13:220 (?)

Adam-D-Lewis:
> Looks like its still failing: [nebari-dev/nebari/actions/runs/13252428164/job/36993020527#step:13:220](https://github.com/nebari-dev/nebari/actions/runs/13252428164/job/36993020527#step:13:220) (?)

Thanks for pointing that out.  Yeah, this won't work until we do a jhub_apps release and include the latest in the jupyterhub image used by nebari.

Adam-D-Lewis:
Amit reminded me that the startup app expects a global-mypanel env which we aren't creating and I'll need to try to switch it over to the nebari-git/dashboards one or build the global-mypanel env if we want to actually try to open the startup server.

viniciusdc:
> Amit reminded me that the startup app expects a global-mypanel env which we aren't creating and I'll need to try to switch it over to the nebari-git/dashboards one or build the global-mypanel env if we want to actually try to open the startup server.

I am curious on how that will work, let me know if you need any help.

Adam-D-Lewis:
> > Amit reminded me that the startup app expects a global-mypanel env which we aren't creating and I'll need to try to switch it over to the nebari-git/dashboards one or build the global-mypanel env if we want to actually try to open the startup server.
> 
> I am curious on how that will work, let me know if you need any help.

According to @aktech, we would need to add:
- jhsingle-native-proxy
- bokeh-root-cmd
to the [dashboard environment](https://github.com/nebari-dev/nebari/blob/3c228407c084d26bab79c58718cb39c34dcd8f36/src/_nebari/stages/kubernetes_services/__init__.py#L323) to match the app [requirements](https://github.com/nebari-dev/jhub-apps-from-git-repo-example/blob/91c20063fe8d8496c3a6d7090455fb200c936358/environment.yml#L4).

so it should be straight forward.

viniciusdc:
@Adam-D-Lewis will follow up

Issue #2193, opened 2024-01-10
[ENH] - Filter all conda environments with given user permissions


aktech:
### Feature description

Currently we only filter the one's in user's namespace.

### Value and/or benefit

It would be nice be be able to use the shared environments and not just the ones in shared namespace.

nkaretnikov:
@marcelovilla asked offline whether there's an API to know what environments a particular user has access to.

It can be done for the _currently logged in user_ via a GET request to `http://localhost:8080/conda-store/api/v1/permission/`.

Example output for the default DB state:

```json
{
    "status": "ok",
    "data": {
        "authenticated": true,
        "primary_namespace": "test",
        "entity_permissions": {
            "default/*": [
                "environment::read",
                "namespace-role-mapping::read",
                "namespace::read"
            ],
            "filesystem/*": [
                "environment::read",
                "namespace-role-mapping::read",
                "namespace::read"
            ],
            "*/*": [
                "build::cancel",
                "build::delete",
                "environment::delete",
                "environment::read",
                "environment::solve",
                "environment::update",
                "environment:create",
                "namespace-role-mapping::create",
                "namespace-role-mapping::delete",
                "namespace-role-mapping::read",
                "namespace-role-mapping::update",
                "namespace::create",
                "namespace::delete",
                "namespace::read",
                "namespace::update",
                "setting::read",
                "setting::update"
            ]
        },
        "entity_roles": {
            "default/*": [
                "viewer"
            ],
            "filesystem/*": [
                "viewer"
            ],
            "*/*": [
                "admin"
            ]
        },
        "expiration": "2024-01-15T17:38:04+00:00"
    },
    "message": null
}
```

----

If it needs to be done for an _arbitrary user_ (given a name), there's code that does this on the backend side, but it's not exposed via the HTTP API.

More details: there are two versions of role mappings, which you select at startup time via the config by setting `role_mappings_version`. Depending on which version is used, one of these two handlers is called during auth, but there's no direct way to call these via the HTTP API.

```py
    _role_mappings_versions = {
        1: _database_role_bindings_v1,
        2: _database_role_bindings_v2,
    }
```

marcelovilla:
@nkaretnikov thank you for your suggestions.

For this issue, we do need to get the environments for an arbitrary username give its name as we'll use this information when spawning a jupyterhub server.

It is still unclear to me how the `conda-store/api/v1/permission/` endpoint would help. Here is an example response I get when calling that endpoint:
```json
{
[...]

PR #2447, opened 2024-05-04
Fetch JupyterHub roles from Keycloak

krassowski:
## Reference Issues or PRs

Closes #2308

It is useful to think about this PR as having two parts:
- (a) when user logs in (or their cookie gets refreshed), we extract the roles list from the auth data to ensure that their roles are up to date (which is handled by JupyterHub)
- (b) when restarting JupyterHub we load all the roles and role assignments from Keycloak using Keycloak REST API; this is to prevent users from being stripped from role assignments because auth-managed roles are not meant to be persisted indefinitely in JupyterHub (by definition). Note, that we _could_ decide not to strip the roles at all, but then if an administrator accidentally grants an elevated role to a user in Keycloak, the user would retain such a role forever (until manually modified in JupyterHub database).

## What does this implement/fix?

_Put a `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds a feature)
- [ ] Breaking change (fix or feature that would cause existing features not to work as expected)
- [ ] Documentation Update
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Other (please describe):

## Testing

- [ ] Did you test the pull request locally?
- [x] Did you add new tests? I will be developing with the feedback loop of the upstream CI because of https://github.com/nebari-dev/nebari/issues/2446. Sorry about the spam each time I push.

## Any other comments?

This PR includes https://github.com/nebari-dev/nebari/pull/2427 for now, but ideally https://github.com/nebari-dev/nebari/pull/2427 would be merged first.

As of this PR the roles have no scopes. There are a few ways to associate the roles with a scope, one is having a key-value map, another is encoding the scope specification in the role name or description; this is tracked in https://github.com/nebari-dev/nebari/issues/2432.

PR #2940, opened 2025-02-06
Filter environments by user access

soapy1:
This PR updates the `get_conda_environments` function called by jhub apps so that only the environments the user has access to are shown. This is achieved by generating a token for permissions that match the jupyterhub users' conda-store permissions.

## Reference Issues or PRs

https://github.com/nebari-dev/nebari/issues/2193

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

## How to test this PR?

0. To test this PR you must have
  a. an installation of Nebari with jhub apps enabled
  b. at least 2 users with some conda environments
1. Log in as the first user
2. Navigate to the nebari home screen
3. Click on the deploy app button
4. Click on the drop down to select the conda store environment for the app
5. Notice that all the environments the user has access to are available - that is default environments + its own environments, but not environments from the second user
6. Repeat for the second user


src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/jupyterhub/files/jupyterhub/02-spawner.py
```diff
@@ -46,12 +46,47 @@ def generate_paged_urls(base_url: str, total_records: int, page_size: int) -> li
...
+                "filesystem/*": ["viewer"],
+                "nebari-git/*": ["viewer"],
+                "global/*": ["viewer"]
+            }
+        }
+
+    if groups is not None:
+        for group in groups:
+            group = group.replace("/", "")
+            role_bindings["role_bindings"][f"{group}/*"] = ["viewer"]
```
soapy1, line 74: There is a pretty big assumption here, that the keycloak groups map directly to the conda-store namespaces. Is that true for nebari?

dcmcand: There is no direct way to create a new namespace through the ui, and group names do match namespaces, though there is also a namespace per user that matches their user name. 

That being said, it is possible to create a new namespace and give permissions to use it with an arbitrary name using the conda-store api.

Additionally, some groups, like admins, have access to more namespaces than just their own.

soapy1: Is there a doc (or piece of existing code) that describes the relationship between keycloak permissions/groups/etc and conda-store privileges?  I want to be able to ensure the user is able to see all the namespaces+environments it has access to.

dcmcand: There is custom auth logic at https://github.com/nebari-dev/nebari/blob/ac88391026377b068aecb8bcd74ce486d261f239/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/conda-store/config/conda_store_config.py#L105 which does the mapping

soapy1: hmmm, I see that's not quite what I'm looking for. The `user_info` dict provided ultimately comes from the jupyter api `/users` endpoint https://jupyterhub.readthedocs.io/en/5.2.1/reference/rest-api.html#operation/get-user which provides the users roles and groups. Here is an example of the output

```
{
  "name": "string",
  "admin": true,
  "roles": [
    "string"
  ],
  "groups": [
    "string"
  ],
  "server": "string",
  "pending": "spawn",
  "last_activity": "2019-08-24T14:15:22Z",
  "servers": {
  },
  "auth_state": {}
}
```

Is there some piece of code or docs that describe the relationship between hub users and keycloak users or conda store permissions?

dcmcand: @viniciusdc do you have any insight here for @soapy1?

krassowski: Chiming in as I worked on this - both groups and roles are taken from Keycloak since:
- https://github.com/nebari-dev/nebari/pull/2361
- https://github.com/nebari-dev/nebari/pull/2447

@viniciusdc can confirm if anything changed since as I was not reviewing PRs that followed, but I suspect this is still the case.

soapy1: Thanks  @krassowski those links were very helpful in piecing together a model of what is going on here. This part of the authenticator is also notable https://github.com/nebari-dev/nebari/blob/main/src/_nebari/stages/kubernetes_services/template/modules/kubernetes/services/conda-store/config/conda_store_config.py#L415.

I think this implementation where the permissions are constructed based on the `group` provided by the Jupyter user object should be correct. I added a few extra comments to explain the motivation.

This was pretty convoluted to piece together. I think it would be good if at least one other person also went thru the exercise of following the permissions flow from keycloak thru Jupyter and conda-store to double check that this approach is valid.

It would be nice if there was some kind of document that outlined the intended design of how permissions work in nebari. Maybe as part of the tasks in this issue https://github.com/nebari-dev/nebari/issues/2304?

dcmcand (changes requested):
@soapy1 this works great on a new deploy, but doesn't work if you have an existing deploy. If you have an existing deploy and upgrade you can still see all of the envs in jhub-apps.

soapy1:
Thanks @dcmcand for testing this out!

> but doesn't work if you have an existing deploy. If you have an existing deploy and upgrade you can still see all of the envs in jhub-apps.

I think I'm going to need a bit more information in order to debug this. In particular:
* what version of nebari did you start with?
* how did you upgrade?
* is this behaviour true for all users, or for a set of users with some characteristic?
* what are the keycloak groups that the offending user is part of? are they an admin?
* what version of keycloak is running?

dcmcand (changes requested):
:rocket:

dcmcand (approved):
:rocket:

marcelovilla:
@dcmcand tested this again and did not run into the [issue that was mentioned before](https://github.com/nebari-dev/nebari/pull/2940#pullrequestreview-2660906566). Thus, I'm merging this.

Thanks @soapy1 🚀 !

PR #3000, opened 2025-03-21
Add metallb as an optional install for the existing provider

dcmcand:
## Reference Issues or PRs
This will allow Nebari to optionally install metallb if installed to a cluster without a native load balancer such as with k3s.

- [X] New feature (non-breaking change which adds a feature)
- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [ ] Did you add new tests?

## How to test this PR?

Start a k3s cluster
```bash
curl -fL https://get.k3s.io | sh -s - server \
--tls-san YOUR_LAN_IP \
--disable traefik \
–disable systemlb \
--write-kubeconfig-mode 644 \
--flannel-backend=wireguard-native \
--cluster-init

```
Add the following to your nebari-config.yaml
```yaml
existing:
  kube_context: default
  metallb:
    enabled: true
    ip_min: YOUR_LAN_IP
    ip_max: YOUR_LAN_IP
```

install nebari off of this branch and verify that it runs.


viniciusdc:
I think we can remediate a few of these security warnings by inspecting the default manifests of MetalLB for our current version. On another note, I tested this yesterday, and it worked flawlessly, though on a second try, I got a pending state with the external IP that Metallb handles. Its most probably related to other custom changes on my system, so I will re-attempt that today once again with a clean plate. 

BTW, to leave here for future pondering, we should update our version of MetalLB since after 1.13+, the confimap object was replaced by new CRDs `IPAddressPool ` and `L2Advertisement `, my main reasoning is to have the most recent version available for security improvements, but also, there are a few new options such as restricting the interfaces Traefik broadcasts the IP.

PR #3048, opened 2025-05-06
initial attempt at spot process

satra:


aaronkanzer:
@satra -- just placing follow-up work here to not overstep your work here: https://github.com/nebari-dev/nebari/pull/3049

PR #3049, opened 2025-05-06
[Draft]: Include functionality for spot instances on AWS

aaronkanzer:
## Reference Issues or PRs

Follow-up on https://github.com/nebari-dev/nebari/pull/3048 (Cc @satra @yarikoptic @kabilar)

Provides ability to pass the `capacity_type` option downstream to AWS node groups in Terraform.

References argument passed in `eks_node_group` Terraform API: https://registry.terraform.io/providers/hashicorp/aws/5.43.0/docs/resources/eks_node_group

## What does this implement/fix?

• Add pydantic validation + field value for `capacity_type`
• Extendable -- string is optional, so shouldn't require any refactoring for current Nebari configs

_Put a `x` in the boxes that apply_

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds a feature)
- [ ] Breaking change (fix or feature that would cause existing features not to work as expected)
- [ ] Documentation Update
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] Other (please describe):

## Documentation

- [ ] For new features or enhancements, a corresponding PR has been opened in the [documentation repository](https://github.com/nebari-dev/nebari-docs) (if applicable)
  - Link to docs PR:

## Testing

- [ ] Did you test the pull request locally?
- [ ] Did you add new tests?

PR #3069, opened 2025-06-11
Enh/spot instance

satra:
## Reference Issues or PRs

Follows up on #3049

### DIFF
--- a/src/_nebari/stages/infrastructure/template/aws/modules/kubernetes/variables.tf
+++ b/src/_nebari/stages/infrastructure/template/aws/modules/kubernetes/variables.tf
@@ -46,7 +46,7 @@ variable "node_groups" {
   type = list(object({
     name            = string
     instance_type   = string
-    capacity_type   = optional(string)
+    spot            = bool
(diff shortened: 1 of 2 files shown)


marcelovilla:
Hey @satra, thanks for taking over https://github.com/nebari-dev/nebari/pull/3049! 

Did you see my comment there? https://github.com/nebari-dev/nebari/pull/3049#pullrequestreview-2916914727. I think some minor changes to make sure we use the `spot` key for the configuration and map it correctly in Terraform and then this should be ready to go. Do you have the bandwidth to add those changes?

satra:
aha i see. i will make those changes.

marcelovilla (approved):
I just tested this by triggering our AWS deployment tests, and everything is working as expected. The spot flag is being correctly passed in the configuration. Additionally, I ran a manual deployment to AWS and confirmed that the correct values are being applied when creating the node group.

image


Thanks for your contribution, @satra! We’ll include this in the next Nebari release, which is expected in the coming weeks.

PR #3105, opened 2025-07-21
bf: do not clobber existing .gitignore

### LINKED ISSUE #3104: secrets should not be committed (stages dir)
### Context

From the docs, it looks like its recommended to commit `stages/`. https://www.nebari.dev/docs/how-tos/nebari-stages-directory#nebari-cli-steps. Indeed the .gitignore produced by `nebari render` does not include `stages/`. However, all the config values, including secrets are in there. 

It seems to me that either `_nebari.tf.json` files should be gitignored, or maybe even the entire stages/ dir?

(A separate but related issue is that `nebari render` clobbers any existing .gitignore, fixed in PR #3105)

```
# ignore terraform state
.terraform
terraform.tfstate
terraform.tfstate.backup
.terraform.tfstate.lock.info

# python
__pycache__
```

If stages/ is intended to be committed, then whats the recommended setup for maintaining 2 or more deployments? Ie, in my case I have a staging deployment to test changes before pushing to prod. If they both must live in stages which are committed, this is very awkward and prone to human error. 

### Value and/or benefit

Secret safety

### Anything else?

_No response_

asmacdo:
Fixes #3104

When running `nebari render` my existing .gitignore was clobbered. If there is already a .gitignore, this step should be a no-op.

- [ ] Documentation Update
- [ ] Build related changes

## Testing

- [X] Did you add new tests?

## How to test this PR?

Use `nebari render` without a previously existing .gitignore, one should be created.
Use `nebari render` with a previously existing .gitignore, it should not be changed.

### DIFF
--- a/src/_nebari/stages/bootstrap/__init__.py
+++ b/src/_nebari/stages/bootstrap/__init__.py
@@ -13,9 +13,14 @@ from nebari.hookspecs import NebariStage, hookimpl
 
 def gen_gitignore():
     """
-    Generate `.gitignore` file.
-    Add files as needed.
+    Generate `.gitignore` file if not present.
     """
+    gitignore_path = pathlib.Path(".gitignore")
+
+    # If .gitignore already exists, don't overwrite it
+    if gitignore_path.exists():
+        return {}
+
     filestoignore = """
         # ignore terraform state
         .terraform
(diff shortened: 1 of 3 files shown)


asmacdo:
@viniciusdc spotted the issue with the tests. fixed in latest push.

viniciusdc (approved):

PR #3157, opened 2025-10-20
Add keycloak tests

### LINKED ISSUE #3156: Create tests for keycloak
### Context

Keycloak currently doesn't have any testing. 

### Value and/or benefit

Keycloak needs to be updated and having tests will enable us to ensure that the newer versions of keycloak are operating as expected. See #2495 

### Anything else?

Here is a breakdown of what needs to be done:

CRUD for users:
Make sure we can create a user, delete and patch/modify any of the user information, this also might include a list of the available users.
Keycloak has something called impersonate might be interesting to see if that can be used to perform a dummy action with the user itself – not really needed since we are kind of doing this within the playwright tests already
CRUD for clients:
We need to be able to create clients, list available ones, and to test validity within the created clients we should use a dummy Oauth2 class(?) (might be an external dep.) to test if the client is working properly (since it requires a secret and key from kc as well as interaction with the auth and token endpoints)

CRUD for roles/permissions:
Kind of tested already as part of jhub-apps tests (somewhere)

CRUD for groups:
Same create, delete and test assignments (e.g if a user is given a range of groups, does it receive the expected attributes and scopes from all the set of permissions attributed to the given groups? – inspect the user token response to check if the roles exists ) 
EG nebari_user: contain superusers, admin, guest
For example the scopes and permissions defined in here https://deploy-preview-554--nebari-docs.netlify.app/docs/how-tos/fine-grained-permissions#roles-groups-and-scopes are expected to be propagated based on being included or not in a given group

User Story:
Admin creates a user, some groups, and a client for testing
Test credentials for the created user, check group association
Test logging in to oauth client using the user
Test if scopes are present on user token

tylerpotts:
## Reference Issues or PRs
Competes and fixes #3156 

- [ ] Documentation Update
- [ ] Build related changes

## Documentation

- [ ] For new features or enhancements, a corresponding PR has been opened in the [documentation repository](https://github.com/nebari-dev/nebari-docs) (if applicable)
  - Link to docs PR:

## Testing

- [ ] Did you add new tests?

### DIFF
--- a/.github/workflows/test_conda_build.yaml
+++ b/.github/workflows/test_conda_build.yaml
@@ -39,7 +39,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          conda install grayskull conda-build conda-verify
+          conda install conda-build conda-verify
+          pip install grayskull
           pip install build
 
       - name: Conda list


tylerpotts:
@viniciusdc @dcmcand It looks like I misinterpreted the green check mark to mean these tests passed in the CI. They are passing locally.

Once the "Await Workloads" test is fixed on the main branch I'll pull that into this branch. Then the CI running these in the following "Deployment Pytests" will run and I can confirm things are green

viniciusdc:
The CI test is now fixed, small misjudgment on my part

tests/tests_deployment/test_keycloak_api.py
```diff
@@ -0,0 +1,1845 @@
...
+def authenticated_keycloak_api(
+    keycloak_base_url: str,
+    keycloak_username: str,
+    keycloak_password: str,
+    keycloak_realm: str,
+    keycloak_client_id: str,
+    verify_ssl: bool,
+) -> KeycloakAPI:
+    """Create an authenticated KeycloakAPI instance for admin operations."""
+    api = KeycloakAPI(
```
dcmcand: What happens if the token expires mid test (say the tests are slow)? Do we have token renewal handled for these tests?

tests/tests_deployment/keycloak_api_utils.py
```diff
@@ -0,0 +1,944 @@
+"""Helper class for Keycloak API interactions."""
+
+import base64
+import json
+
+import requests
+
+TIMEOUT = 10
```
dcmcand: Should this be configurable with a default? I am generally leery of hardcoded timeouts

tests/tests_deployment/keycloak_api_utils.py
```diff
@@ -0,0 +1,944 @@
...
+
+
+def decode_jwt_token(token: str) -> dict:
+    """Decode a JWT token without verification (for testing purposes).
```
dcmcand: Can we add a stronger warning here. "DO NOT USE THIS FOR PRODUCTION" or something like that. not verifying sigs in test is fine, I just don't want someone copying this pattern later in prod

tests/tests_deployment/test_keycloak_api.py
```diff
@@ -0,0 +1,1845 @@
...
+def test_keycloak_login_invalid_credentials(
+    keycloak_api: KeycloakAPI,
+) -> None:
+    """Test that authentication fails with invalid credentials."""
+    # Authenticate with invalid credentials should raise an HTTPError
+    with pytest.raises(requests.exceptions.HTTPError) as exc_info:
+        keycloak_api.authenticate("invalid_user", "invalid_password")
+
+    # Verify it's a 401 Unauthorized error
+    assert exc_info.value.response.status_code == 401
```
dcmcand: For all the cases where we are using numbers for http status codes, let's switch them to actual http.HTTPStatus objects
`assert exc_info.value.response.status_code == HTTPStatus.UNAUTHORIZED` for example. That will make the code clearer without needing to look up what a 401 is

tests/tests_deployment/test_keycloak_api.py
```diff
@@ -0,0 +1,1845 @@
...
+    get_response = authenticated_keycloak_api.get_users(username=test_username)
+    users = get_response.json()
+    user_id = users[0]["id"]
+
+    # Verify user can authenticate with old password
+    try:
+        token_data = keycloak_api.authenticate(test_username, old_password)
+        assert "access_token" in token_data
+    except requests.exceptions.HTTPError:
+        # Some Keycloak configurations may not allow password auth immediately
```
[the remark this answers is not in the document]
tylerpotts: Some keycloak configurations for users require the user to reset the initial password before first login, and I think that's why this was added by Claude.   

I tested with this try/except removed, and it passed. I think this is an LLM artifact and isn't needed.

tests/tests_deployment/test_keycloak_api.py
```diff
@@ -0,0 +1,1845 @@
+"""Keycloak API endpoint tests."""
```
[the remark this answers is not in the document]
tylerpotts: The only thing that I see as an issue is that they all share the same API instance. Since it now has auto-refresh, it would technically be possible for one worker to overwrite another worker's token mid-test. If I change from `session` to `function` for the `keycloak_api` and `authenticated_keycloak_api` it alleviates this problem.

I tested locally with the added dependency of pytest-xdist with 10 workers to test the parallel execution. It reduced the testing time from ~20 seconds to ~10 seconds

My thought is that added dependency and complexity of parallel execution isn't worth the benefit we get of 10s faster execution time.

tests/tests_deployment/keycloak_api_utils.py
tylerpotts: fixed

tests/tests_deployment/test_keycloak_api.py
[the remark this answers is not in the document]
dcmcand: that's fine, by concern was more around side effects from tests interfering and not necessarily about needing to run in parellel.

tylerpotts:
@dcmcand There was a missing expired GPG key in one of the providers. I updated it on this branch and the tests are passing now

Issue #3121, opened 2025-09-17
[BUG] - Nebari should be sending HSTS headers


sumphlet:
### Describe the bug

Currently Nebari does not send an HSTS header when you connect to the hostname: (example.com is replacing our actual hostname being used). HSTS headers for any https sites are required by our security plan.

```
$ curl -s -D- https://example.com |grep -i Strict
$
```

### Expected behavior

A strict transport security header should be returned. For example:

```
$ curl -s -D- https://nasa.gov |grep -i Strict
strict-transport-security: max-age=31536000;includeSubdomains;preload
```

### OS and architecture in which you are running Nebari

Red Hat Enterprise Linux 9

### How to Reproduce the problem?

The easiest way to reproduce the problem is to run a curl command on the hostname. You can also check your headers in your browsers dev console:

Example curl command, replacing example.com with the hostname of your nebari server.

```
curl -s -D- https://example.com |grep -i Strict
```

### Command output

```bash session
$ curl -s -D- https://example.com |grep -i Strict
$
```

### Versions and dependencies used.

$ conda --version
conda 4.14.0

$ kubectl version
Client Version: v1.33.1
Kustomize Version: v5.6.0
Server Version: v1.32.4+k3s1

$ nebari --version
2025.6.1

### Compute environment

None

### Anything else?

We were given some instructions to possibly enable this, but it hasn't worked so far:

Under the nebari folder, create the following yaml file `middleware-hsts-1y.yaml`:
```
apiVersion: [traefik.containo.us/v1alpha1
kind: Middleware
metadata:
  name: hsts-1y
  namespace: dev
spec:
  headers:
    stsSeconds: 31536000
    stsIncludeSubdomains: true
```
push the middleware to the cluster with, kubectl apply -f middleware-hsts-1y.yaml,  then in the nebari-config.yaml file, 

```
ingress:
  terraform_overrides:
    additiona_-arguments:
    - "entrypoints.websecure.http.middlewares=dev-hsts-1y@kubernetescrd"
```
Here's a describe of the middleware entry it created:
```
Name:         hsts-1y
Namespace:    dev
Labels:       <none>
Annotations:  <none>
API Version:  traefik.containo.us/v1alpha1
Kind:         Middleware
Metadata:
  Creation Timestamp:  2025-09-15T15:41:30Z
  Generation:          1
  Resource Version:    67443691
  UID:                 8f1dedf8-8859-4d78-801b-ab39bb4b261a
Spec:
  Headers:
[...]

PR #3165, opened 2025-10-23
[Ready for Review] Add HSTS (HTTP Strict Transport Security) support

Adam-D-Lewis:
Closes https://github.com/nebari-dev/nebari/issues/3121
Docs Issue - https://github.com/nebari-dev/nebari-docs/issues/605
We should add some stuff in the next release upgrade - https://github.com/nebari-dev/nebari/issues/3168

## Summary

This PR adds built-in HSTS (HTTP Strict Transport Security) configuration to Nebari's ingress layer.

- Adds configurable HSTS middleware for Traefik
- Applies HSTS headers to both websecure and minio entrypoints when enabled

If HSTS is unspecified, then it will be enabled for existing, and let's encrypt certs, but not enabled for self-signed or no cert.
### Example Configuration

```yaml
ingress:
  hsts:
    enabled: true
    max_age: 30 
    include_subdomains: true
    preload: false
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,1 +1,10 @@
+import warnings
+
+warnings.filterwarnings(
+    "ignore",
+    message=r"You are using a Python version \(3\.10\..*\) which Google will stop supporting",
+    category=FutureWarning,
+    module="google.api_core",
+)
+
 pytest_plugins = ["tests.common.playwright_fixtures"]


Adam-D-Lewis:
I haven't tested this via a cloud deployment yet, but assuming that works, I think it's ready for review.

PR #3184, opened 2025-11-11
Fix Azure authentication in GitHub Actions

Adam-D-Lewis:
Removes `ARM_USE_OIDC` environment variable that conflicts with the `azure/login@v2` action.


Adam-D-Lewis:
Example of faliing azure login before: https://github.com/nebari-dev/nebari/actions/runs/19272302138/job/55103487006
And example of passing azure login now: https://github.com/nebari-dev/nebari/actions/runs/19273737586

